PR #60, opened 2021-01-28
[Example] Add an animate a line example

ZiZasaurus:
This PR adds an animate a geoJSON line example.

Screen recording:
animation

### DIFF
--- a/Examples/Examples/All Examples/AnimateGeoJSONLine.swift
+++ b/Examples/Examples/All Examples/AnimateGeoJSONLine.swift
@@ -2,56 +2,54 @@ import UIKit
 import MapboxMaps
 import Turf
 
-
-
 @objc(AnimateGeoJSONLine)
 public class AnimateGeoJSONLine: UIViewController, ExampleProtocol {
-    
+
     internal var mapView: MapView!
     internal let sourceIdentifier = "route-source-identifier"
     internal var routeLineSource: GeoJSONSource!
     var currentIndex = 0
-    
+
     public var geoJSONLine = (identifier: "routeLine", source: GeoJSONSource())
-    
+
     override public func viewDidLoad() {
         super.viewDidLoad()
-        
+
         mapView = MapView(with: view.bounds, resourceOptions: resourceOptions())
         mapView.autoresizingMask = [.flexibleWidth, .flexibleHeight]
         self.view.addSubview(mapView)
-        
+
         let centerCoordinate = CLLocationCoordinate2D(latitude: 45.5076, longitude: -122.6736)
-        
+
         mapView.cameraManager.setCamera(centerCoordinate: centerCoordinate,
                                         zoom: 11.0)
-        
+
         // Allows the delegate to receive information about map events.
         mapView.on(.mapLoadingFinished) { [weak self] _ in
-            
+
             guard let self = self else { return }
             self.addLine()
             self.animatePolyline()
-            
+
             // The below line is used for internal testing purposes only.
             self.finish()
         }
     }
-    
+
     func addLine() {
-        
+
         // Create a GeoJSON data source.
         routeLineSource = GeoJSONSource()
         routeLineSource.data = .feature(Feature(LineString([allCoordinates[currentIndex]])))
-        
+
         // Create a line layer
         var lineLayer = LineLayer(id: "line-layer")
         lineLayer.source = sourceIdentifier
         lineLayer.paint?.lineColor = .constant(ColorRepresentable(color: UIColor.red))
-        
+
         let lowZoomWidth = 5
         let highZoomWidth = 20
-        
+
         // Use an expression to define the line width at different zoom extents
         lineLayer.paint?.lineWidth = .expression(
             Exp(.interpolate) {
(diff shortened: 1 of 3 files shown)


Examples/Examples/All Examples/AnimateGeoJSONLine.swift
```diff
@@ -0,0 +1,189 @@
...
+        currentIndex += 1
+        print(currentIndex)
+    }
+    
+    func updatePolylineWithCoordinates(coordinates: [CLLocationCoordinate2D]) {
+        let mutableCoordinates = coordinates
+
+        // Identify the new coordinate to animate to
+        geoJSONLine.source.data = .feature(Feature(LineString(mutableCoordinates)))
+        let geoJSON = Feature.init(geometry: Geometry.lineString(LineString(mutableCoordinates)))
```
captainbarbosa, line 109: We added an extension to `Feature` you can use to avoid the long initializer - you can now turn:

```swift
Feature.init(geometry: Geometry.lineString(LineString(mutableCoordinates)))
```

Into this:

```swift
Feature(LineString(mutableCoordinates)
```

Examples/Examples/All Examples/AnimateGeoJSONLine.swift
```diff
@@ -0,0 +1,189 @@
...
+        (-122.70346, 45.49674),
+        (-122.70517, 45.49758),
+        (-122.70614, 45.49736),
+        (-122.70663, 45.49736),
+        (-122.70807, 45.49767),
+        (-122.70807, 45.49798),
+        (-122.70717, 45.49798),
+        (-122.70713, 45.4984),
+        (-122.70774, 45.49893)
+    ].map({CLLocationCoordinate2D(latitude: $0.1, longitude: $0.0)})
```
captainbarbosa, line 188: Hmmmm this seems confusing - I would just use an array of `CLLocationCoordinates` so we don't have to do this lat/long `map` at the end.

Examples/Examples/All Examples/AnimateGeoJSONLine.swift
```diff
@@ -0,0 +1,189 @@
...
+        // Add the lineLayer to the map.
+        self.mapView.style.addSource(source: geoJSONLine.source, identifier: geoJSONLine.identifier)
+        self.mapView.style.addLayer(layer: lineLayer)
+        
+    }
+    
+    func animatePolyline() {
+        currentIndex = 1
+        
+        // Start a timer that will simulate adding points to our polyline. This could also represent coordinates being added to our polyline from another source, such as a CLLocationManagerDelegate.
```
captainbarbosa, line 83:
```suggestion
        // Start a timer that will add a new coordinate to the line and redraw it every time it repeats.
```

Examples/Examples/All Examples/AnimateGeoJSONLine.swift
```diff
@@ -0,0 +1,189 @@
...
+        }
+    }
+    
+    func addLine() {
+        let coordinate = coordinates[currentIndex]
+        // Create a GeoJSON data source.
+        geoJSONLine.source.data = .feature(Feature(geometry: Geometry.lineString(LineString([coordinate]))))
+        
+        var lineLayer = LineLayer(id: "line-layer")
+        lineLayer.filter = Exp(.eq) {
```
captainbarbosa, line 52: This filtering isn't necessary, because we're not adding multiple shapes from the `GeoJSONSource`.

Examples/Examples/All Examples/AnimateGeoJSONLine.swift
```diff
@@ -0,0 +1,189 @@
...
+            "$type"
+            "LineString"
+        }
+        lineLayer.source = geoJSONLine.identifier
+        lineLayer.paint?.lineColor = .constant(ColorRepresentable(color: UIColor.red))
+        
+        let lowZoomWidth = 5
+        let highZoomWidth = 20
+        lineLayer.paint?.lineWidth = .expression(
```
captainbarbosa, line 61: I don't think this is fundamental to understand the example, but if you're going to keep it please include a lil comment explaining what's happening here.

captainbarbosa (changes requested):
This `AnimateGeoJSONLine` example was added to the examples directory, but the change to the `.pbxproj` file wasn't registered here. This causes `MapboxMaps.xcworkspace` to not see the file, so the project won't run. You'll need to commit the change that adds this to the `.pbxproj` file as part of this PR.

I've also pushed up a branch to limit the number of variables declared on the view controller, just so you can get a different perspective on other ways this example could be refactored. You can have a [look at that here](https://github.com/mapbox/mapbox-maps-ios/commit/848e8bb5bf15a81ad20a54670189ae876f8ee193), but don't feel like you need to implement it entirely - I'll approve your changes as long as you address the comments mentioned here and fix the `.pbxproj` file :)

Examples/Examples/All Examples/AnimateGeoJSONLine.swift
```diff
@@ -0,0 +1,174 @@
...
+        mapView = MapView(with: view.bounds, resourceOptions: resourceOptions())
+        mapView.autoresizingMask = [.flexibleWidth, .flexibleHeight]
+        self.view.addSubview(mapView)
+        
+        let centerCoordinate = CLLocationCoordinate2D(latitude: 45.5076, longitude: -122.6736)
+        
+        mapView.cameraManager.setCamera(centerCoordinate: centerCoordinate,
+                                        zoom: 11.0)
+        
+        // Allows the delegate to receive information about map events.
```
captainbarbosa, line 29: We don't use a delegate-based callback system here anymore, so this comment is outdated.

```suggestion
        // Wait for the map to load its style before adding data.
```

captainbarbosa (approved):
✅ Feel free to merge after addressing the one doc comment and making sure SwiftLint passes.

PR #70, opened 2021-02-02
Refactor Annotation "properties"

neelmistry94:

## Pull request checklist:

### Summary of changes
The `Annotation` protocol has a property called `properties`. This is confusing since that property is more so for user info and results in an incorrect use when setting it to Feature.properties. This update prevents rendering issues and incorrect usage of those properties.


captainbarbosa (approved):
_Technically_, this is a breaking change since this property used to be public, but its so early on and in beta I don't think its worth calling out.

julianrex (approved):

PR #75, opened 2021-02-03
[tests] Adding generated integration tests

nishant-karajgikar:

## Pull request checklist:

### Summary of changes

This PR adds generated integration tests for all layers (sources to come shortly). The tests set a mocked value for every layer property (on every layer -- layout + paint) and checks if the value is being successfully applied.


neelmistry94 (approved):
@nishant-karajgikar One thing I'm curious about here is that if we are making `visibility` optional. Will that break any existing functionality or usage of that property?

nishant-karajgikar:
> @nishant-karajgikar One thing I'm curious about here is that if we are making `visibility` optional. Will that break any existing functionality or usage of that property?

Nope it shouldn't!

PR #76, opened 2021-02-03
[style] Convert `ColorRepresentable` to a struct

nishant-karajgikar:

## Pull request checklist:

### Summary of changes

This PR converts `ColorRepresentable` to a struct (instead of a typedef on `String`).

This makes it a lot easier to discover the intializer as well as supports our test efforts. It also makes the initializer "failable" since we only support the sRGB color space.

### DIFF
--- a/Mapbox/MapboxMapsStyle/Types/Color.swift
+++ b/Mapbox/MapboxMapsStyle/Types/Color.swift
@@ -8,17 +8,28 @@ public struct ColorRepresentable: Codable, Equatable {
 
     /// Create a string representation of a `UIColor`
     /// - Parameter color: A `UIColor` instance in the sRGB color space
-    /// - Returns: Initializes a `ColorRepresentable` instance if the `color` is in sRGB color space. Returns `nil` otherwise.
-    public init?(color: UIColor) {
+    /// - Returns: Initializes a `ColorRepresentable` instance if the `color` is in sRGB color space.
+    public init(color: UIColor) {
         var red: CGFloat = 0.0
         var green: CGFloat = 0.0
         var blue: CGFloat = 0.0
         var alpha: CGFloat = 0.0
-        if color.getRed(&red, green: &green, blue: &blue, alpha: &alpha) {
+        let success = color.getRed(&red, green: &green, blue: &blue, alpha: &alpha)
+        let validColorComponents = Self.isValidColor(red: red, green: red, blue: blue, alpha: alpha)
+        if success && validColorComponents {
             self.colorRepresentation = "rgba(\(red * 255.0), \(green * 255.0), \(blue * 255.0), \(alpha))"
         } else {
-            return nil
+            fatalError("Please use a color in the sRGB color space")
         }
+    }
+
+    /// Checks if all color components are within the 0-1 range
+    internal static func isValidColor(red: CGFloat, green: CGFloat, blue: CGFloat, alpha: CGFloat) -> Bool {
+        let validRange = 0.0...1.0
+        return validRange ~= Double(red)
+            && validRange ~= Double(green)
+            && validRange ~= Double(blue)
+            && validRange ~= Double(alpha)
     }
 
     public func encode(to encoder: Encoder) throws {
@@ -32,7 +43,7 @@ public struct ColorRepresentable: Codable, Equatable {
     }
 }
 
-extension UIColor : ValidExpressionArgument {
+extension UIColor: ValidExpressionArgument {
 
     public var expressionElements: [Expression.Element] {
         var red: CGFloat = 0.0
--- a/Mapbox/MapboxMapsStyle/Types/Color.swift
+++ b/Mapbox/MapboxMapsStyle/Types/Color.swift
@@ -32,38 +32,6 @@ public struct ColorRepresentable: Codable, Equatable {
     }
 }
 
-public extension UIColor {
-
-    /// Initialize a `UIColor` from a `ColorRepresentable`
-    convenience init?(hex: String?) {
-        guard let hex = hex else { return nil }
-
-        let red, green, blue, alpha: CGFloat
-
-        if hex.hasPrefix("#") {
-            let start = hex.index(hex.startIndex, offsetBy: 1)
-            let hexColor = String(hex[start...])
-
-            if hexColor.count == 8 {
-                let scanner = Scanner(string: hexColor)
-                var hexNumber: UInt64 = 0
-
-                if scanner.scanHexInt64(&hexNumber) {
-                    red = CGFloat((hexNumber & 0xff000000) >> 24) / 255
-                    green = CGFloat((hexNumber & 0x00ff0000) >> 16) / 255
-                    blue = CGFloat((hexNumber & 0x0000ff00) >> 8) / 255
-                    alpha = CGFloat(hexNumber & 0x000000ff) / 255
-
-                    self.init(red: red, green: green, blue: blue, alpha: alpha)
-                    return
-                }
-            }
-        }
-
-        return nil
-    }
-}
-
 extension UIColor : ValidExpressionArgument {
 
     public var expressionElements: [Expression.Element] {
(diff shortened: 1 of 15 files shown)


Mapbox/MapboxMapsStyle/Types/Color.swift
```diff
@@ -1,48 +1,34 @@
...
+public struct ColorRepresentable: Codable, Equatable {
+
+    /// String representation of a `UIColor` used by the renderer
+    public let colorRepresentation: String?
 
     /// Create a string representation of a `UIColor`
-    public init(color: UIColor) {
+    /// - Parameter color: A `UIColor` instance in the sRGB color space
+    /// - Returns: Initializes a `ColorRepresentable` instance if the `color` is in sRGB color space. Returns `nil` otherwise.
+    public init?(color: UIColor) {
```
macdrevx, line 12: I'm not convinced that making this initializer failable results in a better API. In the modified examples above, we now have to force unwrap. My guess is that most consumers would do the same, at which point we might as well just have a fatal error on line 20 below instead of returning nil.

Mapbox/MapboxMapsStyle/Types/Color.swift
```diff
@@ -1,48 +1,34 @@
...
+/// Container to represent `UIColor`for use by the map renderer
+public struct ColorRepresentable: Codable, Equatable {
+
+    /// String representation of a `UIColor` used by the renderer
+    public let colorRepresentation: String?
 
     /// Create a string representation of a `UIColor`
-    public init(color: UIColor) {
+    /// - Parameter color: A `UIColor` instance in the sRGB color space
+    /// - Returns: Initializes a `ColorRepresentable` instance if the `color` is in sRGB color space. Returns `nil` otherwise.
```
macdrevx, line 11: Technically, the color only needs to be in a colorspace that is "compatible" with the "extended sRGB" color space for the `getRed(_:green:blue:alpha:)` call to succeed.

https://developer.apple.com/documentation/uikit/uicolor/1621919-getred

Mapbox/MapboxMapsStyle/Types/Color.swift
```diff
@@ -1,48 +1,34 @@
...
+    public init?(color: UIColor) {
         var red: CGFloat = 0.0
         var green: CGFloat = 0.0
         var blue: CGFloat = 0.0
         var alpha: CGFloat = 0.0
-        color.getRed(&red, green: &green, blue: &blue, alpha: &alpha)
-        self = "rgba(\(red * 255.0), \(green * 255.0), \(blue * 255.0), \(alpha))"
+        if color.getRed(&red, green: &green, blue: &blue, alpha: &alpha) {
+            self.colorRepresentation = "rgba(\(red * 255.0), \(green * 255.0), \(blue * 255.0), \(alpha))"
```
macdrevx, line 18: On iOS 10 and later, `getRed(_:green:blue:alpha:)` uses the extended sRGB color space for the RGB values. This means that the returned values may be less than 0 and greater than 1.

How should we handle such cases?

Mapbox/MapboxMapsStyleTests/ColorTests.swift
```diff
@@ -0,0 +1,82 @@
...
+            XCTFail("Color data is nil")
+            return
+        }
+
+        do {
+            let decodedColor = try JSONDecoder().decode(ColorRepresentable.self, from: validData)
+            XCTAssert(decodedColor == color, "Color should not change after encoding - decoding pass")
+        } catch {
+            XCTFail("Could not decode color")
+        }
```
macdrevx, line 30: There's a lot of duplication between this test and the last two in this file. Could we create a helper method?

Mapbox/MapboxMapsStyle/Types/Color.swift
nishant-karajgikar: Yep, but the style spec only supports `sRGB` color space -- https://docs.mapbox.com/mapbox-gl-js/style-spec/types/#color

Mapbox/MapboxMapsStyle/Types/Color.swift
nishant-karajgikar: ah, good point! Considering that the renderer only supports the sRGB color space, feels like we should just return nil / fail in that situation.

Mapbox/MapboxMapsStyle/Types/Color.swift
nishant-karajgikar: Yeah, I'm not opposed to that.. the API does feel sub-optimal. The only thing is we don't seem to be using `fatalError` anywhere else in the SDK.

Mapbox/MapboxMapsStyle/Types/Color.swift
macdrevx: I wonder if there's a standard way to clamp to [0, 1]. Presumably the system does this when it encounters an extended range color that it needs to render on a display that only supports sRGB.

Mapbox/MapboxMapsStyle/Types/Color.swift
nishant-karajgikar: Let's use `fatalError` and see where else it may be a better fit in the sdk.

Mapbox/MapboxMapsStyle/Types/Color.swift
```diff
@@ -1,52 +1,49 @@
...
-                    green = CGFloat((hexNumber & 0x00ff0000) >> 16) / 255
-                    blue = CGFloat((hexNumber & 0x0000ff00) >> 8) / 255
-                    alpha = CGFloat(hexNumber & 0x000000ff) / 255
+    /// Checks if all color components are within the 0-1 range
+    internal static func isValidColor(red: CGFloat, green: CGFloat, blue: CGFloat, alpha: CGFloat) -> Bool {
+        let validRange = 0.0...1.0
+        return validRange ~= Double(red)
+            && validRange ~= Double(green)
+            && validRange ~= Double(blue)
+            && validRange ~= Double(alpha)
```
macdrevx, line 32:
```suggestion
        let validRange: CGFloat = 0.0...1.0
        return validRange ~= red
            && validRange ~= green
            && validRange ~= blue
            && validRange ~= Double(alpha)
```

Does that work?

Mapbox/MapboxMapsStyleTests/ColorTests.swift
```diff
@@ -0,0 +1,50 @@
...
+        encodeAndDecodeColor(testColor: .white)
+    }
+
+    func testNonSRGBColor() throws {
+        let testColor = UIColor(patternImage: UIImage())
+        let color = ColorRepresentable(color: testColor)
+        XCTAssertNil(color)
+    }
+
+    func encodeAndDecodeColor(testColor: UIColor) {
```
macdrevx, line 29:
```suggestion
    func encodeAndDecodeColor(testColor: UIColor, line: UInt = #line) {
```

macdrevx: And then forward the line parameter to the XCT* test assertions in this method. This will make it so that if one of them fails, Xcode will highlight the call site of this method which helps to identify which test failed.

Mapbox/MapboxMapsStyle/Types/Color.swift
nishant-karajgikar: I'd tried that, doesn't work unfortunately.

Mapbox/MapboxMapsStyleTests/ColorTests.swift
nishant-karajgikar: Nice!

Mapbox/MapboxMapsStyle/Types/Color.swift
macdrevx: What about this:

```suggestion
        let validRange: ClosedRange<CGFloat> = 0.0...1.0
        return validRange ~= red
            && validRange ~= green
            && validRange ~= blue
            && validRange ~= alpha
```

macdrevx (approved):

Mapbox/MapboxMapsStyle/Types/Color.swift
macdrevx: For future reference: https://developer.apple.com/documentation/coregraphics/cgcolor/1455493-converted

PR #81, opened 2021-02-04
[style] Sort stops dictionaries before conversion to expression

nishant-karajgikar:

## Pull request checklist:

 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-ios` changelog: `<changelog>Dictionaries passed to expressions are now sorted by default</changelog>`.

### Summary of changes
This PR fixes a bug where unsorted stops dictionaries would not be processed by the style renderer. The fix essentially makes it so that the ExpressionBuilder sorts the stops dictionary (by key) before generating the expression elements.


Mapbox/MapboxMapsStyleTests/ExpressionTests/ExpressionBuilderTests.swift
```diff
@@ -0,0 +1,37 @@
+import XCTest
+#if canImport(MapboxMaps)
+@testable import MapboxMaps
+#else
+@testable import MapboxMapsStyle
+#endif
+
+internal class ExpressionBuilderTests: XCTestCase {
```
neelmistry94, line 8: Would be good to add a test that checks for a negative key to make sure potential future changes doesn't cause this bug

nishant-karajgikar: Hmm can't test for that since there's an `assertionFailure` unfortunately. The original bug was also actually caused by an unsorted stops dictionary. That's what's being tested below.

neelmistry94 (approved):
I tested your branch locally with respect to the issue that was opened and it LGTM. 

I have limited expression experience so you might want to get a second pair of eyes to look at the code just in case.

PR #99, opened 2021-02-12
Fix pan drift with non-zero bearing

macdrevx:
## Pull request checklist:
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-ios` changelog: `<changelog>Fixed: Pan drift did not work correctly when bearing was non-zero.</changelog>`.

### Summary of changes

- The directionality of the antimeridian correction was being calculated under the assumption that the bearing was 0. With a non-zero bearing, we need to instead consider the component of the offset vector that runs due east.

### DIFF
--- a/Mapbox/MapboxMapsFoundation/BaseMapView.swift
+++ b/Mapbox/MapboxMapsFoundation/BaseMapView.swift
@@ -63,7 +63,11 @@ open class BaseMapView: UIView, MapClient, MBMMetalViewProvider {
 
     /// The map's current center coordinate.
     public var centerCoordinate: CLLocationCoordinate2D {
-        return cameraView.centerCoordinate
+        // cameraView.centerCoordinate is allowed to exceed [-180, 180]
+        // so that core animation interpolation works correctly when
+        // crossing the antimeridian. We wrap here to hide that implementation
+        // detail when accessing centerCoordinate via BaseMapView
+        return cameraView.centerCoordinate.wrap()
     }
 
     /// The map's current zoom level.
--- a/Mapbox/MapboxMapsFoundation/Camera/CameraManager.swift
+++ b/Mapbox/MapboxMapsFoundation/Camera/CameraManager.swift
@@ -483,13 +483,25 @@ public class CameraManager {
         var newLong: Double
 
         // This logic is to prevent a rubber band effect when a pan's drift takes you across the antimeridian.
-        if offset.x < 0 {
+
+        // First calculate the scalar projection of the offset onto the unit vector pointing due east.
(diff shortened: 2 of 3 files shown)


Mapbox/MapboxMapsFoundation/BaseMapView.swift
```diff
@@ -63,7 +63,7 @@ open class BaseMapView: UIView, MapClient, MBMMetalViewProvider {
 
     /// The map's current center coordinate.
     public var centerCoordinate: CLLocationCoordinate2D {
-        return cameraView.centerCoordinate
+        return cameraView.centerCoordinate.wrap()
```
macdrevx, line 66: `cameraView.centerCoordinate.longitude` must be allowed to be > 180 and < -180 so that the Core Animation-based interpolation works correctly.

neelmistry94: maybe I'm interpreting something incorrectly but when going through the chain, it looks like the wrap() function normalizes the coordinate to be with in -180..180.

neelmistry94 (approved):
Ran the unit tests and on sim and things look good. This is a nice fix.

One open question up using `wrap()`

Mapbox/MapboxMapsFoundation/BaseMapView.swift
macdrevx: > when going through the chain

Which chain?

neelmistry94: Like I stepped into the wrap function which eventually took me [here](https://github.com/mapbox/mapbox-maps-ios/blob/main/Mapbox/MapboxMapsFoundation/Extensions/Core/Utils.swift#L36)

macdrevx: Added a comment to capture what we discussed.

Mapbox/MapboxMapsFoundation/BaseMapView.swift
```diff
@@ -63,7 +63,11 @@ open class BaseMapView: UIView, MapClient, MBMMetalViewProvider {
 
     /// The map's current center coordinate.
     public var centerCoordinate: CLLocationCoordinate2D {
-        return cameraView.centerCoordinate
+        // cameraView.centerCoordinate is allowed to exceed [-180, 180]
+        // so that core animation interpolation works correctly when
+        // crossing the antimeridian. We wrap here to hide that implementation
+        // detail when accessing centerCoordinate via BaseMapView
```
neelmistry94, line 69: Awesome. Thanks

PR #100, opened 2021-02-12
[bug] Disable pan drift for pitched maps

neelmistry94:
## Pull request checklist:
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-ios` changelog: `<changelog>Pan drift for pitched maps will be disabled. A solution for smooth drifting is being worked on.</changelog>`.

### Summary of changes
This PR will temporarily disable drift for pitched maps when a user is panning. The current experience when panning a pitched maps results in the map jumping to undesired location. We are currently working on a final solution for smooth drifting.


Mapbox/MapboxMapsGestures/GestureManager.swift
```diff
@@ -283,7 +283,10 @@ extension GestureManager: GestureHandlerDelegate {
 
     // Pan has ended on the MapView with a residual `offset`
     internal func panEnded(with offset: CGPoint) {
-        self.cameraManager.moveCamera(by: offset, animated: true)
+        if let pitch = self.cameraManager.mapView?.pitch,
+           pitch == 0.0 {
```
macdrevx, line 287:
```suggestion
        if cameraManager.mapView?.pitch == 0.0 {
```

Mapbox/MapboxMapsGestures/GestureManager.swift
```diff
@@ -283,7 +283,10 @@ extension GestureManager: GestureHandlerDelegate {
...
+            self.cameraManager.moveCamera(by: offset, animated: true)
```
macdrevx, line 288:
```suggestion
            cameraManager.moveCamera(by: offset, animated: true)
```

macdrevx (approved):

nishant-karajgikar (approved):

Mapbox/MapboxMapsGestures/GestureManager.swift
neelmistry94: we are using `self.` in most places so would like to keep that for consistency

Mapbox/MapboxMapsGestures/GestureManager.swift
neelmistry94: I thought about this, but what if mapView is nil for some reason.

Mapbox/MapboxMapsGestures/GestureManager.swift
macdrevx: Sounds good.

Mapbox/MapboxMapsGestures/GestureManager.swift
macdrevx: If mapView is nil, the expression will evaluate to false.

macdrevx: Which seems fine. You'll just be avoiding the assertionFailure inside of moveCamera

macdrevx: This has the same behavior as what you wrote initially, it's just more concise.

PR #101, opened 2021-02-12
[location] Fix bug where locationOptions were not being set correctly

nishant-karajgikar:

## Pull request checklist:
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-ios` changelog: `<changelog>Fix issue where toggling LocationOptions.showsUserLocation resulted in options not being updated</changelog>`.

### Summary of changes

This change ensures that the stored value of `LocationOptions` in `LocationManager` is updated with the newly passed in values before the `updateLocationOptions()` method returns.

This fixes a bug where location options would not get updated if `showsUserLocation == false`.

### DIFF
--- a/Mapbox/MapboxMapsLocation/LocationManager.swift
+++ b/Mapbox/MapboxMapsLocation/LocationManager.swift
@@ -86,15 +86,13 @@ public class LocationManager: NSObject {
 
         guard newOptions != locationOptions else { return }
 
+        // Update the location options
+        self.locationOptions = newOptions
+        self.locationProvider.locationProviderOptions = newOptions
+
         if newOptions.showUserLocation != showUserLocation {
             self.showUserLocation = newOptions.showUserLocation
             toggleUserLocationUpdates(showUserLocation: self.showUserLocation)
-
-            if !newOptions.showUserLocation {
-                // If we should not show user location, then we should
-                // not try and change source or style below
-                return
-            }
         }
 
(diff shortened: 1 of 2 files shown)


Mapbox/MapboxMapsLocation/LocationManager.swift
```diff
@@ -86,26 +86,20 @@ public class LocationManager: NSObject {
...
+
         if newOptions.showUserLocation != showUserLocation {
             self.showUserLocation = newOptions.showUserLocation
             toggleUserLocationUpdates(showUserLocation: self.showUserLocation)
-
-            if !newOptions.showUserLocation {
-                // If we should not show user location, then we should
-                // not try and change source or style below
-                return
-            }
```
nishant-karajgikar, line 97: This wasn't providing any sort of optimization.

neelmistry94: what happens in the case if someone decides to stop showing user location, and for some reason they have change another option like location puck or something below?

nishant-karajgikar: any reason why those changes should not take effect? This `return` feels like premature optimization.

neelmistry94 (approved):

julianrex (approved):

macdrevx (commented):
@nishant-karajgikar can you add a unit test that fails without this fix and passes when the fix is applied?

PR #103, opened 2021-02-12
Prevent pitch and zoom from exceeding limits

macdrevx:
## Pull request checklist:
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-ios` changelog: `<changelog>Prevent pitch and zoom from exceeding limits. Also updates default maximum pitch to 85 degrees.</changelog>`.

### Summary of changes

- Consolidates logic from `setCamera(centerCoordinate:padding:anchor:zoom:bearing:pitch:animated:duration:completion:)` into `setCamera(to:animated:duration:completion:)`
- Clamps pitch to min and max from `MapCameraOptions`
- Refactors the clamp helper into an extension on `Comparable` that uses a ClosedRange parameter
- Improves `testSetCamera()` which had several assertions that checked for "changes" that were identical to the initial values.
- Updates default maximumPitch to 85 degrees

### User impact (optional)

When continuing to use the pitch gesture after reaching the maximum or minimum pitch, the map will no longer require you to "undo" those superfluous gestures to reverse the pitch.

### DIFF
--- a/Mapbox/MapboxMapsFoundation/Camera/CameraManager.swift
+++ b/Mapbox/MapboxMapsFoundation/Camera/CameraManager.swift
@@ -165,7 +165,7 @@ public class CameraManager {
      - Parameter completion: The completion block to execute after the transition has occurred.
      */
 
-    public func setCamera(to newCamera: CameraOptions,
+    public func setCamera(to camera: CameraOptions,
                              animated: Bool = false,
                              duration: TimeInterval? = 0,
                              completion: ((Bool) -> Void)? = nil) {
@@ -175,13 +175,21 @@ public class CameraManager {
             return
         }
 
-        guard mapView.cameraView.camera != newCamera else {
+        let clampedCamera = CameraOptions(
+            center: camera.center,
+            padding: camera.padding,
+            anchor: camera.anchor,
+            zoom: camera.zoom?.clamped(to: mapCameraOptions.minimumZoomLevel...mapCameraOptions.maximumZoomLevel),
+            bearing: optimizeBearing(startBearing: mapView.bearing, endBearing: camera.bearing),
+            pitch: camera.pitch?.clamped(to: mapCameraOptions.minimumPitch...mapCameraOptions.maximumPitch))
+
+        guard mapView.cameraView.camera != clampedCamera else {
             completion?(true)
             return
         }
 
         let animation = {
-            mapView.cameraView.camera = newCamera
+            mapView.cameraView.camera = clampedCamera
         }
 
         performCameraAnimation(animated: animated, duration: duration, animation: animation, completion: completion)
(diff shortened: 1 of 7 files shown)


Mapbox/MapboxMapsFoundation/Camera/CameraManager.swift
```diff
@@ -165,7 +165,7 @@ public class CameraManager {
      - Parameter completion: The completion block to execute after the transition has occurred.
      */
 
-    public func setCamera(to newCamera: CameraOptions,
+    public func setCamera(to camera: CameraOptions,
```
neelmistry94, line 168: @julianrex we spoke about cleaning up these methods. But with this change to only having this single setCamera method. It removes the ability to opt out of optimizing bearing. We should probably add that logic here

Mapbox/MapboxMapsFoundation/Camera/CameraManager.swift
```diff
@@ -175,13 +175,21 @@ public class CameraManager {
...
         }
 
-        guard mapView.cameraView.camera != newCamera else {
+        let clampedCamera = CameraOptions(
+            center: camera.center,
+            padding: camera.padding,
+            anchor: camera.anchor,
+            zoom: camera.zoom?.clamped(to: mapCameraOptions.minimumZoomLevel...mapCameraOptions.maximumZoomLevel),
+            bearing: optimizeBearing(startBearing: mapView.bearing, endBearing: camera.bearing),
+            pitch: camera.pitch?.clamped(to: mapCameraOptions.minimumPitch...mapCameraOptions.maximumPitch))
```
nishant-karajgikar, line 184: Could we also increase the default `maximumPitch` value in `MapCameraOptions`?

Mapbox/MapboxMapsFoundation/Extensions/Comparable+Clamped.swift
```diff
@@ -0,0 +1,11 @@
+internal extension Comparable {
```
nishant-karajgikar, line 1: Super nice!

Mapbox/MapboxMapsFoundation/Camera/CameraManager.swift
macdrevx: Yes. What should I change it to?

nishant-karajgikar (approved):

Mapbox/MapboxMapsFoundation/Camera/CameraManager.swift
macdrevx: Updated to 85. cc: @mapbox/maps-android

Mapbox/MapboxMapsFoundation/Camera/CameraManager.swift
macdrevx: Based on discussion with @julianrex we'll move forward with this change and revisit adding an option to opt-out of bearing optimization.

PR #118, opened 2021-02-18
Cocoapods Build From Source

neelmistry94:
PRs must be submitted under the terms of our Contributor License Agreement [CLA](https://github.com/mapbox/mapbox-maps-ios/blob/main/CONTRIBUTING.md#contributor-license-agreement).

Fixes: < Link to related issues that will be fixed by this pull request, if they exist >

## Pull request checklist:
 - [x] Briefly describe the changes in this PR.
 - [x] Apply changelog label ('breaking change', 'bug :beetle:', 'build', 'docs', 'feature :green_apple:', 'performance :zap:', 'testing :100:') or use the label 'skip changelog'
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-ios` changelog: `<changelog>Consumption of MapboxMaps via Cocoapods will now build from source</changelog>`.

### Summary of changes
Updates the podspec to use a build from source approach. It also defaulting to updating source builds for MME and Turf.

In order to test
1) Create a new Xcode project
2) In the command line, navigate to the root of your new tester app, and then type `pod init`. This will create a Podfile for you.
3) Open your Podfile, and under your app target, add this line `pod 'MapboxMaps', :git => 'https://github.com/mapbox/mapbox-maps-ios.git', :branch => 'nm/pods_build_from_source'`
4) In the command line, type `pod install --repo-update` to install this as a dependency
5) Open your project.xcworkspace and then try building a map

PR #125, opened 2021-02-19
Update our Package.swift to build MapboxMaps from source

jmkiley:
## Pull request checklist:
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-ios` changelog: `<changelog>MapboxMaps can now be built and tested using Swift Package Manager</changelog>`.

#### Configuration Changes

- Updated `Package.swift` to build the project from source
- Reorganized the source files according to SPM conventions
- Updated the paths of the groups in the MapboxMaps Xcode project to point to the new file locations
- Marked `OrnamentsLocalizable.strings` as a localized resource
- Updated `MBXGeometry.swift` to conditionally compile the `@unknown default:` of a switch statement with an expression of type `Geometry` from Turf. This is required since when we build the SDK as an XCFramework, Turf is compiled in library evolution mode, so a `default` case is required, but when we build the SDK as a Swift package, Turf is not built in library evolution mode, so the `default` case is unnecessary (and the compiler warns if it is present).
- `.gitignore` updates:
  - Stopped ignoring `IDEWorkspaceCHecks.plist`
  - Stopped ignoring `.swiftpm`
  - Stopped ignoring `Package.resolved`
  - Started ignoring `MapboxAccessToken`

#### Dependency Version Changes

- Switched to MapboxMobileEvents v0.10.8 (Carthage & SPM only)
- Switched to Turf v1.2.0 (SPM only)

### Bundle & Info.plist Usage Changes

- Added a Build Pre-Action to MapboxMaps's SPM scheme to copy the Mapbox access token from either `~/.mapbox` or `~/mapbox` into the new `MapboxAccessToken` file. The integration tests need this token. Previously, they looked for it in the test bundle's `Info.plist`, but this approach does not work for SPM. `IntegrationTestCase.setupAccessToken()` has been updated accordingly.
- Added `Bundle.mapboxMaps` and `Bundle.mapboxMapsTests` abstractions to allow accessing the SDK's resources. These abstractions handle differences that occur when building the SDK as a Swift package vs building it as a framework. In the SDK source, `Bundle.main` is used when the intent is to access the bundle of the containing application, and `Bundle.mapboxMaps` is used when the intent is to use the SDK's bundle.
- Replaced usages of `Bundle.object(forInfoDictionaryKey:)` with `Bundle.infoDictionary?[]`
- Hard-coded the SDK version in `EventsManager.swift` because we will not be able to customize the SDK's `Info.plist` when the SDK is built as a Swift package.

### Fixes

- Fixed an issue in `MapViewIntegrationTests.setUpWithError()` in which `super.setUpWithError()` was invoked with `try?` instead of `try`, preventing it from being able to skip the test.
- Reviewed Xcode 12.4's upgrade recommendations
- Fixed a `canImport` in `MapboxMapsAnnotationsTests.swift`

### Deletions

- Deleted unused installation tests
- Deleted some placeholder SPM-style tests in the `Tests` directory
- Deleted some unused source files and dangling file references


Examples/Examples.xcodeproj/xcshareddata/xcschemes/Examples.xcscheme
```diff
@@ -38,6 +38,16 @@
...
+         <TestableReference
+            skipped = "NO">
+            <BuildableReference
+               BuildableIdentifier = "primary"
+               BlueprintIdentifier = "077C4F0A252F7E89007636F1"
+               BuildableName = "ExamplesUITests.xctest"
+               BlueprintName = "ExamplesUITests"
+               ReferencedContainer = "container:Examples.xcodeproj">
+            </BuildableReference>
+         </TestableReference>
```
macdrevx, line 50: Undo?

Package.swift
```diff
@@ -1,151 +1,67 @@
...
+                "Annotations/Info.plist",
+                "Foundation/Info.plist",
+                "Gestures/Info.plist",
+                "Location/Info.plist",
+                "MapView/Info.plist",
+                "Offline/Info.plist",
+                "Ornaments/Info.plist",
+                "Snapshot/Info.plist",
+                "Style/Info.plist",
+                "MapView/Integration Tests/HTTP/HTTPIntegrationTests.swift"
```
macdrevx, line 51:
```suggestion
                "MapView/Integration Tests/HTTP/HTTPIntegrationTests.swift",
```

For consistency with other arrays wrapped to multiple lines.

macdrevx: Main benefit of this style is cleaner diffs as things are added/removed to the end of these arrays.

Sources/MapboxMaps/Foundation/Events/EventsManager.swift
```diff
@@ -9,12 +9,7 @@ internal class EventsManager: EventsListener {
     var telemetry: TelemetryProtocol!
 
     init(accessToken: String) {
-        let bundle = Bundle(for: type(of: self))
-        var sdkVersionOptional = bundle.infoDictionary?["MGLSemanticVersionString"]
-        if sdkVersionOptional == nil {
-            sdkVersionOptional = bundle.infoDictionary?["CFBundleShortVersionString"]
-        }
-        guard let sdkVersion = sdkVersionOptional as? String else { return }
+        let sdkVersion = "10.0.0"
```
macdrevx, line 12: Create a ticket to find a better solution

Examples/Examples.xcodeproj/xcshareddata/xcschemes/Examples.xcscheme
macdrevx: It's fine

Cartfile.resolved
```diff
@@ -1,4 +1,4 @@
 binary "https://api.mapbox.com/downloads/v2/carthage/mapbox-common/MapboxCommon-ios.json" "10.0.0-beta.11"
 binary "https://api.mapbox.com/downloads/v2/carthage/mobile-maps-core/MapboxCoreMaps-dynamic.json" "10.0.0-beta.15"
-github "mapbox/mapbox-events-ios" "v0.10.7"
+github "mapbox/mapbox-events-ios" "v0.10.8"
 github "mapbox/turf-swift" "v2.0.0-alpha.2"
```
jmkiley, line 4: Do we want to update Turf here and in the podspec as well?

macdrevx (approved):

Cartfile.resolved
macdrevx: No, we still need this version for Carthage so that it gets built in library evolution mode. For CocoaPods, we will switch to 1.2.0 as part of #118.

Package.swift
```diff
@@ -1,151 +1,67 @@
...
+    platforms: [.iOS(.v11)],
+    products: [
+        .library(
+            name: "MapboxMaps",
+            targets: ["MapboxMaps"]),
+    ],
+    dependencies: [
+        .package(name: "MapboxCommon", url: "https://github.com/mapbox/mapbox-common-ios.git", .exact("10.0.0-beta.11")),
+        .package(name: "MapboxCoreMaps", url: "https://github.com/mapbox/mapbox-core-maps-ios.git", .exact("10.0.0-beta.15")),
+        .package(name: "MapboxMobileEvents", url: "https://github.com/mapbox/mapbox-events-ios.git", .exact("0.10.8")),
```
1ec5, line 18: Does this need to be `exact`, or can it be `from:` like the Turf dependency and Cartfile?

PR #131, opened 2021-02-22
[CI] Script fixes

neelmistry94:

## Pull request checklist:

### Summary of changes
This PR updates our recent CI failures by fixing branch filtering and updating the manifest script.

Previously, the filter on the CI config was not properly filtering for branches. It resulted in all jobs being ignored for the "Public Beta Release" job. This PR correctly uses a regex to make sure that the "Public Beta Release" job is only run on a release branch.

This also fixes a broken script when updating manifest, while also reflecting the changes made to building from source.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -114,7 +114,7 @@ workflows:
                 name: "Public Beta Release"
                 filters:
                     branches:
-                        only: Release/*
+                        only: /nm\/.*/
 
             - ios-release-tag:
                 name: "iOS release"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -357,7 +357,7 @@ jobs:
             - run:
                 name: Update Package Manifest & Podspec
                 command: |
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -114,7 +114,7 @@ workflows:
                 name: "Public Beta Release"
                 filters:
                     branches:
-                        only: Release/*
+                        only: /Release\/.*/
```
macdrevx, line 117: @neelmistry94 can you update this PR to explain in more detail how this change works? What was broken about the old way? How does the new way fix it?

macdrevx: @neelmistry94 I still don't understand what was broken. Can you give an example of a branch name that didn't match the original pattern and explain why it didn't match and then explain why it does match the new pattern?

neelmistry94: So what I had previously without the "//" it was not recognizing it as a regex. 

So when we push up a branch called Release/v10.0.0.... it did not work because it was not a valid regex. The addition of the slashes will make it correct

macdrevx (approved):

PR #134, opened 2021-02-23
Swiftlint fixes & configuration changes; Make GestureManager.delegate weak

macdrevx:
## Pull request checklist:
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-ios` changelog: `<changelog>Breaking: GestureManager.delegate is now weak</changelog>`.

### Summary of changes

- Migrated swiftlint config to use opt_in_rules and disabled_rules so that we can more easily discover new rules as they are added.
- Updated swiftlint excludes for new source layout
- Autocorrected various lint violations
- Corrected `GestureManager.delegate` weak delegate violation
- Disabled trailing_comma rule since we'd need to wrap closing array braces consistently for it to make sense
- Added `swiftlint:disable` commands to ignore other lint warnings
- Added a child config for the Tests directory that is slightly more permissive.
- Run swiftlint in CI directly rather than depending on it running as a build phase. This will be helpful as we migrate to SPM-based builds in CI.
- Run swiftlint in strict mode in CI. This should fail builds that have lint so that we don't regress.


julianrex (approved):
LGTM

.swiftlint.yml
```diff
@@ -2,84 +2,18 @@ allow_zero_lintable_files: true
...
-  - redundant_void_return
-  - return_arrow_whitespace
-  - statement_position
-  - superfluous_disable_command
-  - switch_case_alignment
-  - syntactic_sugar
+disabled_rules:
+  - comment_spacing
+  - computed_accessors_order
+  - force_try
```
neelmistry94, line 13: Why are disabling force_try. This feels like something we'd want to lint on

Sources/MapboxMaps/Style/Generated/Terrain.swift
```diff
@@ -7,7 +7,7 @@ import Foundation
 
  See the [Mapbox Style Specification](https://www.mapbox.com/mapbox-gl-style-spec/#terrain) for more information.
  */
-public struct Terrain: Codable  {
+public struct Terrain: Codable {
```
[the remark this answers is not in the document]
macdrevx: Oops, I meant to remove changes to the generated sources. Looks like this one slipped through the cracks.

neelmistry94 (approved):
One open question, nothing that prevents me from approving

.swiftlint.yml
macdrevx: I agree, but it wasn't enabled previously, and enabling it will be a rather large undertaking that I didn't have time to tackle in this PR.

It's on my list of things to attempt in the future.

PR #148, opened 2021-02-25
[build] CI Script Fixes

neelmistry94:

## Pull request checklist:

### Summary of changes
This PR fixes errors found from the last release in our CI scripts. The following have been fixed:
- generate changelog script now works and will point to the branch that the user passes as input.
- The update dependency script (renamed to reflect spm and pods) will push after completion, and read versions from versions.json instead of the Cartfile.
- Downstream impacts have been adjusted


scripts/release/generate_changelog.sh
```diff
@@ -13,34 +13,33 @@ set -euo pipefail
...
 
-# Remove any existing tags if there are any, and then fetch all remote tags
+# Remove any existing local tags if there are any, and then fetch all remote tags
 git tag -d $(git tag -l)
 git fetch --tags
 
 # Get all tags matching '10.0.0*' in order of latest activity
 # TODO: Modularize this so that it is not hardcoded to version 10.0.0
-ORDERED_TAGS=`git tag -l --sort=-committerdate -l "10.0.0*"`
+ORDERED_TAGS=`git tag -l --sort=-committerdate -l "v10.0.*"`
```
nishant-karajgikar, line 31: This isn't a valid assumption right? What happens if there's a patch release for `beta.12` for example?

neelmistry94: correct, I do have a TODO there to modularize it.

To your point though, this script is designed to compare w.e branch you provide it, to the last tag. So if we did a patch to beta 12, we wouldn't want to compare that to like beta 14.

neelmistry94: Also beta 12(public) onwards all have the v prefix. So this change would only break for private beta patches

nishant-karajgikar (approved):

macdrevx (approved):

PR #154, opened 2021-03-02
Example: use SceneKit in custom layer

astojilj:
Port "antenna building" example from gl-js https://docs.mapbox.com/mapbox-gl-js/example/add-3d-model/

https://user-images.githubusercontent.com/549216/111467256-e4b04480-872c-11eb-8573-87d496f72671.mov

The example demonstrates direct port from https://docs.mapbox.com/mapbox-gl-js/example/add-3d-model/ using Three.js to SceneKit.

For that purpose, [original GLTF model](https://docs.mapbox.com/mapbox-gl-js/assets/34M_17/34M_17.gltf) is converted (using Blender) to dae. Textures are embedded in attached dae file. When exporting in Blender, it is important to set with Y-up in export setting. Original model is available from 

[Code comments](https://github.com/mapbox/mapbox-maps-ios/compare/astojilj-scenekit?expand=1#diff-8671d0aeace7da36367f139a5c042eff4960a3a7f6b0bd4da9a1ec4ca1ad8967R182) provide in depth explanation on matrices transformations required to geo-reference models on map.

Initial version used SceneKit [SCNMatrix4 operations](https://developer.apple.com/documentation/scenekit/scnmatrix4) for positioning model on the map. That led to visible jittering of model relative to camera. Similar issue was observed using simd floating point calculations. The jittering looked like this:

https://user-images.githubusercontent.com/549216/111467039-a7e44d80-872c-11eb-9a37-55c5c10aa421.mov

To resolve that, it is necessary to use simd double calculation for following:

```
        let transform = transformSimd * translateModel * modelScale
```

Issues:
- [x] There was a lag/delay when panning the map. projectionTransform property is animatable and it is need to call SCNTransaction.flush() to use it in current frame.

Follow up: After #155 , implement CPU side occlusion as an option.


Examples/Examples/All Examples/SceneKitExample.swift
```diff
@@ -0,0 +1,227 @@
...
+        transformSimd[2, 0] = m[8].doubleValue
+        transformSimd[2, 1] = m[9].doubleValue
+        transformSimd[2, 2] = m[10].doubleValue
+        transformSimd[2, 3] = m[11].doubleValue
+        transformSimd[3, 0] = m[12].doubleValue
+        transformSimd[3, 1] = m[13].doubleValue
+        transformSimd[3, 2] = m[14].doubleValue
+        transformSimd[3, 3] = m[15].doubleValue
+
+        let origin = try! Projection.project(for: modelOrigin, zoomScale: 1.0 / 512.0);
```
julianrex, line 174: What's the 512 here?

Examples/Examples/All Examples/SceneKitExample.swift
```diff
@@ -0,0 +1,227 @@
...
+        matrix[1, 1] = yScale
+        matrix[2, 2] = zScale
+    
+        return matrix
+    }
+    
+    public func render(_ parameters: CustomLayerRenderParameters, mtlCommandBuffer: MTLCommandBuffer, mtlRenderPassDescriptor: MTLRenderPassDescriptor) {
+        let m = parameters.projectionMatrix;
+        
+        // It is essential to use double precision for computation below: using simd instead
```
julianrex, line 154: Can you say why - presumably because of jittering with floats?

Examples/Examples/All Examples/SceneKitExample.swift
```diff
@@ -0,0 +1,227 @@
...
+        cameraNode.camera = camera;
+        camera.usesOrthographicProjection = false;
+        scene.rootNode.addChildNode(cameraNode);
+        renderer.pointOfView = cameraNode;
+        self.setupLight()
+        // In order to use depth occlusion, align with gl-native Z handling (doesn't use reverse Z).
+        if #available(iOS 13.0, *) {
+            renderer.usesReverseZ = false
+        } else {
+            // Fallback on earlier versions, disable depth in render()
```
julianrex, line 100: What happens in these cases? Should we disable this example completely if < iOS 13?

Examples/Examples/All Examples/SceneKitExample.swift
astojilj: Yes, mercator coordinate addresses whole world and there's just not enough precision if using 32-bit floating point computation.

Examples/Examples/All Examples/SceneKitExample.swift
astojilj: We should enable it, just there is a need to use CPU side occlusion (the same approach that could be used when geo-referencing UI elements to map). The idea is to disable depth buffer (now disabled in demo) and show or hide model with animated fade in / fade out (similar to behavior of markers in gl-js).

90252440-2cde5000-de48-11ea-8f3f-6bffa52e07e1

As we cannot have occlusion done yet, current behavior on iOS < 13 is like here (model always visible):

https://user-images.githubusercontent.com/549216/109847499-ee7b7780-7c57-11eb-8e3d-e33b9c1c0f75.mov

Examples/Examples/All Examples/SceneKitExample.swift
astojilj: Explained in code comment:

        // Projection.project(for: modelOrigin, zoomScale: 1.0 / 512.0) corresponds to gl-js's
        // mapboxgl.MercatorCoordinate.fromLngLat(). origin is in spherical mercator normalized to
        // 0..1 for the width of the world. In other words, (x,y) E [0..1) is used to represent
        // coordinates in one world copy, values of x +/- 1 represent wrap.
        let origin = try! Projection.project(for: modelOrigin, zoomScale: 1.0 / 512.0)

astojilj:
There was a lag/delay when panning the map. projectionTransform property is animatable and it is need to call SCNTransaction.flush() to use it in current frame. Updated video shows improved behavior.

@julianrex PTAL

Examples/Examples/All Examples/SceneKitExample.swift
```diff
@@ -0,0 +1,246 @@
...
+        let properties = [
+            "id": "terrain_hillshade",
+            "type": "hillshade",
+            "source": "mapbox-dem",
+            "hillshade-illumination-anchor": "map"
+        ] as [ String: Any ]
+
+        let insertHillshadeBelow = try! map.styleLayerExists(forLayerId: "water") ?
+            LayerPosition(above: nil, below: "water", at: nil) : try! map.styleLayerExists(forLayerId: "hillshade") ?
+            LayerPosition(above: nil, below: "hillshade", at: nil) : nil
```
julianrex, line 78: This is a little hard to understand; can we split this into multiple lines.

Examples/Examples/All Examples/SceneKitExample.swift
```diff
@@ -0,0 +1,246 @@
...
+        // origin is in normalized MercatorCoordinates. Normalized refers to the
+        // world copy represented with (x, y) values in range [0..1], and corresponds
+        // to modelAsMercatorCoordinate in https://docs.mapbox.com/mapbox-gl-js/example/add-3d-model/
+        let transformModel = makeTranslationMatrix(tx: origin.x, ty: origin.y, tz: elevation * meterInMercatorCoordinateUnits)
+
+        // the same scale as in gl-js example, scale from meters to mercator.
+        let modelScale = makeScaleMatrix(xScale: meterInMercatorCoordinateUnits, yScale: -meterInMercatorCoordinateUnits, zScale: meterInMercatorCoordinateUnits)
+
+        // mercator scale is specific to gl-native example because gl-js's customLayerMatrix computes this
+        // internaly: https://github.com/mapbox/mapbox-gl-js/blob/main/src/geo/transform.js#L1316
```
julianrex, line 197: nit: internally.

Examples/Examples/All Examples/SceneKitExample.swift
```diff
@@ -0,0 +1,246 @@
...
+        transformSimd[3, 0] = m[12].doubleValue
+        transformSimd[3, 1] = m[13].doubleValue
+        transformSimd[3, 2] = m[14].doubleValue
+        transformSimd[3, 3] = m[15].doubleValue
+
+        // Projection.project(for: modelOrigin, zoomScale: 1.0 / 512.0) corresponds to gl-js's
+        // mapboxgl.MercatorCoordinate.fromLngLat(). origin is in spherical mercator normalized to
+        // 0..1 for the width of the world. In other words, (x,y) E [0..1) is used to represent
+        // coordinates in one world copy, values of x +/- 1 represent wrap.
+        let origin = try! Projection.project(for: modelOrigin, zoomScale: 1.0 / 512.0)
```
julianrex, line 180: I still don't understand what the 512 represents here and in the subsequent code; shouldn't the projection calculation here involve `parameters.zoom`?

1ec5 (commented):
This is super exciting to see. The simd usage significantly improves upon the approach in https://github.com/mapbox/ios-sdk-examples/issues/111#issuecomment-423516559.

It would be wonderful to see MapboxMaps better encapsulate SceneKit integration – perhaps even as a “SceneKitLayer” class. There are a lot of transformations and settings here that would be common to most usage of SceneKit. I think developers would likely copy-paste the whole example rather than take advantage of any flexibility gained by implementing it at the application level.

Examples/Examples/All Examples/SceneKitExample.swift
astojilj: Unnecessary complication - replaced the code with simple `LayerPosition(above: nil, below: "water", at: nil)` as it is present in style used.

Examples/Examples/All Examples/SceneKitExample.swift
```diff
@@ -0,0 +1,246 @@
...
+
+        self.mapView.on(.styleLoadingFinished) { [weak self] _ in
+            self?.addModelAndTerrain()
+        }
+    }
+
+    override public func viewDidAppear(_ animated: Bool) {
+        super.viewDidAppear(animated)
+         // The below line is used for internal testing purposes only.
+        self.finish()
```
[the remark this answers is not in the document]
astojilj: Done. Test passes. However, as it is the case with this and several other tests, I could see 

```
Example timed out, was this intentional? Call finish() if possible.
Test Case '-[ExamplesTests.TestableExampleTests testSceneKitExample]' passed (20.534 seconds).
```

Examples/Examples/All Examples/SceneKitExample.swift
astojilj: This is a good point. I have apparently approached this from wrong angle; the idea was to complicate matrices and calculation, enhance the complexity with differences between gl-js and gl-native custom layer transformation matrix and Web mercator scaling internals, in order to present 1-1 mapping of gl-js example to approach here. It is wrong and if using parameters.zoom this is much simpler to follow (and less code).

astojilj:
@julianrex PTAL

Examples/Examples/All Examples/SceneKitExample.swift
```diff
@@ -0,0 +1,228 @@
...
+    }
+
+    func addModelAndTerrain() {
+        try! mapView.__map.addStyleCustomLayer(forLayerId: "Custom",
+                                                layerHost: self,
+                                            layerPosition: LayerPosition(above: nil, below: "waterway-label", at: nil))
+
+        var demSource = RasterDemSource()
+        demSource.url = "mapbox://mapbox.mapbox-terrain-dem-v1"
+        demSource.tileSize = 514
```
julianrex, line 52: Is 514 correct?

Examples/Examples/All Examples/SceneKitExample.swift
julianrex: Can you add that to the comment please?

Examples/Examples/All Examples/SceneKitExample.swift
```diff
@@ -0,0 +1,228 @@
...
+        var elevation = 0.0
+        if let elevationData = parameters.elevationData, let elevationValue = elevationData.getElevationFor(self.modelOrigin) {
+            elevation = elevationValue.doubleValue
+        }
+        let translateModel = makeTranslationMatrix(tx: origin.x, ty: origin.y, tz: elevation)
+
+        let transform = transformSimd * translateModel * modelScale
+
+        let scnMat = SCNMatrix4(
+            m11: Float(transform[0, 0]),
```
julianrex, line 187: Have you noticed any jittering of the model relative to the world?

Examples/Examples/All Examples/SceneKitExample.swift
```diff
@@ -0,0 +1,228 @@
...
+        if self.useCPUOcclusion {
+            mtlRenderPassDescriptor.depthAttachment = nil
+            mtlRenderPassDescriptor.stencilAttachment = nil
+            // Example uses depth buffer to occlude model when e.g. behind the hill.
+            // If depth buffer (SCNRenderer.usesReverseZ = false) is not available, or if wished to
+            // to indicate that model is occluded or e.g. implement fade out / fade in model occlusion,
+            // the example here needs to provide CPU side occlusion implementation, too.
+            // TODO: this is blocked on https://github.com/mapbox/mapbox-maps-ios/issues/155
+        }
+        if let colorTexture = mtlRenderPassDescriptor.colorAttachments[0].texture {
```
julianrex, line 219: Can we add this as a guard at the top of this function? So 

```swift
guard let colorTexture = mtlRenderPassDescriptor.colorAttachments[0].texture else {
    return
}
```

What would it mean if the model can be rendered without the map?

julianrex (approved):
A few comments that would be good to address, but otherwise LGTM. Approved so you're not blocked.

Examples/Examples/All Examples/SceneKitExample.swift
astojilj: This code won't get called if there is no mtlRenderPassDescriptor.colorAttachments[0].texture.

Examples/Examples/All Examples/SceneKitExample.swift
astojilj: 514 is used for padded (border ready) DEM tiles. I wanted to test it here and left the value.

Padded DEM tiles are performance improvement, especially for hillshade: borders are prepared on server side and there is no need to [backfill borders from neighboring DEM tiles](https://github.com/mapbox/mapbox-gl-js/blob/4d32dbe61128d279821f7f50dc0c1d9d804b7ed4/src/data/dem_data.js#L53). 

Anyway, it is unnecessary, additional information for this PR. I'll remove before merging.

jmkiley (approved):
Thank you for the PR!

Examples/Examples/All Examples/SceneKitExample.swift
astojilj: Now that simd is used, no.

Cal-um:
hey @astojilj thanks for this great example. Im looking at using something similar to this with models on a map. Do you have any idea how to respond to a touch on one of these models? I am trying out the hittest method on the renderer but having no luck unless you are directly above it and close.

NickLambert95:
@astojilj Is there an example for this on Android SDK?

astojilj:
@Cal-um,
Please check https://github.com/mapbox/mapbox-gl-js/issues/10595#issuecomment-823906783: there you'll find explanation for a solution in gl-js with custom layer in THREE.js. Following the same example, this would mean that we could try to use

```
- (NSArray<SCNHitTestResult *> *)hitTestWithSegmentFromPoint:(SCNVector3)pointA 
                                                     toPoint:(SCNVector3)pointB 
                                                     options:(NSDictionary<NSString *,id> *)options;
```
for hit testing. If you have time to experiment with it, please try, I will also allocate some time later this week to check it.
It could be useful to know a bit more about number, size and shape of models you use and if terrain is active.

astojilj:
> @astojilj Is there an example for this on Android SDK?

@NickLambert95 ,
I've answered in https://github.com/mapbox/mapbox-navigation-android/issues/4502#issuecomment-861248252. Let's see if that helps.

NickLambert95:
@astojilj I just replied to your comment, it seems you closed the issue thread so I just wanted to let you know

Cal-um:
Hey @astojilj Sorry for pestering you on this but I am a complete noob at this. Perhaps I am approaching this wrong but I have tried taking the CGPoint of the tap and unprojecting it from the renderer method with a SCNVector3 of z 0 and 1. It seems to work when rendering a line with it but it's at completely the wrong point on the map. 

On a different note would you need a new scene and layer for every distinct model rendered on the map?

Thanks in advance.

PR #179, opened 2021-03-15
Migrate DebugApp to SPM

macdrevx:
## Pull request checklist:

### Summary of changes

* Creates a new Apps workspace that contains DebugApp
* Updates DebugApp to include its dependencies via the working copy of the MapboxMaps Swift package
* The xcconfigs used by the DebugApp needed to be modified, so they were duplicated to avoid interfering with the other projects that also depend on them.
* DebugApp's config has been updated to allow it to build on Apple Silicon
* DebugApp no longer uses a custom derived data path since that was not working with SPM. CI has been updated to locate the derived data path dynamically.
* DebugApp's config has been updated to remove search directory overrides that are no longer required
* DebugApp has been removed from MapboxMaps.xcworkspace
* Developing guide has been updated
* Codecov config has been updated
* CI has been updated, splitting Debug App builds into a separate job
    * Includes a new setup step for netrc since we're no longer using `make deps`
    * Includes a workaround for Xcode's SSH issue when installing Swift packages
    * No longer builds DebugApp with its Release configuration (only builds in Debug configuration)
    * Log archive structure has been slightly modified — previously, it contained a root-level `build` directory. Now, the contents of that directory are at root level.
    * Uses Xcode 12.4 for DebugApp jobs


.circleci/config.yml
```diff
@@ -23,6 +23,9 @@ workflows:
             parameters:
               configuration: ["Debug", "Release"]
               report_failure: [true]
+      - build-debug-app:
+          report_failure: true
+          xcode: "12.4.0"
```
neelmistry94, line 28: our min version of Xcode is 12.2, should we switch to that? or maybe add matrix

Apps/Apps.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -0,0 +1,43 @@
+{
```
neelmistry94, line 1: Not sure what the protocol here is, but does the `Package.resolved` need to be committed? If not, we should remove this file and `Package.resolved` to `.gitignore`

.circleci/config.yml
macdrevx: We should definitely consider setting up matrixes for supported Xcode versions. I'll create a separate ticket for that though, because I think it deserves a well-thought-out plan and discussion with the team.

Apps/Apps.xcworkspace/xcshareddata/swiftpm/Package.resolved
macdrevx: It's okay for it to be committed. It makes it so that we can control *when* to update dependencies once their version requirements are specified more flexibly.

neelmistry94:
@macdrevx This is amazing. I think since we are moving apps to this "Apps" directory. Can we move the Stress Test there as well. There is no CI allocated with it so I think it would be a nice change to help clean up our code.

neelmistry94 (approved):

macdrevx:
@neelmistry94 I do plan to move the Stress Test and Examples apps to the Apps workspace and directory as well, but those changes will come in separate PRs.

PR #187, opened 2021-03-16
Migrate Examples to Apps

### LINKED ISSUE #188: Run Examples tests fails: Process SceneKit document ... 34M_17.dae
https://app.circleci.com/pipelines/github/mapbox/mapbox-maps-ios/1114/workflows/2c53bb56-cfa9-4d98-9dad-952f0991d77a/jobs/7289/steps

```
CpResource /Users/distiller/project/Examples/Examples/All\ Examples/Sample\ Data/GradientLine.geojson /Users/distiller/project/build/Build/Products/Release-iphoneos/Examples.app/GradientLine.geojson (in target 'Examples' from project 'Examples')
    cd /Users/distiller/project/Examples
    builtin-copy -exclude .DS_Store -exclude CVS -exclude .svn -exclude .git -exclude .hg -resolve-src-symlinks /Users/distiller/project/Examples/Examples/All\ Examples/Sample\ Data/GradientLine.geojson /Users/distiller/project/build/Build/Products/Release-iphoneos/Examples.app

** TEST BUILD FAILED **


The following build commands failed:
	Process SceneKit document /Users/distiller/Downloads/34M_17.dae

```

macdrevx:
## Pull request checklist:

### Summary of changes

- Moves Examples app in to Apps directory and workspace
- Removes Examples app from MapboxMaps workspace
- Processes upgrade checks for Xcode 12.4
- Removes swiftlint build phase from Examples app
- Updates the Examples README
- Configures CI to be able to run Examples tests on Device Farm
- Upgrades Examples CI workflows to Xcode 12.4
- Examples changes:
    - Refactored Examples to disable push/pop animations during tests
    - Fixed a potential race condition by adding an object filter when waiting for finish notifications
    - Fixes #188


.circleci/config.yml
```diff
@@ -333,23 +332,95 @@ jobs:
           name: "Parsing xcresults for errors"
           command: |
             RESULTS=`find build/testruns -name '*.xcresult'`
-            pushd scripts/xcparty && swift build
-            popd
-            xargs ./scripts/xcparty/.build/debug/xcparty \<<< "$RESULTS" | tee build/testruns/failures.txt
+            xargs swift run --package-path scripts/xcparty \<<< "$RESULTS" | tee build/testruns/failures.txt
```
julianrex, line 335: TIL

Apps/Examples/Examples/Models/ExampleProtocol.swift
```diff
@@ -0,0 +1,28 @@
...
+        let resourceOptions = ResourceOptions(accessToken: accessToken)
+        return resourceOptions
+    }
+
+    public func finish() {
+        DispatchQueue.main.asyncAfter(deadline: .now() + 3.0) {
+            NotificationCenter.default.post(name: Example.finishNotificationName, object: self)
+        }
+    }
+}
```
julianrex, line 28: The previous code was added to ensure that if finish is called multiple times, that the notification is still only posted once. Is that still a concern with the refactor?

Apps/Examples/Examples/Models/Examples.swift
```diff
@@ -155,9 +112,6 @@ public struct Examples {
         Example(title: "Display buildings in 3D",
                 description: "Use extrusions to display buildings' height in 3D.",
                 type: BuildingExtrusionsExample.self),
-        Example(title: "Use OfflineRegionManager to download a region",
-                description: "Use the deprecated OfflineRegionManager to download regions for offline use.",
-                type: OfflineRegionManagerExample.self),
```
julianrex, line 160: Why is this deleted?

julianrex (approved):
A few questions, but LGTM

Apps/Examples/Examples/Models/ExampleProtocol.swift
macdrevx: It shouldn't matter anymore because the test is now looking for the notification from a specific object.

Apps/Examples/Examples/Models/Examples.swift
macdrevx: It was generating deprecation warnings, which failed the build because I reconfigured the Examples app to use the shared base.xcconfig which treats warnings as errors. Since the API is deprecated, we're still in beta, and there's a [note](https://github.com/mapbox/mapbox-maps-ios/pull/187/files#diff-df75f4ee000e948ae046c8e408d7bc38a1750908a947bdf531f3274f687d0d3eL52) warning that it will be removed in a future release, I decided the easiest path was to delete it.

I realized later that I could override the warnings as errors for the Release config using the Examples_testhost.xcconfig file, so if we do need to keep it, I'll update the PR to do that instead.

julianrex: Let's keep it if we can until there's an alternative.

Apps/Examples/Examples/Models/ExampleProtocol.swift
macdrevx: …so finish notifications from one test shouldn't impact any other tests (unless somehow the view controllers get allocated at the same memory address, in which case the [object comparison](https://developer.apple.com/documentation/foundation/nsnotification) would not help)

> The objects of a notification are compared using pointer equality for local notifications.

PR #189, opened 2021-03-17
Add layerId to AnnotationManager

julianrex:

## Pull request checklist:
 - [ ] ~Include before/after visuals or gifs if this PR includes visual changes.~
 - [X] Write tests for all new functionality. If tests were not written, please explain why.
 - [ ] ~Add example if relevant.~
 - [X] Add an entry inside this element for inclusion in the `mapbox-maps-ios` changelog: `<changelog>Added function to get layer identifier for an annotation type.</changelog>`.
 - [ ] ~Update the migration guide, API Docs, Markdown files - Readme, Developing, etc~

### Summary of changes

This PR adds a function to get the layer id for a particular annotation type. It also adds a WIP convenience function (`_moveLayer`) to `Style` to allow layers to be moved. These combined give developers a method to move the underlying annotation layer. 

The `_moveLayer` change is marked as `:nodoc:` and should be consider prone to change.

cc @Neel

### DIFF
--- a/Sources/MapboxMaps/Annotations/Configuration/AnnotationOptions.swift
+++ b/Sources/MapboxMaps/Annotations/Configuration/AnnotationOptions.swift
@@ -9,7 +9,7 @@ public struct AnnotationSourceOptions: Equatable {
 public struct AnnotationOptions: Equatable {
     /// The identifier of the layer above the annotation layers. Default is nil, which will position the layers
     /// above existing layers.
-    public var belowLayerId: String? = nil
+    public var belowLayerId: String?
 
     // TODO: Handle multiple layer Ids
 //    /// The layer identifier for the layer used by the AnnotationManager. Default is nil, meaning a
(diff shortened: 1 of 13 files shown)


julianrex:
cc @Chaoba @MaximAlien @jill-cardamon

Sources/MapboxMaps/Annotations/Configuration/AnnotationOptions.swift
```diff
@@ -0,0 +1,25 @@
...
+/// :nodoc:
+public struct AnnotationSourceOptions: Equatable {
+    // TODO: Add source and cluster options
+}
+
+/// Configuration options for the AnnotationManager
+public struct AnnotationOptions: Equatable {
+    /// The identifier of the layer above the annotation layers. Default is nil, which will position the layers
+    /// above existing layers.
+    public var belowLayerId: String?
```
neelmistry94, line 12: why are we only account for below and not above?

julianrex: This is to align with Android - but agreed, this should probably use `LayerPosition` for both platforms.

neelmistry94: can we ticket that out then. I'm concerned that we only provide the one option

neelmistry94 (approved):

Sources/MapboxMaps/Annotations/Configuration/AnnotationOptions.swift
julianrex: I changed layer position handling in a subsequent commit. Will ticket out difference.

neelmistry94 (approved):

PR #191, opened 2021-03-17
[Animations] Define Camera Animator

neelmistry94:

## Pull request checklist:
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-ios` changelog: `<changelog>Sets up a `CameraAnimator` to be consumed for animations which is backed by UIViewPropertyAnimator.</changelog>`.

### Summary of changes
- This PR is the first step in the CameraAnimations rewrite effort. This creates a new class that is backed by `UIViewPropertyAnimator` to expose a new way of writing camera animations. 
- It exposes a way for the `CameraManager` to make these `CameraAnimators`, while maintaining their lifecycle in the code.
- Schedules completion blocks for CameraAnimators appropriately

### DIFF
--- a/Sources/MapboxMaps/Foundation/BaseMapView.swift
+++ b/Sources/MapboxMaps/Foundation/BaseMapView.swift
@@ -72,7 +72,7 @@ open class ObserverConcrete: Observer {
     }
 }
 
-internal typealias AnimationCompletion = (UIViewAnimatingPosition) -> Void
+internal typealias PendingAnimationCompletion = (completion: AnimationCompletion, animatingPosition: UIViewAnimatingPosition)
 
 open class BaseMapView: UIView, MapClient, MBMMetalViewProvider, CameraViewDelegate {
 
@@ -86,7 +86,7 @@ open class BaseMapView: UIView, MapClient, MBMMetalViewProvider, CameraViewDeleg
     internal var resourceOptions: ResourceOptions?
 
     /// List of completion blocks that need to be completed by the displayLink
-    internal var pendingAnimatorCompletionBlocks: [(completion: AnimationCompletion, animatingPosition: UIViewAnimatingPosition)] = []
+    internal var pendingAnimatorCompletionBlocks: [PendingAnimationCompletion] = []
 
     public var needsDisplayRefresh: Bool = false
     public var dormant: Bool = false
@@ -339,9 +339,12 @@ open class BaseMapView: UIView, MapClient, MBMMetalViewProvider, CameraViewDeleg
             needsDisplayRefresh = false
             self.cameraView.update()
 
+            /// This executes the series of scheduled animation completion blocks and also removes them from the list
             while !pendingAnimatorCompletionBlocks.isEmpty {
-                let completionBlock = pendingAnimatorCompletionBlocks.removeFirst()
-                completionBlock.completion(completionBlock.animatingPosition)
+                let pendingCompletion = pendingAnimatorCompletionBlocks.removeFirst()
+                let completion = pendingCompletion.completion
+                let animatingPosition = pendingCompletion.animatingPosition
+                completion(animatingPosition)
             }
 
             self.displayCallback?()
--- a/Sources/MapboxMaps/Foundation/BaseMapView.swift
+++ b/Sources/MapboxMaps/Foundation/BaseMapView.swift
@@ -83,6 +83,9 @@ open class BaseMapView: UIView, MapClient, MBMMetalViewProvider, CameraViewDeleg
     /// Resource options for this map view
     internal var resourceOptions: ResourceOptions?
 
+    /// List of completion blocks that need to be completed by the displayLink
+    internal var pendingAnimatorCompletionBlocks: [(() -> Void)] = []
+
     public var needsDisplayRefresh: Bool = false
     public var dormant: Bool = false
     public var displayCallback: (() -> Void)?
@@ -98,8 +101,6 @@ open class BaseMapView: UIView, MapClient, MBMMetalViewProvider, CameraViewDeleg
             updateDisplayLinkPreferredFramesPerSecond()
         }
     }
-
-    internal var pendingAnimatorCompletionBlocks: [(() -> Void)] = []
 
     /// Returns the camera view managed by this object.
     public var cameraView: CameraView!
@@ -334,9 +335,13 @@ open class BaseMapView: UIView, MapClient, MBMMetalViewProvider, CameraViewDeleg
 
         if needsDisplayRefresh {
             needsDisplayRefresh = false
-            // execute completion blocks
-            // remove from list once executed
             self.cameraView.update()
+
+            while !pendingAnimatorCompletionBlocks.isEmpty {
+                let completionBlock = pendingAnimatorCompletionBlocks.removeFirst()
+                completionBlock()
+            }
+
             self.displayCallback?()
         }
     }
--- a/Sources/MapboxMaps/Foundation/BaseMapView.swift
+++ b/Sources/MapboxMaps/Foundation/BaseMapView.swift
@@ -99,6 +99,8 @@ open class BaseMapView: UIView, MapClient, MBMMetalViewProvider, CameraViewDeleg
         }
     }
 
+    internal var pendingAnimatorCompletionBlocks: [(() -> Void)] = []
+
     /// Returns the camera view managed by this object.
     public var cameraView: CameraView!
 
@@ -332,6 +334,8 @@ open class BaseMapView: UIView, MapClient, MBMMetalViewProvider, CameraViewDeleg
 
         if needsDisplayRefresh {
             needsDisplayRefresh = false
+            // execute completion blocks
+            // remove from list once executed
             self.cameraView.update()
             self.displayCallback?()
         }
--- a/Sources/MapboxMaps/Foundation/BaseMapView.swift
+++ b/Sources/MapboxMaps/Foundation/BaseMapView.swift
@@ -72,6 +72,8 @@ open class ObserverConcrete: Observer {
     }
 }
 
+internal typealias AnimationCompletion = (UIViewAnimatingPosition) -> Void
+
 open class BaseMapView: UIView, MapClient, MBMMetalViewProvider, CameraViewDelegate {
 
     /// The underlying renderer object responsible for rendering the map
@@ -84,7 +86,7 @@ open class BaseMapView: UIView, MapClient, MBMMetalViewProvider, CameraViewDeleg
     internal var resourceOptions: ResourceOptions?
 
     /// List of completion blocks that need to be completed by the displayLink
-    internal var pendingAnimatorCompletionBlocks: [(() -> Void)] = []
+    internal var pendingAnimatorCompletionBlocks: [(completion: AnimationCompletion, animatingPosition: UIViewAnimatingPosition)] = []
 
     public var needsDisplayRefresh: Bool = false
     public var dormant: Bool = false
@@ -339,7 +341,7 @@ open class BaseMapView: UIView, MapClient, MBMMetalViewProvider, CameraViewDeleg
 
             while !pendingAnimatorCompletionBlocks.isEmpty {
                 let completionBlock = pendingAnimatorCompletionBlocks.removeFirst()
-                completionBlock()
+                completionBlock.completion(completionBlock.animatingPosition)
             }
 
             self.displayCallback?()
--- a/Sources/MapboxMaps/Foundation/Camera/CameraAnimationProtocols.swift
+++ b/Sources/MapboxMaps/Foundation/Camera/CameraAnimationProtocols.swift
@@ -1,0 +1,50 @@
+import UIKit
+
+// MARK: Typealias
+public typealias AnimationCompletion = (UIViewAnimatingPosition) -> Void
+
+// MARK: CameraAnimatorDelegate Protocol
+internal protocol CameraAnimatorDelegate {
+
+    /**
+    This delegate function notifies that the completion block needs to be scheduled
+
+    - Parameter animator: The current animator that this delegate function is being called from
+    - Parameter completion: The completion block that needs to be scheduled
+    - Parameter animatingPosition  The position of the animation needed for the closure
+    */
+    func schedulePendingCompletion(forAnimator animator: CameraAnimator,
+                                   completion: @escaping AnimationCompletion,
+                                   animatingPosition: UIViewAnimatingPosition)
+
+    /**
+    This delegate function notifies that the animation is finished
+
+    - Parameter animator: The current animator that this delegate function is being called from
+    */
+    func animatorIsFinished(forAnimator animator: CameraAnimator)
+
+}
+
+// MARK: AnimationOwnerProtocol
+internal protocol AnimationOwnerProtocol {
+    var id: String { get }
+}
+
+// MARK: AnimationOwner Enum
+public enum AnimationOwner: AnimationOwnerProtocol {
+    case gestures
+    case unspecified
+    case custom(id: String)
+
+    var id: String {
+        switch self {
+        case .gestures:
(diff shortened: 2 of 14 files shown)


Sources/MapboxMaps/Foundation/Camera/CameraAnimator.swift
```diff
@@ -0,0 +1,88 @@
...
+// MARK: CameraAnimator Class
+public class CameraAnimator: NSObject {
+
+    // MARK: Stored Properties
+
+    /// Instance of the property animator that will run animations.
+    private let propertyAnimator: UIViewPropertyAnimator
+
+    /// Delegate that conforms to `CameraAnimatorDelegate`.
+    private var delegate: CameraAnimatorDelegate?
```
nishant-karajgikar, line 12: should be weak

Sources/MapboxMaps/Foundation/Camera/CameraManager.swift
```diff
@@ -48,6 +48,9 @@ public class CameraManager {
         }
     }
 
+    /// Pointer array for holding camera animators
```
nishant-karajgikar, line 51: It's not a pointer array anymore right?

Sources/MapboxMaps/Foundation/Camera/CameraAnimator.swift
```diff
@@ -0,0 +1,88 @@
...
+    public func addAnimations(_ animations: @escaping () -> Void, delayFactor: CGFloat) {
+        propertyAnimator.addAnimations(animations, delayFactor: delayFactor)
+    }
+
+    /// Add animations block to the animator.
+    public func addAnimations(_ animations: @escaping () -> Void) {
+        propertyAnimator.addAnimations(animations)
+    }
+
+    /// Add a completion block to the animator. This function will notifiy the delegate to schedule the completion block.
```
nishant-karajgikar, line 77: > This function will notifiy the delegate to schedule the completion block.
 
This is an implementation detail -- let's not expose it.

Sources/MapboxMaps/Foundation/Camera/CameraManager.swift
```diff
@@ -691,6 +694,134 @@ public class CameraManager {
...
+    // MARK: Animator Functions
+
+    /**
+     Convenience to create a `CameraAnimation` and will add it to a List of `CameraAnimators` to track the lifecycle of that animation.
+
+     - Parameter duration: The duration of the animation, in seconds.
+     - Parameter timingParameters: The object providing the timing information. This object must adopt the `UITimingCurveProvider` protocol.
+     - Parameter animationOwner: The `id` of what owns this animation.
+     - Returns `CameraAnimator`: A class that represents an animator with the provided configuration.
+     */
```
nishant-karajgikar, line 708: nit: Could you use the auto-generated documentation format here? Whatever is outputted by holding down `Cmd+Shift+\`

nishant-karajgikar: Same for all the multi-line docs below.

Sources/MapboxMaps/Foundation/Camera/CameraManager.swift
```diff
@@ -691,6 +694,134 @@ public class CameraManager {
...
+    func makeCameraAnimator(duration: TimeInterval,
```
nishant-karajgikar, line 709: Each of these `makeCameraAnimator*` methods should be public

Sources/MapboxMaps/Foundation/Camera/CameraManager.swift
```diff
@@ -691,6 +694,134 @@ public class CameraManager {
...
+                                                                             delay: delay,
+                                                                             options: options,
+                                                                             animations: animations,
+                                                                             completion: nil)
+
+        let cameraAnimator = CameraAnimator(delegate: self, propertyAnimator: runningAnimator, owner: animationOwner)
+
+        runningAnimator.addCompletion({ [weak self] animatingPosition in
+            guard let completion = completion,
+                  let validSelf = self
```
nishant-karajgikar, line 804: nit: You can use `guard let self = self` to avoid using a new variable.

Tests/MapboxMapsTests/Foundation/Camera/CameraAnimatorMocks.swift
```diff
@@ -0,0 +1,21 @@
+import XCTest
+
+#if canImport(MapboxMaps)
+@testable import MapboxMaps
+#else
+@testable import MapboxMapsFoundation
+#endif
+
+public class CameraAnimatorMock: CameraAnimatorDelegate {
```
nishant-karajgikar, line 9: Look into using https://github.com/mapbox/mapbox-maps-ios/blob/main/Tests/MapboxMapsTests/Helpers/Stub.swift instead of creating this.

Sources/MapboxMaps/Foundation/BaseMapView.swift
```diff
@@ -333,6 +338,15 @@ open class BaseMapView: UIView, MapClient, MBMMetalViewProvider, CameraViewDeleg
         if needsDisplayRefresh {
             needsDisplayRefresh = false
             self.cameraView.update()
+
+            /// This executes the series of scheduled animation completion blocks and also removes them from the list
+            while !pendingAnimatorCompletionBlocks.isEmpty {
+                let pendingCompletion = pendingAnimatorCompletionBlocks.removeFirst()
```
nishant-karajgikar, line 344: why are you only removing the first? We should be executing and and all pending completions here.

Sources/MapboxMaps/Foundation/Camera/CameraAnimationProtocols.swift
```diff
@@ -0,0 +1,50 @@
+import UIKit
```
nishant-karajgikar, line 1: Let's split this file up?

nishant-karajgikar: One file for the `CameraAnimatorDelegate` and the other for the `AnimationOwner` + `AnimationOwnerProtocol`

Sources/MapboxMaps/Foundation/Camera/CameraAnimator.swift
```diff
@@ -0,0 +1,88 @@
...
+    // MARK: Stored Properties
+
+    /// Instance of the property animator that will run animations.
+    private let propertyAnimator: UIViewPropertyAnimator
+
+    /// Delegate that conforms to `CameraAnimatorDelegate`.
+    private var delegate: CameraAnimatorDelegate?
+
+    /// The ID of the owner of this `CameraAnimator`.
+    internal var owner: AnimationOwnerProtocol
```
nishant-karajgikar, line 15: should this be weak?

Sources/MapboxMaps/Foundation/Camera/CameraManager.swift
```diff
@@ -691,6 +694,134 @@ public class CameraManager {
...
+                            animationOwner: AnimationOwnerProtocol = AnimationOwner.unspecified,
+                            animations: (() -> Void)? = nil) -> CameraAnimator {
+        let propertyAnimator = UIViewPropertyAnimator(duration: duration, controlPoint1: point1, controlPoint2: point2, animations: animations)
+        let cameraAnimator = CameraAnimator(delegate: self, propertyAnimator: propertyAnimator, owner: animationOwner)
+        cameraAnimators.add(cameraAnimator)
+        return cameraAnimator
+    }
+
+    /**
+     Convenience to create a `CameraAnimation` and will add it to a List of `CameraAnimators` to track the lifecycle of that animation
```
nishant-karajgikar, line 759: This should be "Convenience to create a `CameraAnimator`"

Sources/MapboxMaps/Foundation/Camera/CameraManager.swift
```diff
@@ -691,6 +694,134 @@ public class CameraManager {
...
+
+     - Parameter duration: The duration of the animation, in seconds.
+     - Parameter timingParameters: The type of `AnimationCurve` for the animation.
+     - Parameter animationOwner: The `id` of what owns this animation.
```
nishant-karajgikar, line 763: What do you mean by `id`? This is just a conformer to the `AnimationOwnerProtocol` right?

Sources/MapboxMaps/Foundation/Camera/CameraManager.swift
neelmistry94: I did that before I pushed

Sources/MapboxMaps/Foundation/BaseMapView.swift
neelmistry94: removeFirst will remove the first out of the array so it utilizes a queue based system. This way I am maintaining order while also removing from the array when I am executing the function.

The while loop will continue to execute all blocks in the array until it is empty

Sources/MapboxMaps/Foundation/Camera/CameraAnimator.swift
neelmistry94: Its an enum so I don't think so

Sources/MapboxMaps/Foundation/Camera/CameraManager.swift
neelmistry94: correct, but the protocol when used is representing a way to ID the animator

nishant-karajgikar: The `id` isn't all the animationOwner could have -- it just happens to have that at this time. That will change in the future.

Sources/MapboxMaps/Foundation/Camera/CameraAnimator.swift
nishant-karajgikar: we're passing in an enum but what happens if a developer conforms to it themselves and passes in an object?

Sources/MapboxMaps/Foundation/Camera/CameraManager.swift
```diff
@@ -691,6 +694,134 @@ public class CameraManager {
...
+                  let self = self
+            else { return }
+
+            self.schedulePendingCompletion(forAnimator: cameraAnimator, completion: completion, animatingPosition: animatingPosition)
+        })
+        cameraAnimators.add(cameraAnimator)
+        return cameraAnimator
+    }
+
+    // MARK: Delegate Functions
```
nishant-karajgikar, line 813: please provide more detail about which delegate functions

Sources/MapboxMaps/Foundation/BaseMapView.swift
nishant-karajgikar: ah i missed the `while`.

Sources/MapboxMaps/Foundation/Camera/CameraAnimator.swift
```diff
@@ -0,0 +1,88 @@
...
+    public func startAnimation(afterDelay delay: TimeInterval) {
+        propertyAnimator.startAnimation(afterDelay: delay)
+    }
+
+    /// Pauses the animation
+    public func pauseAnimation() {
+        propertyAnimator.pauseAnimation()
+    }
+
+    /// Stops the animation and notifies the delegate that the animation finished.
```
nishant-karajgikar, line 60: `notifies the delegate that the animation finished` : This isn't relevant for an end developer.

Sources/MapboxMaps/Foundation/Camera/CameraAnimator.swift
neelmistry94: Yes that would possibly cause a retain cycle, if not handled correctly.

neelmistry94: One complication here is that in order for that to be weak, then the protocol would need to be `:class` so then `AnimationOwner` can't be an enum

nishant-karajgikar: Hmm, maybe we get rid of the protocol then and just use the enum directly.

Sources/MapboxMaps/Foundation/Camera/CameraManager.swift
nishant-karajgikar: Could we reword the "MARK: Delegate functions" to "CameraAnimatorDelegate functions"

Sources/MapboxMaps/Foundation/Camera/CameraManager.swift
```diff
@@ -691,6 +694,134 @@ public class CameraManager {
...
+    func schedulePendingCompletion(forAnimator animator: CameraAnimator, completion: @escaping AnimationCompletion, animatingPosition: UIViewAnimatingPosition) {
+        guard let mapView = mapView else { return }
+        mapView.pendingAnimatorCompletionBlocks.append((completion, animatingPosition))
+        cameraAnimators.remove(animator)
+    }
+
+    func animatorIsFinished(forAnimator animator: CameraAnimator) {
+        cameraAnimators.remove(animator)
```
nishant-karajgikar, line 821: Hmm, do we need `animatorIsFInished` at all? Seems like `schedulePendingCompletion` takes care of it all?

Tests/MapboxMapsTests/Foundation/Camera/CameraAnimatorMock.swift
```diff
@@ -0,0 +1,24 @@
+import XCTest
+
+#if canImport(MapboxMaps)
+@testable import MapboxMaps
+#else
+@testable import MapboxMapsFoundation
+#endif
+
+final class CameraAnimatorMock: CameraAnimatorDelegate {
```
nishant-karajgikar, line 9: Shouldn't this be `CameraAnimatorDelegateMock` or `CameraAnimatorDelegateStub` ?

Sources/MapboxMaps/Foundation/Camera/CameraAnimator.swift
neelmistry94: I think that makes a lot of sense, no reason to expose that protocol right now.

Sources/MapboxMaps/Foundation/Camera/CameraManager.swift
neelmistry94: yeah my bad, forgot this one

Sources/MapboxMaps/Foundation/Camera/CameraManager.swift
neelmistry94: What about the case that StopAnimation is manually called and that would trigger these changes to occur. Which is when `animatorIsFinished` gets called

Tests/MapboxMapsTests/Foundation/Camera/CameraAnimatorMock.swift
neelmistry94: yes

Sources/MapboxMaps/Foundation/Camera/CameraManager.swift
nishant-karajgikar: Hmm but doesn't `schedulePendingCompletion` get called automatically in that case?

nishant-karajgikar: Ah wait -- an animation might not have a completion. Never mind, this is necessary.

nishant-karajgikar: Let's just change it so as to check if the animator exists in the hashTable.. If it does, remove it. Else, do nothing.

Sources/MapboxMaps/Foundation/Camera/CameraAnimator.swift
nishant-karajgikar: sounds good, let's remove the protocol for now.

nishant-karajgikar (approved):

PR #208, opened 2021-03-24
[Camera Animators v2] Update gestures to use drag API

nishant-karajgikar:

## Pull request checklist:

### Summary of changes

- Makes the `GestureManager` call into the drag API when the map is panned.
- Deletes the now unused `moveBy` and `shiftCenterCoordinate` methods
- Makes the `propertyAnimator` held by `CameraAnimator` optional -- and makes it so that a `deinit` of a `CameraAnimator` 
   results in the `propertyAnimator` being nil'ed out.
- Deletes unused `TransitionState` and associated notification
- Ensures that gestures call into `cancelTransition` when invoked
- Reworks `cancelTransitions` to operate on list of `cameraAnimators`
- Ensures that the `cameraView` wraps center coordinate longitudes to within [-180, +180] range before calling `jumpTo`.
- Ensuring that internal infrastructure never removes a `CameraAnimator` from the list of `cameraAnimators` directly -- instead we rely on ARC to clean up the `cameraAnimators` hashTable for us.

https://user-images.githubusercontent.com/6844889/112365244-d07cc200-8cad-11eb-9691-0b9acccd7180.mov

### DIFF
--- a/Apps/DebugApp/DebugApp/DebugViewController.swift
+++ b/Apps/DebugApp/DebugApp/DebugViewController.swift
@@ -80,13 +80,14 @@ public class DebugViewController: UIViewController {
         mapView.on(.mapLoaded) { (event) in
             print("The map has finished loading... Event = \(event)")
             
-//            let initialCenter = CLLocationCoordinate2D(latitude: 39.01305735102963, longitude: -77.01570412528032)
-//            self.mapView.cameraManager.setCamera(centerCoordinate: initialCenter, zoom: 12)
-//            
-//            self.runningAnimator = self.mapView.cameraManager.makeRunningCameraAnimator(duration: 10, delay: 2, animations: {
-//                
-//                self.mapView.cameraManager.setCamera(centerCoordinate: CLLocationCoordinate2D(latitude: 36.0893334370578, longitude: -78.06549948618996), zoom: 12)
-//            })
+            let initialCenter = CLLocationCoordinate2D(latitude: 39.01305735102963, longitude: -77.01570412528032)
+            self.mapView.cameraManager.setCamera(centerCoordinate: initialCenter, zoom: 12)
+            
+            self.runningAnimator = self.mapView.cameraManager.makeCameraAnimator(duration: 10, curve: .linear) {
+                self.mapView.cameraManager.setCamera(centerCoordinate: CLLocationCoordinate2D(latitude: 36.0893334370578, longitude: -78.06549948618996), zoom: 12)
+            }
+            
+            self.runningAnimator!.startAnimation(afterDelay: 2)
         }
 
         /**
--- a/Apps/DebugApp/DebugApp/DebugViewController.swift
+++ b/Apps/DebugApp/DebugApp/DebugViewController.swift
@@ -79,14 +79,14 @@ public class DebugViewController: UIViewController {
          */
         mapView.on(.mapLoaded) { (event) in
             print("The map has finished loading... Event = \(event)")
-            
+
             let initialCenter = CLLocationCoordinate2D(latitude: 39.01305735102963, longitude: -77.01570412528032)
             self.mapView.cameraManager.setCamera(centerCoordinate: initialCenter, zoom: 12)
-            
+
             self.runningAnimator = self.mapView.cameraManager.makeCameraAnimator(duration: 10, curve: .linear) {
                 self.mapView.cameraManager.setCamera(centerCoordinate: CLLocationCoordinate2D(latitude: 36.0893334370578, longitude: -78.06549948618996), zoom: 12)
             }
-            
+
             self.runningAnimator!.startAnimation(afterDelay: 2)
         }
 
--- a/Apps/DebugApp/DebugApp/DebugViewController.swift
+++ b/Apps/DebugApp/DebugApp/DebugViewController.swift
@@ -11,6 +11,7 @@ import Turf
 public class DebugViewController: UIViewController {
 
     internal var mapView: MapView!
+    internal var runningAnimator: CameraAnimator?
 
     var resourceOptions: ResourceOptions {
         guard let accessToken = AccountManager.shared.accessToken else {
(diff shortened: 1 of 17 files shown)


Apps/DebugApp/DebugApp/DebugViewController.swift
```diff
@@ -11,6 +11,7 @@ import Turf
 public class DebugViewController: UIViewController {
 
     internal var mapView: MapView!
+    internal var runningAnimator: CameraAnimator?
```
neelmistry94, line 14: I know this is debug code, but since you are using `makeCameraAnimator` and not `makeRunningCameraAnimator` let's call this `cameraAnimator`

Or you can just remove all changes to the debugVC as a whole

nishant-karajgikar: we'll be doing that before merging to `main`

Sources/MapboxMaps/Foundation/Camera/CameraAnimator.swift
```diff
@@ -6,7 +6,7 @@ public class CameraAnimator: NSObject {
     // MARK: Stored Properties
 
     /// Instance of the property animator that will run animations.
-    private let propertyAnimator: UIViewPropertyAnimator
+    private var propertyAnimator: UIViewPropertyAnimator?
```
neelmistry94, line 9: Why is this now an optional var?

This class as a whole would be useless without the property animator and so I think making it optional makes no sense

Sources/MapboxMaps/Foundation/Camera/CameraView.swift
```diff
@@ -233,3 +233,16 @@ public class CameraView: UIView {
         }
     }
 }
+
+fileprivate extension CameraOptions {
+
+    func wrap() -> CameraOptions {
```
neelmistry94, line 239: Can we add some docs about the purpose/use case of this function

Sources/MapboxMaps/Foundation/Camera/CameraAnimator.swift
nishant-karajgikar: From the PR description:
> Makes the propertyAnimator held by CameraAnimator optional -- and makes it so that a deinit of a CameraAnimator
results in the propertyAnimator being nil'ed out.

Sources/MapboxMaps/Gestures/GestureManager+GestureHandlerDelegate.swift
```diff
@@ -0,0 +1,186 @@
...
+        if numberOfTaps == 2 && numberOfTouches == 2 {
+            cameraManager.setCamera(to: CameraOptions(zoom: mapView.zoom - 1.0),
+                                    animated: true,
+                                    duration: 0.3,
+                                    completion: nil)
+        }
+    }
+
+    internal func panBegan(at point: CGPoint) {
+        try! cameraManager.mapView?.__map.dragStart(forPoint: point.screenCoordinate)
```
neelmistry94, line 31: I believe this is the use of the drag API. I'm concerned that we are not handling the force try. I'm new to Drag API so I think we should really catch this error here.

Sources/MapboxMaps/Foundation/Camera/CameraAnimator.swift
neelmistry94: so when it was a `let` before, the deinit of `CameraAnimator` doesn't automatically nil the `propertyAnimator`. 

What if we make property animator a `weak let` ?

Sources/MapboxMaps/Gestures/GestureManager+GestureHandlerDelegate.swift
nishant-karajgikar: This isn't actually a throwing function -- this is a result of the fact that the drag API exists on the map  object. All APIs on the `map` are currently marked as `throws`. This will be changing in the near future. We'll handle this case along with the other instances of `try!`.

Sources/MapboxMaps/Gestures/GestureManager+GestureHandlerDelegate.swift
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    // MapView has been panned
+    internal func panned(from startPoint: CGPoint, to endPoint: CGPoint) {
+
+        if let cameraOptions = try! cameraManager.mapView?.__map.getDragCameraOptionsFor(fromPoint: startPoint.screenCoordinate, toPoint: endPoint.screenCoordinate) {
```
neelmistry94, line 37: same as above

Sources/MapboxMaps/Gestures/GestureManager+GestureHandlerDelegate.swift
```diff
@@ -0,0 +1,186 @@
...
+                                    completion: nil)
+
+        }
+    }
+
+    // Pan has ended on the MapView with a residual `offset`
+    func panEnded(at endPoint: CGPoint, shouldDriftTo driftEndPoint: CGPoint) {
+
+        if endPoint != driftEndPoint,
+           let driftCameraOptions = try! cameraManager.mapView?.__map.getDragCameraOptionsFor(fromPoint: endPoint.screenCoordinate, toPoint: driftEndPoint.screenCoordinate) {
```
neelmistry94, line 51: same as above

Sources/MapboxMaps/Gestures/GestureManager+GestureHandlerDelegate.swift
nishant-karajgikar: responded above

Sources/MapboxMaps/Gestures/GestureManager+GestureHandlerDelegate.swift
nishant-karajgikar: responded above

Sources/MapboxMaps/Gestures/GestureManager+GestureHandlerDelegate.swift
```diff
@@ -0,0 +1,186 @@
...
+
+            cameraManager.setCamera(to: driftCameraOptions,
+                                    animated: true,
+                                    duration: Double(UIScrollView.DecelerationRate.normal.rawValue),
```
neelmistry94, line 55: We should probably use `cameraManager.mapCameraOptions.decelerationRate` instead of scroll view in case if the app developer provided a specific deceleration rate

nishant-karajgikar: yep good point!

Sources/MapboxMaps/Gestures/GestureManager+GestureHandlerDelegate.swift
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    internal func pitchChanged(newPitch: CGFloat) {
+        cameraManager.setCamera(to: CameraOptions(pitch: newPitch),
+                                animated: false,
+                                duration: 0,
+                                completion: nil)
+    }
+
+    internal func pitchEnded() {
```
neelmistry94, line 184: empty function, let's at least a "not yet implemented" statement or remove it entirely

Sources/MapboxMaps/Gestures/GestureManager+GestureHandlerDelegate.swift
neelmistry94: gotcha thanks for the clarification

Sources/MapboxMaps/Gestures/GestureManager+GestureHandlerDelegate.swift
nishant-karajgikar: this existed previously. I've just moved this code out of the `GestureManager.swift` file

Sources/MapboxMaps/Foundation/Camera/CameraAnimator.swift
nishant-karajgikar: Then it will fall out of scope immediately.

nishant-karajgikar: > so when it was a let before, the deinit of CameraAnimator doesn't automatically nil the propertyAnimator.

Not in all cases.

neelmistry94: Could we be specific in the case then? I see the solution and why it works, but I'm concerned that someone can technically make a `CameraAnimator` with no property animator and that feels wrong. I don't think we should allow that

neelmistry94 (approved):

PR #219, opened 2021-03-25
Fix Puck3D Deinitialization

macdrevx:
## Pull request checklist:

### Summary of changes

- Updates the pre-build action in the MapboxMaps scheme to work when using Apps.xcworkspace
- Adds an integration test to exercise deinitialization of a map view configured with a 3D puck
- Extracts `func mapboxAccessToken() throws -> String` into an extension on `XCTestCase`
- Updates `Puck3D` so that it no longer keeps a strong reference to `Style`. This strong reference led to an issue in which the `Style`'s `StyleManager` had been deallocated prior to when `Puck3D.deinit` was invoked. This resulted in a crash because `Puck3D.deinit` invokes `Style.removeStyleLayer(forLayerId:)` which accesses `Style.styleManager` which is a weak, implicitly unwrapped optional (IUO). Since the object to which `styleManager` had referred had already been deallocated, the reference was `nil`, but since the type is an IUO and the call site did not use optional chaining, a crash occurred.

### Future improvements

This change is at best a stop-gap solution. A more thorough fix should be completed in the future which:

* Audits the codebase for weak IUOs and either converts them to `unowned` (if we're really sure) or, more likely, converts them to weak optionals (that are not implicitly unwrapped)
* More generally audits the object ownership and reference graph and performs changes to achieve simpler, less error-prone design

### DIFF
--- a/.swiftpm/xcode/xcshareddata/xcschemes/MapboxMaps.xcscheme
+++ b/.swiftpm/xcode/xcshareddata/xcschemes/MapboxMaps.xcscheme
@@ -10,7 +10,7 @@
             ActionType = "Xcode.IDEStandardExecutionActionsCore.ExecutionActionType.ShellScriptAction">
             <ActionContent
                title = "Run Script"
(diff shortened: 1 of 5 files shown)


jmkiley (approved):
Thanks for fixing!

PR #231, opened 2021-03-29
Make addLayer/addSource not generic.

julianrex:

## Pull request checklist:

### Summary of changes

This PR removes the generic parameter from `addLayer` and `addSource`, instead moving the functionality to a new `StyleEncodable`.

I have not done the same for `getLayer/getSource`, but a similar approach (and a `StyleDecodable`) would work here (though this may be something we don't want to do).

### User impact (optional)

Although these functions have been made non-generic, there is no change to user code.

### DIFF
--- a/Sources/MapboxMaps/Style/Style.swift
+++ b/Sources/MapboxMaps/Style/Style.swift
@@ -40,14 +40,11 @@ public class Style {
                 success case. Else, returns a `LayerError` in the `Result` failure case.
      */
     @discardableResult
-    public func addLayer<T: Layer>(layer: T, layerPosition: LayerPosition? = nil) -> Result<Bool, LayerError> {
-
+    public func addLayer(layer: Layer, layerPosition: LayerPosition? = nil) -> Result<Bool, LayerError> {
         // Attempt to encode the provided layer into JSON and apply it to the map
         do {
-            let layerData = try JSONEncoder().encode(layer)
-            //swiftlint:disable force_cast
-            let layerJSON = try JSONSerialization.jsonObject(with: layerData) as! [String: AnyObject]
-            let expected = try! styleManager.addStyleLayer(forProperties: layerJSON, layerPosition: layerPosition)
+            let layerJSON = try layer.jsonObject()
+            let expected = try styleManager.addStyleLayer(forProperties: layerJSON, layerPosition: layerPosition)
 
             if expected.isError() {
                 return .failure(.addStyleLayerFailed(expected.error as? String))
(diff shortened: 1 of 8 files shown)


Sources/MapboxMaps/Style/Style.swift
```diff
@@ -40,14 +40,11 @@ public class Style {
                 success case. Else, returns a `LayerError` in the `Result` failure case.
      */
     @discardableResult
-    public func addLayer<T: Layer>(layer: T, layerPosition: LayerPosition? = nil) -> Result<Bool, LayerError> {
-
+    public func addLayer(layer: Layer, layerPosition: LayerPosition? = nil) -> Result<Bool, LayerError> {
```
macdrevx, line 43: I want to make these methods throw instead of returning `Result`. Not for this PR though.

Only downside is the errors would be untyped, but I'm not convinced that's a big deal. It would clean up the error handling significantly.

Sources/MapboxMaps/Style/StyleErrors.swift
```diff
@@ -26,7 +30,7 @@ public enum SourceError: Error {
 /// Error enum for all layer-related errors
 public enum LayerError: Error {
     /// The layer provided to the map in `addLayer()` could not be encoded
-    case layerEncodingFailed(Error)
+    case layerEncodingFailed(Error?)
```
macdrevx, line 33: Why is this change needed?

Sources/MapboxMaps/Style/StyleEncodable.swift
```diff
@@ -0,0 +1,19 @@
...
+    func jsonObject() throws -> [String: AnyObject]
+}
+
+public extension StyleEncodable where Self: Encodable {
+    /// Given an Encodable object return the JSON dictionary representation
+    /// - Throws: Errors occurring during encoding, or `StyleEncodingError.invalidJSONObject`
+    /// - Returns: A JSON dictionary representing the object.
+    func jsonObject() throws -> [String: AnyObject] {
+        let data = try JSONEncoder().encode(self)
+        guard let jsonObject = try JSONSerialization.jsonObject(with: data) as? [String: AnyObject] else {
```
macdrevx, line 14: @nishant-karajgikar are we still planning to refactor this to avoid the serialize/deserialize dance?

macdrevx (approved):
Nice change

Sources/MapboxMaps/Style/Style.swift
julianrex: Ticketed.

Sources/MapboxMaps/Style/StyleErrors.swift
julianrex: Ah, test code left-over. Let's revisit as we look at throwing vs Results. I suspect our error cases will change too.

PR #248, opened 2021-04-01
[style] Decode empty json if needed, regenerate sources + layers

nishant-karajgikar:

## Pull request checklist:
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-ios` changelog: `<changelog>
 - Fix issue where empty json objects in expressions failed to decoded. 
 - Set correct raw values for style enum properties.
 </changelog>`

### Summary of changes

1. Decodes an empty JSON if needed at the end of an expression decoding pass. This fixes a subtle bug that resulted in `Value` types not being decoded correctly if there was an empty JSON part of the expression.

1. Regenerates sources + layers to incorporate a change to set raw values for enum properties.

1. Remove `ExpressionOption` in favor of new `Expression.Option` which encapsulates the logic better.


jmkiley (approved):
LGTM. Can we add a changelog entry, since this PR should fix issues around some style layers failing to decode?

Sources/MapboxMaps/Style/Types/Expression.swift
```diff
@@ -245,6 +245,13 @@ public struct Expression: Codable, CustomStringConvertible, Equatable {
 
             if let validArray = try? container.decode([Double].self) {
                 self = .array(validArray)
+                return
+            }
+
+            // Attempt to decode an "empty" json if all else fails
```
neelmistry94, line 251: This comment seems like that in decoding, an empty json will be the fail all case. 

I don't think a comment is needed. Or maybe say that this is to account for empty json's and provide an example.

neelmistry94 (approved):

neelmistry94 (approved):

PR #249, opened 2021-04-06
Switch to using _layer in test

julianrex:
PR adds a `type` property to `LayerType` and updates the layer test to use `_layer` accordingly.


Sources/MapboxMaps/Style/Layers.swift
```diff
@@ -39,6 +39,36 @@ public enum LayerType: String, Codable {
 
     /// Layer used for a 3D model
     case model = "model"
+
+    /// The associated Swift struct type
+    public var type: Layer.Type {
```
macdrevx, line 44:
```suggestion
    public var layerType: Layer.Type {
```

PR #255, opened 2021-04-12
Disable Update Manifest Script

neelmistry94:
### Summary of changes

This PR removes a step in our release CI flow because it is now unnecessary. It will be replaced by a version checker script


.circleci/config.yml
```diff
@@ -497,10 +497,10 @@ jobs:
...
           command: ./scripts/release/create-api-downloads-pr.sh mobile-maps-ios ${VERSION}
           when: always
-      - run:
-          name: Update Package Manifest & Podspec
-          command: ./scripts/release/update-dependency-manager-manifests.sh ${VERSION}
-          when: on_success
+      # - run:
+      #     name: Update Package Manifest & Podspec
+      #     command: ./scripts/release/update-dependency-manager-manifests.sh ${VERSION}
+      #     when: on_success
```
macdrevx, line 503: Can we delete this & the script to which it refers?

macdrevx (approved):

PR #258, opened 2021-04-12
Fix an issue where updating compass visibility doesn't take effect

jmkiley:

## Pull request checklist:

### Summary of changes

This PR:
- Renames `OrnamentOptions.compassVisiblity` to `OrnamentOptions.compassVisibility`.
- Fixes an issue where updating the visibility of the compass doesn't take effect. 
   - This seems to have been because the compass' view property was `nil` when we tried to remove it from its superview. My current approach is to check whether the map view contains a scale bar or compass subview, then set that as the ornament's view.  
- Allows scale bar visibility to be updated. 
- `MapboxCompassOrnamentView.currentBearing`, only update compass visibility if the old value is different from the new value.
### User impact (optional)

- The rename of `OrnamentOptions.compassVisibility` impacts users currently using that property.

### DIFF
--- a/Sources/MapboxMaps/Ornaments/ScaleBar/MapboxScaleBarOrnamentView.swift
+++ b/Sources/MapboxMaps/Ornaments/ScaleBar/MapboxScaleBarOrnamentView.swift
@@ -11,12 +11,6 @@ internal class MapboxScaleBarOrnamentView: UIView {
 
     // MARK: - Properties
 
-    internal var visibility: OrnamentVisibility = .adaptive {
-        didSet {
-            updateVisibility()
-        }
-    }
-
     internal var metersPerPoint: CLLocationDistance = 1 {
         didSet {
             guard metersPerPoint != oldValue else {
@@ -349,9 +343,7 @@ internal class MapboxScaleBarOrnamentView: UIView {
     }
 
     private func updateVisibility() {
-        if visibility == .hidden {
-            self.alpha = 0
-        } else {
+        if !isHidden {
             let maximumDistance: CLLocationDistance = Double(maximumWidth) * unitsPerPoint
             let allowedDistance = isMetricLocale ?
                                   Constants.metricTable.last!.distance : Constants.imperialTable.last!.distance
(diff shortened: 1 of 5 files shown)


Sources/MapboxMaps/Ornaments/ScaleBar/MapboxScaleBarOrnamentView.swift
```diff
@@ -343,19 +349,23 @@ internal class MapboxScaleBarOrnamentView: UIView {
...
-        if alpha != self.alpha {
-            UIView.animate(withDuration: 0.2,
-                           delay: 0,
-                           options: .beginFromCurrentState,
-                           animations: {
-                            self.alpha = alpha
-            },
-                           completion: nil)
+        if visibility == .hidden {
+            self.alpha = 0
```
neelmistry94, line 353: `MapboxScaleBarOrnamentView` is of type `UIView`. Use the `isHidden` property instead

Sources/MapboxMaps/Ornaments/ScaleBar/MapboxScaleBarOrnamentView.swift
```diff
@@ -343,19 +349,23 @@ internal class MapboxScaleBarOrnamentView: UIView {
...
+        } else {
+            let maximumDistance: CLLocationDistance = Double(maximumWidth) * unitsPerPoint
```
neelmistry94, line 355: With the above change, alter the `isHidden` property to be `false`

neelmistry94 (changes requested):

macdrevx:
@jmkiley you've definitely identified the root cause of this bug (`Ornament.view` is never set). In terms of fixing it, I think we should take a step back as a team and reevaluate the architecture of the ornaments implementation. We may decide that it'd be better to make a larger change to simplify things internally instead of proceeding with trying to patch this specific issue.

neelmistry94 (approved):

macdrevx:
Closing in favor of #288

PR #265, opened 2021-04-14
Encapsulate MapboxCoreMaps protocol conformances

macdrevx:
## Pull request checklist:
 - [x] Update changelog

### Summary of changes

Implement Core interfaces via delegation

- [breaking] BaseMapView no longer conforms to MapClient or MBMMetalViewProvider.
  Instead, it owns and acts as the delegate for a DelegatingMapClient
  which implements the protocols instead. This allows the required
  methods to be internal instead of public.
- [breaking] Snapshotter no longer conforms to Observer
- ObserverConcrete was refactored into DelegatingObserver and is now
  used by both BaseMapView and Snapshotter.

Misc ACL Changes

- BaseMapView setters for __map, metalView, resourceOptions,
  mapInitOptionsDataSource, and cameraView are now private
- BaseMapView properties needsDisplayRefresh, dormant, displayCallback,
  and styleURI__ are now private
- Snapshotter.eventHandlers is now private

Other Refactoring

- The metal support checks in BaseMapView.commonInit were factored out
  into a separate helper method to avoid the need for a local variable
- In awakeFromNib(), the default styleURI is accessed via StyleURI
  rather than via a string literal.
- PreferredFPS moved to separate file
- Adds test xib to Xcode project

### DIFF
--- a/Sources/MapboxMaps/Foundation/BaseMapView.swift
+++ b/Sources/MapboxMaps/Foundation/BaseMapView.swift
@@ -5,98 +5,38 @@ import Turf
 
 // swiftlint:disable file_length type_body_length
 
-public enum PreferredFPS: RawRepresentable, Equatable {
-
-    /**
-     Create a `PreferredFPS` value from an `Int`.
-     - Parameter rawValue: The `Int` value to use as the preferred frames per second.
-     */
-    public init?(rawValue: Int) {
-        switch rawValue {
-        case Self.lowPower.rawValue:
-            self = .lowPower
-        case Self.normal.rawValue:
-            self = .normal
-        case Self.maximum.rawValue:
-            self = .maximum
-        default:
-            self = .custom(fps: rawValue)
-        }
-    }
-
-    public typealias RawValue = Int
-
-    /// The default frame rate. This can be either 30 FPS or 60 FPS, depending on
-    /// device capabilities.
-    case normal
-
-    /// A conservative frame rate; typically 30 FPS.
-    case lowPower
-
-    /// The maximum supported frame rate; typically 60 FPS.
-    case maximum
-
-    /// A custom frame rate. The default value is 30 FPS.
-    case custom(fps: Int)
-
-    /// The preferred frames per second as an `Int` value.
-    public var rawValue: Int {
-        switch self {
-        case .lowPower:
-            return 30
-        case .normal:
-            return -1
-        case .maximum:
-            return 0
-        case .custom(let fps):
-            // TODO: Check that value is a valid FPS value.
-            return fps
-        }
-    }
-
-}
-
-open class ObserverConcrete: Observer {
-
-    /// Map of event types to subscribed event handlers
-    internal var eventHandlers: [String: [(MapboxCoreMaps.Event) -> Void]] = [:]
-
-    /// Notify correct handler
-    public func notify(for event: MapboxCoreMaps.Event) {
-        let handlers = eventHandlers[event.type]
-        handlers?.forEach({ (handler) in
-            handler(event)
-        })
-    }
-}
-
 internal typealias PendingAnimationCompletion = (completion: AnimationCompletion, animatingPosition: UIViewAnimatingPosition)
 
-open class BaseMapView: UIView, MapClient, MBMMetalViewProvider, CameraViewDelegate {
+open class BaseMapView: UIView, CameraViewDelegate {
 
     /// The underlying renderer object responsible for rendering the map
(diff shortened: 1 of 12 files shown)


Sources/MapboxMaps/Foundation/BaseMapView.swift
```diff
@@ -3,100 +3,40 @@
...
     /// The underlying renderer object responsible for rendering the map
-    public var __map: Map!
+    public private(set) var __map: Map!
+
+    private let mapClient = DelegatingMapClient()
+    private let observer = DelegatingObserver()
 
     /// The underlying metal view that is used to render the map
-    internal var metalView: MTKView?
+    internal private(set) var metalView: MTKView?
```
neelmistry94, line 19: if this is internal, what value does a private(set) give us?

Sources/MapboxMaps/Foundation/BaseMapView.swift
```diff
@@ -3,100 +3,40 @@
...
 
     /// Resource options for this map view
-    internal var resourceOptions: ResourceOptions?
+    internal private(set) var resourceOptions: ResourceOptions?
```
neelmistry94, line 22: same as above

Sources/MapboxMaps/Foundation/BaseMapView.swift
```diff
@@ -3,100 +3,40 @@
...
+    private var displayCallback: (() -> Void)?
     @objc dynamic internal var displayLink: CADisplayLink?
 
-    @IBInspectable internal var styleURI__: String = ""
+    @IBInspectable private var styleURI__: String = ""
 
     /// Outlet that can be used when initializing a MapView with a Storyboard or
     /// a nib.
-    @IBOutlet internal weak var mapInitOptionsProvider: MapInitOptionsProvider?
+    @IBOutlet internal private(set) weak var mapInitOptionsProvider: MapInitOptionsProvider?
```
neelmistry94, line 39: same as above

Sources/MapboxMaps/Foundation/BaseMapView.swift
```diff
@@ -483,6 +389,48 @@ open class BaseMapView: UIView, MapClient, MBMMetalViewProvider, CameraViewDeleg
...
 
+extension BaseMapView: DelegatingObserverDelegate {
+    /// Notify correct handler
+    internal func notify(for event: MapboxCoreMaps.Event) {
+        let handlers = eventHandlers[event.type]
+        handlers?.forEach { (handler) in
+            handler(event)
+        }
+    }
+}
```
neelmistry94, line 400: Can we move this to a new file called BaseMapView+DelefatingObserverDelegate

Sources/MapboxMaps/Foundation/BaseMapView.swift
```diff
@@ -483,6 +389,48 @@ open class BaseMapView: UIView, MapClient, MBMMetalViewProvider, CameraViewDeleg
...
+        metalView.isPaused = true
+        metalView.enableSetNeedsDisplay = true
+        metalView.presentsWithTransaction = true
+
+        insertSubview(metalView, at: 0)
+        self.metalView = metalView
+
+        return metalView
+    }
+}
```
neelmistry94, line 432: similar to above, Can we move this to a new file called BaseMapView+DelefatingMapClientDelegate

Sources/MapboxMaps/Foundation/BaseMapView.swift
macdrevx: It's a way of saying that only the BaseMapView is allowed to modify its metalView. The same principles we use when designing public APIs can be valuable when designing the internal APIs as well.

Sources/MapboxMaps/Foundation/BaseMapView.swift
macdrevx: We could, but it seems unnecessary to me. What problem would we be solving by moving it?

neelmistry94: I don't think it solves a problem, but if these extensions have the potential to grow in the future, it would be nice to have a bloated 500 + line map view file.

macdrevx: If BaseMapView gets that big, the solution is probably to refactor it into multiple components rather than breaking it up into multiple files. I'd be in favor of leaving this as-is for now.

neelmistry94 (approved):

PR #267, opened 2021-04-14
Cleaning up StyleTransition to use swift values

neelmistry94:

## Pull request checklist:

### Summary of changes
Updated the StyleTransition so that swift code will be in seconds and conversions are handled in the SDK, since Core uses milliseconds.

### DIFF
--- a/Sources/MapboxMaps/Style/Style.swift
+++ b/Sources/MapboxMaps/Style/Style.swift
@@ -416,6 +416,11 @@ public class Style {
  */
 public struct StyleTransition: Codable {
 
+    public enum CodingKeys: String, CodingKey {
+        case duration
+        case delay
+    }
+
     /// Time allotted for transitions to complete in milliseconds.
     public var duration: TimeInterval = 0
 
@@ -427,8 +432,8 @@ public struct StyleTransition: Codable {
     ///   - duration: Time for transition in milliseconds.
     ///   - delay: Time before transition begins in milliseconds.
     public init(duration: TimeInterval, delay: TimeInterval) {
-        self.duration = duration * 1000
-        self.delay = delay * 1000
+        self.duration = duration
+        self.delay = delay
     }
 
     public init(from decoder: Decoder) throws {
@@ -436,5 +441,10 @@ public struct StyleTransition: Codable {
 
         duration = try container.decode(Double.self, forKey: .duration) * 1000
         delay = try container.decode(Double.self, forKey: .delay) * 1000
+    }
+
+    public func encode(to encoder: Encoder) throws {
+        var container = encoder.singleValueContainer()
+        try container.encode(self)
     }
 }
--- a/Sources/MapboxMaps/Style/Style.swift
+++ b/Sources/MapboxMaps/Style/Style.swift
@@ -416,7 +416,7 @@ public class Style {
  */
 public struct StyleTransition: Codable {
 
-    public enum CodingKeys: String, CodingKey {
+    internal enum CodingKeys: String, CodingKey {
         case duration
         case delay
     }
@@ -439,12 +439,14 @@ public struct StyleTransition: Codable {
     public init(from decoder: Decoder) throws {
         let container = try decoder.container(keyedBy: CodingKeys.self)
 
-        duration = try container.decode(Double.self, forKey: .duration) * 1000
-        delay = try container.decode(Double.self, forKey: .delay) * 1000
+        duration = try container.decode(Double.self, forKey: .duration) / 1000
+        delay = try container.decode(Double.self, forKey: .delay) / 1000
     }
 
     public func encode(to encoder: Encoder) throws {
-        var container = encoder.singleValueContainer()
-        try container.encode(self)
+        var container = encoder.container(keyedBy: CodingKeys.self)
+
+        duration = try container.encode(Double.self, forKey: .duration) * 1000
+        delay = try container.encode(Double.self, forKey: .delay) * 1000
     }
 }
--- a/Sources/MapboxMaps/Style/Style.swift
+++ b/Sources/MapboxMaps/Style/Style.swift
@@ -427,7 +427,14 @@ public struct StyleTransition: Codable {
     ///   - duration: Time for transition in milliseconds.
     ///   - delay: Time before transition begins in milliseconds.
     public init(duration: TimeInterval, delay: TimeInterval) {
-        self.duration = duration
-        self.delay = delay
+        self.duration = duration * 1000
+        self.delay = delay * 1000
+    }
+
+    public init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+
+        duration = try container.decode(Double.self, forKey: .duration) * 1000
+        delay = try container.decode(Double.self, forKey: .delay) * 1000
     }
 }
--- a/Sources/MapboxMaps/Style/Style.swift
+++ b/Sources/MapboxMaps/Style/Style.swift
@@ -421,16 +421,16 @@ public struct StyleTransition: Codable {
         case delay
     }
 
-    /// Time allotted for transitions to complete in milliseconds.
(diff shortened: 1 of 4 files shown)


Sources/MapboxMaps/Style/Style.swift
```diff
@@ -427,7 +427,14 @@ public struct StyleTransition: Codable {
     ///   - duration: Time for transition in milliseconds.
     ///   - delay: Time before transition begins in milliseconds.
     public init(duration: TimeInterval, delay: TimeInterval) {
-        self.duration = duration
-        self.delay = delay
+        self.duration = duration * 1000
+        self.delay = delay * 1000
```
macdrevx, line 431: `self.duration` and `self.delay` are also of type `TimeInterval`, so we shouldn't do the conversion here. If the goal is for the JSON serialized representation of a `StyleTransition` to represent these time intervals in ms, you can handle the conversions in custom `encode(to:)` and `init(from:)` implementations. Your goal is for the JSON to use ms, and this struct to use `TimeInterval`.

Sources/MapboxMaps/Style/Style.swift
```diff
@@ -416,6 +416,11 @@ public class Style {
  */
 public struct StyleTransition: Codable {
 
+    public enum CodingKeys: String, CodingKey {
```
macdrevx, line 419: Can this be internal?

Sources/MapboxMaps/Style/Style.swift
```diff
@@ -430,4 +435,16 @@ public struct StyleTransition: Codable {
...
+    public init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+
+        duration = try container.decode(Double.self, forKey: .duration) * 1000
+        delay = try container.decode(Double.self, forKey: .delay) * 1000
+    }
+
+    public func encode(to encoder: Encoder) throws {
+        var container = encoder.singleValueContainer()
+        try container.encode(self)
```
macdrevx, line 448: This looks like an infinite recursion

You need to encode each field * 1000

Sources/MapboxMaps/Style/Style.swift
```diff
@@ -430,4 +435,16 @@ public struct StyleTransition: Codable {
         self.duration = duration
         self.delay = delay
     }
+
+    public init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+
+        duration = try container.decode(Double.self, forKey: .duration) * 1000
```
macdrevx, line 442: These should divide to convert from ms to s

macdrevx (changes requested):

Sources/MapboxMaps/Style/Style.swift
nishant-karajgikar: This is probably not needed at all.. This struct should generate codingKeys automatically.

Sources/MapboxMaps/Style/Style.swift
```diff
@@ -430,4 +435,18 @@ public struct StyleTransition: Codable {
...
+
+        duration = try container.decode(Double.self, forKey: .duration) / 1000
+        delay = try container.decode(Double.self, forKey: .delay) / 1000
+    }
+
+    public func encode(to encoder: Encoder) throws {
+        var container = encoder.container(keyedBy: CodingKeys.self)
+
+        duration = try container.encode(Double.self, forKey: .duration) * 1000
+        delay = try container.encode(Double.self, forKey: .delay) * 1000
```
macdrevx, line 450:
```suggestion
        try container.encode(duration, forKey: .duration) * 1000
        try container.encode(delay, forKey: .delay) * 1000
```

neelmistry94: This change does not work, it brings a compiler error and we need to encode the key so I don't see the purpose of this change

Tests/MapboxMapsTests/Style/StyleTransitionTests.swift
```diff
@@ -0,0 +1,40 @@
+import XCTest
+
+#if canImport(MapboxMaps)
+@testable import MapboxMaps
+#else
+@testable import MapboxMapsStyle
+#endif
```
macdrevx, line 7:
```suggestion
@testable import MapboxMaps
```

This can be simplified since we don't maintain the module structure anymore.

Tests/MapboxMapsTests/Style/StyleTransitionTests.swift
```diff
@@ -0,0 +1,40 @@
...
+
+//swiftlint:disable explicit_top_level_acl explicit_acl
```
macdrevx, line 9: If these checks are still enabled for tests, let's disable them.

Tests/MapboxMapsTests/Style/StyleTransitionTests.swift
```diff
@@ -0,0 +1,40 @@
...
+
+    func testDecodeHasCorrectConversion() {
+        let jsonString = """
+        {
+            "duration": 1000,
+            "delay": 500
+        }
+        """
+
+        if let jsonData = jsonString.data(using: .utf8) {
```
macdrevx, line 20: I'd just force unwrap here since you control the input and you know that it doesn't fail.

Tests/MapboxMapsTests/Style/StyleTransitionTests.swift
```diff
@@ -0,0 +1,40 @@
...
+            let decodedTransition = try! JSONDecoder().decode(StyleTransition.self, from: jsonData)
+            let comparingTransition = StyleTransition(duration: 1.0, delay: 0.5)
+
+            XCTAssertEqual(comparingTransition.duration, decodedTransition.duration, "Duration should be equal to 1.0")
+            XCTAssertNotEqual(decodedTransition.duration, 1000, "Duration should be equal to 1.0")
+            XCTAssertEqual(comparingTransition.delay, decodedTransition.delay, "Delay should be equal to 0.5")
+            XCTAssertNotEqual(decodedTransition.delay, 500, "Duration should be equal to 1.0")
+        }
```
macdrevx, line 28: These assertions are more complicated than necessary. Just have 1 assertion for each property that compares to a constant.

Tests/MapboxMapsTests/Style/StyleTransitionTests.swift
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    func testEncodeHasCorrectConversion() {
```
macdrevx, line 31: I'd like to see this test isolate encoding instead of using both encode and decode.

macdrevx (changes requested):

Tests/MapboxMapsTests/Style/StyleTransitionTests.swift
```diff
@@ -0,0 +1,26 @@
...
+
+    func testDecodeHasCorrectConversion() {
+        let jsonData = jsonString.data(using: .utf8)!
+        let decodedTransition = try! JSONDecoder().decode(StyleTransition.self, from: jsonData)
+        let comparingTransition = StyleTransition(duration: 1.0, delay: 0.5)
+
+        XCTAssertEqual(comparingTransition.duration, 1.0, "Duration should be equal to 1.0")
+        XCTAssertEqual(decodedTransition.duration, 1.0, "Duration should be equal to 1.0")
+        XCTAssertEqual(comparingTransition.delay, 0.5, "Delay should be equal to 0.5")
+        XCTAssertEqual(decodedTransition.delay, 0.5, "Delay should be equal to 0.5")
```
macdrevx, line 16:
```suggestion
        XCTAssertEqual(decodedTransition.duration, 1.0, "Duration should be equal to 1.0")
        XCTAssertEqual(decodedTransition.delay, 0.5, "Delay should be equal to 0.5")
```

macdrevx (commented):
You don't need the comparing transition

macdrevx (approved):

Issue #252, opened 2021-04-07
StyleURI should be a struct


1ec5:
StyleURI appears to have the same semantics as what’s called an extensible enumeration in Objective-C. In Swift, it would be more elegant to represent it as a struct: the cases would become `static let`s, and the `custom(url:)` case would go away in favor of the existing `StyleURI(rawValue:)` initializer, which would be much simpler. This is the same way an Objective-C extensible enumeration gets bridged to Swift.

A struct would be more maintainable than the existing approach, especially if the set of available named styles changes in the future: there would be only one place to update instead of keeping four different parts of the file in sync. Anyone who needs to specify a custom URL can write `.init(rawValue: …)` instead of having to choose between `.init(rawValue: …)` and `.custom(url: …)`; the choice may not be obvious to a developer who doesn’t have the implementation code on hand.

https://github.com/mapbox/mapbox-maps-ios/blob/14fe6c2cbd25692a6946311b5df40f24eacb307b/Sources/MapboxMaps/Style/StyleURI.swift#L3-L6

For reference, the MapboxDirections library has a [DirectionsProfileIdentifier](https://github.com/mapbox/mapbox-directions-swift/blob/bbe54bc196e71f9603ecdde50fa125d27785bc5c/Sources/MapboxDirections/DirectionsProfileIdentifier.swift) struct with similar semantics.

/cc @mapbox/maps-ios

1ec5:
Moreover, if there are no plans to add methods specific to StyleURI or use StyleURIs in places that can’t accept arbitrary URLs, then it would be even simpler to move constants like `streets` to an extension on URL itself.

macdrevx:
I like this. cc: @julianrex

knov:
We should also consider applying this to:
- AnimationOwner
- PreferredFPS

macdrevx:
I decided to keep StyleURI instead of adding static methods to URL for 2 reasons:

1. StyleURI has extra requirements (non-nil scheme)
2. A custom type should reduce noise when using code completion

neelmistry94:
https://github.com/mapbox/mapbox-maps-ios/pull/342

PR #285, opened 2021-04-21
Convert StyleURI, PreferredFPS, and AnimationOwner to structs

macdrevx:
Fixes: #252 

## Pull request checklist:
 - [x] Update the changelog

### Summary of changes

- Converts StyleURI, PreferredFPS, and AnimationOwner to structs.

### DIFF
--- a/Sources/MapboxMaps/Style/StyleURI.swift
+++ b/Sources/MapboxMaps/Style/StyleURI.swift
@@ -2,91 +2,48 @@ import Foundation
 
 /// Enum representing the latest version of the Mapbox styles (as of publication). In addition,
 /// you can provide a custom URL or earlier version of a Mapbox style by using the `.custom` case.
-public enum StyleURI: Hashable, RawRepresentable {
-    public typealias RawValue = URL
+public struct StyleURI: Hashable, RawRepresentable {
+    public typealias RawValue = String
+
+    public let rawValue: String
+
+    /// Create a custom StyleURI from a String. The String may be a full HTTP or HTTPS URI, a Mapbox style URI
+    /// (mapbox://styles/{user}/{style}), or a path to a local file relative to the application’s
+    /// resource path.
+    /// Returns nil if the String is invalid.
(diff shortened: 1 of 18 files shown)


nishant-karajgikar (approved):

Sources/MapboxMaps/Style/StyleURI.swift
```diff
@@ -2,91 +2,39 @@ import Foundation
...
+    public typealias RawValue = String
+
+    public let rawValue: String
+
+    /// Create a custom StyleURI from a URL. The URL may be a full HTTP or HTTPS URL, a Mapbox style URL
+    /// (mapbox://styles/{user}/{style}), or a path to a local file relative to the application’s
+    /// resource path.
+    /// Returns nil if the URI is invalid.
+    /// - Parameter rawValue: URL for the style
+    public init?(rawValue: String) {
```
1ec5, line 15: Should the raw value be string or URL? Making it URL would guarantee well-formedness at the earliest possible opportunity, so that we wouldn’t need to manually validate the raw value before using it. Either way, consider adding a URL-typed convenience initializer to head off developer mistakes like using `path` instead of `absoluteString`.

macdrevx: > Should the raw value be string or URL?

I switched to String because that's what we ultimately pass to GL Native.

> Making it URL would guarantee well-formedness at the earliest possible opportunity, so that we wouldn’t need to manually validate the raw value before using it.

I believe the next line provides the safety you're looking for:

`guard let url = URL(string: rawValue), url.scheme != nil else`

Even when the raw value was a URL, we were still validating that its scheme was non-nil.

> Either way, consider adding a URL-typed convenience initializer to head off developer mistakes like using path instead of absoluteString.

I'm on board with this.

neelmistry94:
Closing in favor of new PR due to conflicts

neelmistry94:
closed in favor of https://github.com/mapbox/mapbox-maps-ios/pull/342

PR #307, opened 2021-04-28
[style][wip] Expose computed vars for operators + arguments of expressions

nishant-karajgikar:

## Pull request checklist:
 - [ ] Add an entry inside this element for inclusion in the `mapbox-maps-ios` changelog: `<changelog> Expose the operator and arguments of an expression </changelog>`.

### Summary of changes

This PR exposes computed vars to allow for introspection of the `operator` and `arguments` for an expression.


Sources/MapboxMaps/Style/Types/Expression.swift
```diff
@@ -7,17 +7,41 @@ public struct Expression: Codable, CustomStringConvertible, Equatable {
     /// The individual elements of the expression in an array
     internal var elements: [Element]
 
+    /// The operator of this expression
+    public var `operator`: Operator {
+        guard let first = elements.first, case Element.operator(let op) = first else {
+            fatalError("First element of the expression is not an operator.")
```
1ec5, line 13: If it’s possible to create a malformed expression like this, consider adding a test that this error is thrown, so it doesn’t accidentally fall out during a refactor down the road.

Sources/MapboxMaps/Style/Types/Expression.swift
```diff
@@ -7,17 +7,41 @@ public struct Expression: Codable, CustomStringConvertible, Equatable {
...
             elements = []
         }
 
         elements.insert(.operator(op), at: 0)
-        self.init(with: elements)
+        self.init(elements: elements)
+    }
+
+    /// Initialize an expression with an operator and arguments
+    public init(operator op: Operator, arguments: [Argument] = [.null]) {
```
1ec5, line 43: Does the default of `[.null]` cause the `arguments` getter to return a non-empty array?

Sources/MapboxMaps/Style/Types/Expression.swift
```diff
@@ -112,6 +136,7 @@ public struct Expression: Codable, CustomStringConvertible, Equatable {
         case string(String)
         case boolean(Bool)
         case numberArray([Double])
+        case stringArray([String])
```
1ec5, line 139: Does the style specification allow for other types to occur inside a literal array, such as an array of booleans or objects, or even heterogeneous arrays (which are allowed in JSON)? What if, instead of separate cases for each type that could occur inside an array, we have just a single `array` case that takes the evaluated type as an additional associated value?

Sources/MapboxMaps/Style/Types/Expression.swift
nishant-karajgikar: Ah it's not actually possible to do that.. I'm just fatally erroring to avoid making this return an optional.

Sources/MapboxMaps/Style/Types/Expression.swift
nishant-karajgikar: This was an oversight. It's been refactored out.

Sources/MapboxMaps/Style/Types/Expression.swift
nishant-karajgikar: Yeah, i've considered doing this... It's just that modeling those kind of heterogenous arrays is hard. Open to suggestions here.

Sources/MapboxMaps/Style/Types/ExpressionArgumentBuilder.swift
```diff
@@ -0,0 +1,77 @@
...
+    }
+}
+
+extension Bool: ExpressionArgumentConvertible {
+    public var expressionArguments: [Expression.Argument] {
+        return [.boolean(self)]
+    }
+}
+
+extension Array: ExpressionArgumentConvertible where Element == Double {
```
julianrex, line 53: Does string array need to be here too?

Tests/MapboxMapsTests/Style/ExpressionTests/ExpressionTests.swift
```diff
@@ -13,54 +13,17 @@ internal class ExpressionTests: XCTestCase {
...
-
-        if case let Exp.Element.argument(someArg) = sumExp.elements[2],
-           case let Exp.Argument.number(someDouble) = someArg,
-           someDouble == 12.0 { } else {
-            XCTFail("Third element is not an expression argument with the correct value (12.0)")
-        }
+        XCTAssertEqual(sumExp.operator, .sum)
+        XCTAssertEqual(sumExp.arguments.count, 2)
+        XCTAssertEqual(sumExp.arguments[0], .number(10))
+        XCTAssertEqual(sumExp.arguments[1], .number(12))
```
julianrex, line 19: Nice

julianrex (approved):
Just the one question, but refactoring looks good

Sources/MapboxMaps/Style/Types/ExpressionArgumentBuilder.swift
nishant-karajgikar: yep it does! Good catch.

PR #312, opened 2021-05-04
Change `contentMode` for `metalView` to prevent rendering issues on CarPlay.

MaximAlien:
Fixes: `MapView` rendering issues on CarPlay.

## Pull request checklist:

### Summary of changes

Fix resolves rendering `MapView` issues on CarPlay when device has notch (e.g. iPhone 12). Previously `MapView` was rendered like this:

Screen Shot 2021-05-04 at 10 30 39 AM

After fix it looks like this:

Screen Shot 2021-05-04 at 10 26 16 AM


nishant-karajgikar:
@MaximAlien , thanks for the PR!  My primary question here is whether this change could result in the map size maintained by the renderer being out of sync with the view size. What happens if the size of the containing view changes arbitrarily?

MaximAlien:
> @MaximAlien , thanks for the PR! My primary question here is whether this change could result in the map size maintained by the renderer being out of sync with the view size. What happens if the size of the containing view changes arbitrarily?

@nishant-karajgikar, looks like such change doesn't cause any issues with random changes in case if parent view contain `MapView`. For example, here's behavior when `metalView.contentMode = .center` is used:

https://user-images.githubusercontent.com/1496498/117827945-44573780-b226-11eb-8be2-30786f609e75.mov

Here's when `metalView.contentMode = .scaleToFill` is used:

https://user-images.githubusercontent.com/1496498/117828033-5638da80-b226-11eb-9e49-516be287b730.mov

Sources/MapboxMaps/Foundation/BaseMapView.swift
```diff
@@ -297,7 +297,7 @@ extension BaseMapView: DelegatingMapClientDelegate {
         metalView.autoresizingMask = [.flexibleHeight, .flexibleWidth]
         metalView.autoResizeDrawable = true
         metalView.contentScaleFactor = UIScreen.main.scale
-        metalView.contentMode = .center
+        metalView.contentMode = .scaleToFill
```
knov, line 300: @mapbox/maps-ios I've reviewed this in context of the updates to MapView/BaseMapView and am wondering if there is a reason that the contentMode is set to `.center` (default would be `.scaleToFill`). Thoughts? 

cc @alexshalamov

knov: Having reviewed this further, we do not want this view to scale as it would create a disconnect between the rendered content.

neelmistry94:
Closing in favor of #420

PR #320, opened 2021-05-05
Add source and source properties to StyleManagerProtocol

julianrex:
Continuation of #317 

## Pull request checklist:
 - [X] Add an entry inside this element for inclusion in the `mapbox-maps-ios` changelog: `<changelog>Refactored Style source and source property management.</changelog>`.

### Summary of changes

This is PR 2 of N (see #317):

- Adds source and source property functions that throw rather than returning a `Result` type.
- Removes `updateSourceProperty(...)`, replacing it with `setSourceProperty(...)`
- Updates tests and examples accordingly.

### DIFF
--- a/Sources/MapboxMaps/Style/Style.swift
+++ b/Sources/MapboxMaps/Style/Style.swift
@@ -130,6 +130,20 @@ public class Style {
         }
     }
 
+    // MARK: Layer properties
+
+    /// Gets the value of style layer property.
+    ///
+    /// - Parameters:
+    ///   - layerId: Style layer identifier.
+    ///   - property: Style layer property name.
+    ///
+    /// - Returns:
+    ///     The value of the property in the layer with layerId.
+    public func layerProperty(for layerId: String, property: String) -> Any {
+        return _layerProperty(for: layerId, property: property).value
+    }
+
     // MARK: Style images
 
     /**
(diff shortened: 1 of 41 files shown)


Sources/MapboxMaps/Style/Style.swift
```diff
@@ -433,7 +410,6 @@ extension Style: StyleManagerProtocol {
 
     // MARK: Layer Properties
 
-    /// :nodoc:
     public func _layerProperty(for layerId: String, property: String) -> StylePropertyValue {
```
neelmistry94, line 413: if we are removing nodoc, can we add docs to this since it is a public API

Sources/MapboxMaps/Style/Style.swift
```diff
@@ -445,7 +421,6 @@ extension Style: StyleManagerProtocol {
         }
     }
 
-    /// :nodoc:
```
neelmistry94, line 448: same as above

neelmistry94 (approved):
one nit about adding docs

Sources/MapboxMaps/Style/Style.swift
julianrex: The docs are inherited from the StyleManagerProtocol, which already had `:nodoc:` - this was about removing the redundancy.

PR #339, opened 2021-05-11
Move MapEvents `on` handling from BaseMapView to MapboxMap

julianrex:

## Pull request checklist:
 - [X] Write tests for all new functionality. If tests were not written, please explain why.
 - [X] ~Add example if relevant.~ Updated
 - [X] Add an entry inside this element for inclusion in the `mapbox-maps-ios` changelog: `<changelog>Moved `on()` from `BaseMapView` to `MapboxMap`.</changelog>`.

### Summary of changes

This PR changes the behavior of `on(_:handler:)` to handle map events and moves the function to `MapboxMap`.

`on(...)` has been replaced by 

- `onNext(...) -> Cancelable`
- `onEvery(...) -> Cancelable` 

Assuming there is no retain cycle that stops the `MapView` or `Snapshotter` from being deallocated, then unsubscribing will be handled on your behalf when the `MapView` is deallocated: The underlying listener will unsubscribe from the map, and the closure will be released.

If you do have a retain cycle or you need to stop listening, for example if you want to listen to 3 events then stop, then call `cancel` on the returned `Cancelable` object.

Here's an example of how the call-site changes:

```swift
mapView.on(.styleLoaded) { event in ... }
```

becomes

```swift
mapView.mapboxMap.onNext(.styleLoaded) { (event) in ... }
```

cc @mapbox/maps-android


Sources/MapboxMaps/Foundation/MapEventHandler.swift
```diff
@@ -0,0 +1,64 @@
...
+    /// - Parameters:
+    ///   - eventType: The event type to listen to.
+    ///   - handler: The block of code to execute when the event occurs. Return
+    ///     `true` to indicate that you have handled the event(s) and no longer
+    ///     wish to receive them.
+    ///
+    /// - Returns: A `Cancelable` object that you can use to stop listening for
+    ///     events, in the case your closure does not return `true`.
+    @discardableResult
+    func on(_ eventType: MapEvents.EventKind, handler: @escaping (Event) -> Bool) -> Cancelable
```
nishant-karajgikar, line 15: Is using a `bool` here as a return expressive enough? It feels a little obscure.. Maybe we should have two methods like so:
```swift
// This is a one-shot subscription
func onNext(_ eventType: MapEvents.EventKind, handler: @escaping (Event) -> Void)

// This is a many-shot subscription
func onEvery(_ eventType: MapEvents.EventKind, handler: @escaping (Event) -> Void) -> Cancelable
```

Sources/MapboxMaps/Foundation/WeakSet.swift
```diff
@@ -1,22 +1,22 @@
 import Foundation
 
 // swiftlint:disable force_cast
-internal class WeakCameraAnimatorSet {
+internal class WeakSet<T> {
```
nishant-karajgikar, line 4: Nice, I'd tried to do something like this earlier but was running into some weird issue with it.

neelmistry94 (approved):

Sources/MapboxMaps/Foundation/MapboxMap.swift
```diff
@@ -169,3 +177,47 @@ public final class MapboxMap {
...
+    }
+
+    @discardableResult
+    public func onEvery(_ eventType: MapEvents.EventKind, handler: @escaping (Event) -> Void) -> Cancelable {
+        let handler = MapEventHandler(for: [eventType.rawValue],
+                                      observable: self,
+                                      handler: { event in
+                                        handler(event)
+                                        return false
+                                      })
```
nishant-karajgikar, line 219: nit: could we do a trailing closure here?

Sources/MapboxMaps/Foundation/MapboxMap.swift
```diff
@@ -15,6 +15,14 @@ public final class MapboxMap {
...
     }
 
+    private var eventHandlers = WeakSet<MapEventHandler>()
+
+    deinit {
+        eventHandlers.allObjects.forEach {
+            $0.cancel()
+        }
+    }
+
```
nishant-karajgikar, line 25: Do we need a test for this?

Tests/MapboxMapsTests/Foundation/MapEventHandlerTests.swift
```diff
@@ -0,0 +1,161 @@
...
+
+    func testWeakContainer() {
+
+        class WeakHandlerContainer {
+            weak var handler: MapEventHandler?
+
+            deinit {
+                handler?.cancel()
+            }
+        }
```
julianrex, line 130: FYI @nishant-karajgikar

Sources/MapboxMaps/Foundation/MapboxMap.swift
julianrex: There is..sort of. See [below](https://github.com/mapbox/mapbox-maps-ios/pull/339/files#r631080035).

Sources/MapboxMaps/Foundation/MapboxMap.swift
julianrex: Done.

PR #349, opened 2021-05-13
Adding public access to the SourceInfo Id and SourceType properties

alexrrouse:
## Pull request checklist:

### Summary of changes
`Style` has a public computed variable that returns all the Source Identifiers currently added to the StyleManager, `allSourceIdentifiers` which is helpful to validate what has already been added, however the properties of the `SourceInfo` that it returns are not public so you can't validate it against local models.

This PR makes the `id` and `type` on `SourceInfo` public so consuming applications can use `allSourceIdentifiers`.  

Example usage would be:
```
var existingSourceIds = map.style.allSourceIdentifiers.map { $0.id }
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ Mapbox welcomes participation and contributions from everyone.
 
  - `StyleURI`, `PreferredFPS`, and `AnimationOwner` are now structs. ([#285](https://github.com/mapbox/mapbox-maps-ios/pull/285))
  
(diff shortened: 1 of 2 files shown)


julianrex:
@alexrrouse thanks for the PR - this was definitely an oversight. We have a pending change that does the same for `LayerInfo`, so we'll roll this change in that PR, and will update this one when it's live.

alexrrouse:
Appreciate the commit.  Will close this.

PR #390, opened 2021-05-25
Add internal map delegates, create platform Projection class

julianrex:

## Pull request checklist:

### Summary of changes

This PR:
- Introduces `MapTransformDelegate` and `MapProjectionDelegate` protocols, both currently `internal`
- Creates a platform specific `Projection` class that implements `MapProjectionDelegate`, shadowing the one from MapboxCoreMaps.
- Fixes a minor TileStore bug with `tileRegionGeometry`

The two protocols match Android, except for [`getMetersPerPixelAtLatitude(latitude: Double)`](https://github.com/mapbox/mapbox-maps-android/blob/52cb487484d3b0c5db48fcd5b7df9e0ebbcffcbc/sdk-base/src/main/java/com/mapbox/maps/plugin/delegates/MapProjectionDelegate.kt#L22-L30). Also on Android `MapboxMap` conforms to `MapProjectionDelegate`, for iOS it's a new platform `Projection`

### Developer impact

Developers will need to update any existing use of `Projection` in their code.

### DIFF
--- a/Sources/MapboxMaps/Foundation/MapProjectionDelegate.swift
+++ b/Sources/MapboxMaps/Foundation/MapProjectionDelegate.swift
@@ -51,7 +51,7 @@ internal protocol MapProjectionDelegate: AnyObject {
 }
 
 public class Projection: MapProjectionDelegate {
-    private init() {}
+    internal init() {}
 
     public static func metersPerPoint(for latitude: CLLocationDegrees, zoom: CGFloat) -> Double {
         return MapboxCoreMaps.Projection.getMetersPerPixelAtLatitude(forLatitude: latitude, zoom: Double(zoom))
--- a/Sources/MapboxMaps/Foundation/MapProjectionDelegate.swift
+++ b/Sources/MapboxMaps/Foundation/MapProjectionDelegate.swift
@@ -1,0 +1,73 @@
+import CoreLocation
+
+public protocol MapProjectionDelegate {
+    /// Calculate distance spanned by one pixel at the specified latitude and
+    /// zoom level.
+    ///
+    /// - Parameters:
+    ///   - latitude: The latitude for which to return the value
+    ///   - zoom: The zoom level
+    ///
+    /// - Returns: Meters
+    static func metersPerPoint(for latitude: CLLocationDegrees, zoom: CGFloat) -> Double
+
+    /// Calculate Spherical Mercator ProjectedMeters coordinates.
+    /// - Parameter coordinate: Coordinate at which to calculate the projected
+    ///     meters
+    ///
+    /// - Returns: Spherical Mercator ProjectedMeters coordinates
+    static func projectedMeters(for coordinate: CLLocationCoordinate2D) -> ProjectedMeters
+
+    /// Calculate a coordinate for a Spherical Mercator projected
+    /// meters.
+    ///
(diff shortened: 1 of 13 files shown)


neelmistry94 (approved):
Seems like a trivial change. Could we add some testing around these new protocols though? Seems like the projection delegate could be mocked for a series of unit tests

julianrex:
Not sure what testing the protocol will give us, but will add a couple of small tests to test some assumptions (e.g. conformance)

neelmistry94:
> Not sure what testing the protocol will give us, but will add a couple of small tests to test some assumptions (e.g. conformance)

They are simple pass through functions, but unit tests for those will help us Catch any breaking changes upstream from Core. It will be a nice catch all

Sources/MapboxMaps/Foundation/MapProjectionDelegate.swift
```diff
@@ -0,0 +1,75 @@
...
+    ///   - mercatorCoordinate: Point on the map in Mercator projection.
+    ///   - zoomScale: The current zoom factor applied on the map, is used to
+    ///         calculate the world size as tileSize * zoomScale (i.e.
+    ///         512 * 2 ^ Zoom level) where tileSize is the width of a tile in
+    ///         points.
+    /// - Returns: Unprojected coordinate
+    static func unproject(_ mercatorCoordinate: MercatorCoordinate, zoomScale: CGFloat) -> CLLocationCoordinate2D
+}
+
+public class Projection: MapProjectionDelegate {
```
pengdev, line 53: On Android the `MapboxMap : MapProjectionDelegate` to be more discoverable, where the functions inside MapProjectionDelegate is not static.

gl-native exposed generated static APIs can be directly used as `Projection.project(point, zoomScale)`.

Sources/MapboxMaps/Foundation/MapTransformDelegate.swift
```diff
@@ -0,0 +1,31 @@
+internal protocol MapTransformDelegate: AnyObject {
+    /// Gets the size of the map in points
+    var size: CGSize { get set }
```
pengdev, line 3: on Android, the size in MapTransformDelegate is read only, it is set through `MapboxRenderer`.

Sources/MapboxMaps/Foundation/MapboxMap.swift
```diff
@@ -174,6 +160,51 @@ public final class MapboxMap {
     }
 }
 
+extension MapboxMap: MapTransformDelegate {
```
pengdev, line 163: on Android these are not extension functions, but the functions directly inside MapboxMap, but I think this is just a implementation detail for iOS.

Sources/MapboxMaps/Foundation/MapTransformDelegate.swift
julianrex: We're ok here, as this property is currently internal.

Sources/MapboxMaps/Foundation/MapProjectionDelegate.swift
julianrex: Let's leave this as a point of difference? At the moment there's no need on iOS because of the lack of plugins. /cc @tobrun

neelmistry94 (approved):

PR #396, opened 2021-05-26
Introduce ResourceOptionsManager

julianrex:

## Pull request checklist:
 - [ ] ~Write tests for all new functionality. If tests were not written, please explain why.~ Updated
 - [ ] ~Add an entry inside this element for inclusion in the `mapbox-maps-ios` changelog: `<changelog></changelog>`.~ Manual edit.

### Summary of changes

This PR replaces `CredentialsManager` with `ResourceOptionsManager` creating a struct based `ResourceOptions` that shadows the type from MapboxCoreMaps. The bulk of the other changes are related to this except for the following:

- Update the cache path to live in `.mapbox`. This aligns with other frameworks.
- CI creates the `MapboxAccessToken` file - since it appears that the build pre-action script is not being run
- Updated min/max latitude projection test
- Re-enabled `testCacheManagerIsReleasedAndHandlerRetainingManagerIsCalled`
- Fix MapView retain issue with integration tests
- Fix `ResourceEvent` and associated test


Sources/MapboxMaps/Foundation/Extensions/Core/ResourceOptions.swift
```diff
@@ -3,7 +3,50 @@ import Foundation
 
 // MARK: - ResourceOptions
 
-extension ResourceOptions {
+/// Options to configure access to a resource
+public struct ResourceOptions {
+
+    /// The access token to access the resource. This should be a valid, non-empty,
+    /// Mapbox access token
+    public var accessToken: String
```
julianrex, line 11: I'd like us to consider making access token a custom type, for example an enum. Not unlike an optional, but instead of a `.none` case, we could have a `.search` or `.default` case that would look for the token in the Info.plist etc. This way it's more explicit and we could deprecate it later if we ever choose to.

Sources/MapboxMaps/Foundation/Extensions/Core/ResourceOptions.swift
```diff
@@ -41,90 +84,122 @@ extension ResourceOptions {
...
+        hasher.combine(assetPathURL)
         hasher.combine(cacheSize)
         hasher.combine(tileStore)
         hasher.combine(tileStoreUsageMode)
-        return hasher.finalize()
+    }
+}
+
+extension ResourceOptions: CustomStringConvertible, CustomDebugStringConvertible {
+    private func redactedAccessToken() -> String {
```
julianrex, line 155: If we do add a custom AccessToken type, then this description logic would move there too.

Sources/MapboxMaps/Foundation/ResourceOptionsManager.swift
```diff
@@ -0,0 +1,151 @@
...
+        }
+        return defaultInstance!
+    }
+
+    /// :nodoc:
+    /// Convenience function to remove the default instance. Calling `default`
+    /// again will re-create the default instance.
+    internal static func destroyDefault() {
+        defaultInstance = nil
+    }
```
julianrex, line 45: This is in preparation of being able to clean up TileStores completely - useful for tests.

Sources/MapboxMaps/Foundation/ResourceOptionsManager.swift
```diff
@@ -0,0 +1,151 @@
...
+    /// we will use appropriate defaults for the `ResourceOptions`, including
+    /// searching for an access token in the application's Info.plist.
+    ///
+    /// You can override the shared global access token, using
+    /// `ResourceOptionsManager.default`:
+    ///
+    ///     ```
+    ///     ResourceOptionsManager.default.update { resourceOptions in
+    ///         resourceOptions.accessToken = "overridden-access-token"
+    ///     }
```
julianrex, line 75: Old comments will be updated.

Sources/MapboxMaps/Foundation/ResourceOptionsManager.swift
```diff
@@ -0,0 +1,151 @@
...
+    ///     ```
+    ///
+    /// - Parameter accessToken: Valid access token or `nil`
+    public convenience init(accessToken: String? = nil) {
```
julianrex, line 79: Also added a convenience that just takes an access token.

Sources/MapboxMaps/Foundation/ResourceOptionsManager.swift
```diff
@@ -0,0 +1,151 @@
...
+        self.bundle = bundle
+        update(resourceOptions)
+    }
+
+    /// Reset the manager to a default `ResourceOptions` using the specified
+    /// access token.
+    ///
+    /// - Parameter accessToken: Valid access token or `nil`
+    internal func reset(resourceOptions: ResourceOptions? = nil) {
+        self.resourceOptions = resourceOptions ?? ResourceOptions(accessToken: defaultAccessToken())
```
julianrex, line 105: The default access token logic should probably move to where the core ResourceOptions object is initialized from the struct.

Sources/MapboxMaps/Foundation/ResourceOptionsManager.swift
```diff
@@ -0,0 +1,151 @@
...
+        let token = resourceOptions.accessToken.isEmpty ?
+            defaultAccessToken() : resourceOptions.accessToken
+
+        _resourceOptions = resourceOptions
+        _resourceOptions.accessToken = token
+
+        let resolvedTileStore = resourceOptions.tileStore ?? TileStore.default
+        if resourceOptions.tileStoreUsageMode != .disabled {
+            resolvedTileStore.setAccessToken(token)
+        }
```
julianrex, line 119: @pengdev this is what we were talking about earlier. ResourceOptions on its down no longer does this, only the `ResourceOptionsManager`.

Tests/MapboxMapsTests/Foundation/Projection/ProjectionTests.swift
```diff
@@ -87,5 +87,14 @@ class ProjectionTests: XCTestCase {
...
         XCTAssertEqual(southPole2.latitude, -85.051, accuracy: 0.001)
+
+        // Check clamping
+        let northMax = CLLocationCoordinate2D(latitude: Projection.latitudeMax, longitude: 0)
+        let northMaxMercator = projectorType.project(northMax, zoomScale: zoomScale)
+        XCTAssertEqual(northMercator.y, northMaxMercator.y, accuracy: 0.000001)
+
+        let southMin = CLLocationCoordinate2D(latitude: Projection.latitudeMin, longitude: 0)
+        let southMinMercator = projectorType.project(southMin, zoomScale: zoomScale)
+        XCTAssertEqual(southMercator.y, southMinMercator.y, accuracy: 0.000001)
```
julianrex, line 98: Unrelated test - will likely leave in this PR for simplicity

Tests/MapboxMapsTests/MapView/Integration Tests/Map/DidIdleFailureIntegrationTest.swift
```diff
@@ -33,7 +34,7 @@ internal struct ResourceEventResponse: Decodable {
     var noContent: Bool
     var notModified: Bool
     var mustRevalidate: Bool
-    var offlineData: Bool
+    var source: String
```
julianrex, line 37: Unrelated test fix.

Sources/MapboxMaps/Foundation/Extensions/Core/CacheManager.swift
```diff
@@ -1,6 +1,13 @@
 import Foundation
 
-extension CacheManager {
+extension MapboxCoreMaps.CacheManager {
```
neelmistry94, line 3: I know there was some convo around this, but does this need the full qualified name? Can we type alias it?

Sources/MapboxMaps/Foundation/Extensions/Core/ResourceOptions.swift
nishant-karajgikar: I like this idea.. may be we could even make it `expressibleByString` in some way?

nishant-karajgikar (approved):

Sources/MapboxMaps/Foundation/Extensions/Core/MapSnapshotOptions.swift
```diff
@@ -10,19 +10,21 @@ extension MapSnapshotOptions {
...
     ///   - resourceOptions: Resource fetching options to be used by the
     ///         snapshotter. Default uses the access token provided by
-    ///         `CredentialsManager.default`
+    ///         `ResourceOptionsManager.default`
     public convenience init(size: CGSize,
                             pixelRatio: CGFloat,
-                            glyphsRasterizationOptions: GlyphsRasterizationOptions? = GlyphsRasterizationOptions(fontFamilies: []),
-                            resourceOptions: ResourceOptions = ResourceOptions()) {
+                            glyphsRasterizationOptions: GlyphsRasterizationOptions? = GlyphsRasterizationOptions(),
+                            resourceOptions: ResourceOptions = ResourceOptionsManager.default.resourceOptions) {
```
pengdev, line 17: MapSnapshotOptions is a gl-native generated class in Android, and we couldn't override the default value here...

Sources/MapboxMaps/Foundation/Extensions/Core/ResourceOptions.swift
pengdev: cc @Chaoba 
on Android we can not overwrite the type in the ResourceOptions..

Sources/MapboxMaps/Foundation/Extensions/Core/ResourceOptions.swift
```diff
@@ -41,90 +84,122 @@ extension ResourceOptions {
...
+                                                                in: .userDomainMask,
+                                                                appropriateFor: nil,
+                                                                create: true)
 
-    private static func cacheURLIncludingSubdirectory() throws -> URL? {
-        var cacheDirectoryURL = try FileManager.default.url(for: .applicationSupportDirectory,
-                                                            in: .userDomainMask,
-                                                            appropriateFor: nil,
-                                                            create: true)
+            cacheDirectoryURL = cacheDirectoryURL.appendingPathComponent(".mapbox")
```
pengdev, line 110: so you make this change in the same PR? probably need to include this change to the change log.

Sources/MapboxMaps/Foundation/MapView.swift
```diff
@@ -102,7 +102,7 @@ open class MapView: UIView {
     /// - Parameters:
     ///   - frame: frame for the MapView.
     ///   - mapInitOptions: `MapInitOptions`; default uses
-    ///    `CredentialsManager.default` to retrieve a shared default access token.
+    ///    `ResourceOptionsManager.default` to retrieve a shared default access token.
```
pengdev, line 105:
```suggestion
    ///    `ResourceOptionsManager.default` to retrieve a shared default resource option, including the access token.
```

Sources/MapboxMaps/Foundation/ResourceOptionsManager.swift
```diff
@@ -0,0 +1,147 @@
...
+/// `application(_:didFinishLaunchingWithOptions:)` rather than relying on the
+/// value in your application's Info.plist:
+///
+///     ```
+///     func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?) -> Bool {
+///         // Override point for customization after application launch.
+///         ResourceOptionsManager.default.resourceOptions.accessToken = "overridden-access-token"
+///         return true
+///     }
+///     ```
```
pengdev, line 19: @Chaoba let's also align on the documentation

Sources/MapboxMaps/Foundation/ResourceOptionsManager.swift
```diff
@@ -0,0 +1,147 @@
...
+    /// This shared instance is used by the default initializers
+    /// for `ResourceOptions` and `MapInitOptions`.
+    ///
+    /// The application's Info.plist will be searched for a valid access token
+    /// under the key `MBXAccessToken`
+    ///
+    /// A valid access token must be provided or found.
+    public static var `default`: ResourceOptionsManager {
+        if defaultInstance == nil {
+            defaultInstance = ResourceOptionsManager(accessToken: nil)
```
pengdev, line 33: on iOS, do you throw exceptions when the accessToken is null?

Sources/MapboxMaps/Foundation/ResourceOptionsManager.swift
pengdev: we are considering to expose the same cc @Chaoba

Sources/MapboxMaps/Foundation/ResourceOptionsManager.swift
```diff
@@ -0,0 +1,147 @@
...
+        }
+        // Check for a bundled file
+        else if let url = bundle.url(forResource: "MapboxAccessToken", withExtension: nil),
+                let tokenFromFile = try? String(contentsOf: url) {
+            token = tokenFromFile
+        }
+
+        if token.isEmpty {
+            Log.warning(forMessage: "Empty access token.", category: "ResourceOptions")
+        }
```
pengdev, line 143: So you don't throw exceptions when the token is not found, and instead just show a warning message? What does it look like to users? will the MapView will show a black screen?

pengdev: fyi, if the access token is not found anywhere, we will throw a runtime exception that crashes the app..

cc @Chaoba

Sources/MapboxMaps/Foundation/Extensions/Core/ResourceOptions.swift
julianrex: Yep good call. Adding here for convenience :)

Sources/MapboxMaps/Foundation/ResourceOptionsManager.swift
julianrex: Not at the moment, does Android? Also, `nil` in this context means "search for the token".

Sources/MapboxMaps/Foundation/ResourceOptionsManager.swift
julianrex: We could do the same - though there's some trickiness around testing.

Sources/MapboxMaps/Foundation/ResourceOptionsManager.swift
julianrex: This has been removed in #429

PR #420, opened 2021-06-01
Fix Car Play Stretching

neelmistry94:

Fixes: MapView stretching on CarPlay displays

## Pull request checklist:

### Summary of changes

Removes the explicit setting of `contentScaleFactor` on metalView (MTKView) so that we use [nativeScale](https://developer.apple.com/library/archive/documentation/3DDrawing/Conceptual/MTLBestPracticesGuide/NativeScreenScale.html) by default. 

Verified by testing on 
- 2x handheld device (iPhone SE simulator) 
Simulator Screen Shot - iPhone SE (2nd generation) - 2021-05-25 at 12 19 39

- 3x handheld device (iPhone 11 Pro Max simulator) 
Simulator Screen Shot - iPhone 11 Pro Max - 2021-05-25 at 12 18 46

- iPad simulator (various)
Simulator Screen Shot - iPad Pro (9 7-inch) - 2021-05-25 at 12 12 15

- CarPlay simulator with 2x scale factor
CarPlay-2x

- CarPlay simulator with 3x scale factor
CarPlay-3x

Previously we implemented a default contentScaleFactor, however, Metal implements this by default.  (Reference: https://developer.apple.com/library/archive/documentation/3DDrawing/Conceptual/MTLBestPracticesGuide/NativeScreenScale.html)

This change means that when setting up a `MapView`, you should provide a pixelRatio for effective scaling. Consider using the active screen scaleFactor to find the most appropriate value. For example, if you are using `CarPlay`, you should initialize your `MapView` with `MapOptions` that contains a `pixelRatio: CPWindow.screen.scale`

### DIFF
--- a/Sources/MapboxMaps/Foundation/MapView.swift
+++ b/Sources/MapboxMaps/Foundation/MapView.swift
@@ -347,7 +347,7 @@ extension MapView: DelegatingMapClientDelegate {
         metalView.autoresizingMask = [.flexibleHeight, .flexibleWidth]
         metalView.autoResizeDrawable = true
         metalView.contentScaleFactor = pixelRatio
-        metalView.contentMode = .scaleAspectFill
+        metalView.contentMode = .scaleToFill
         metalView.isOpaque = isOpaque
         metalView.layer.isOpaque = isOpaque
         metalView.isPaused = true
--- a/Sources/MapboxMaps/Foundation/MapView.swift
+++ b/Sources/MapboxMaps/Foundation/MapView.swift
@@ -347,7 +347,7 @@ extension MapView: DelegatingMapClientDelegate {
         metalView.autoresizingMask = [.flexibleHeight, .flexibleWidth]
         metalView.autoResizeDrawable = true
         metalView.contentScaleFactor = pixelRatio
-        metalView.contentMode = .scaleToFill
+        metalView.contentMode = .scaleAspectFill
         metalView.isOpaque = isOpaque
         metalView.layer.isOpaque = isOpaque
         metalView.isPaused = true
--- a/Sources/MapboxMaps/Foundation/MapView.swift
+++ b/Sources/MapboxMaps/Foundation/MapView.swift
@@ -69,6 +69,8 @@ open class MapView: UIView {
     private var displayCallback: (() -> Void)?
     @objc dynamic internal var displayLink: CADisplayLink?
 
+    /// Holding onto this value that comes from `MapOptions` since there is a race condition between
+    /// getting a `MetalView`, and intializing a `MapView`
     private var pixelRatio: CGFloat = 0.0
 
     @IBInspectable private var styleURI__: String = ""
--- a/Sources/MapboxMaps/Foundation/MapView.swift
+++ b/Sources/MapboxMaps/Foundation/MapView.swift
@@ -347,7 +347,7 @@ extension MapView: DelegatingMapClientDelegate {
         metalView.autoresizingMask = [.flexibleHeight, .flexibleWidth]
         metalView.autoResizeDrawable = true
         metalView.contentScaleFactor = pixelRatio
-        metalView.contentMode = .scaleToFill
+        metalView.contentMode = .center
         metalView.isOpaque = isOpaque
         metalView.layer.isOpaque = isOpaque
         metalView.isPaused = true
--- a/Sources/MapboxMaps/Foundation/MapView.swift
+++ b/Sources/MapboxMaps/Foundation/MapView.swift
@@ -69,6 +69,8 @@ open class MapView: UIView {
     private var displayCallback: (() -> Void)?
     @objc dynamic internal var displayLink: CADisplayLink?
 
+    private var pixelRatio: CGFloat = 3.0 /// Same default value as found in `MapInitOptions`
+
     @IBInspectable private var styleURI__: String = ""
 
     /// Outlet that can be used when initializing a MapView with a Storyboard or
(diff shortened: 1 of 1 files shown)


Sources/MapboxMaps/Foundation/MapView.swift
```diff
@@ -341,8 +346,8 @@ extension MapView: DelegatingMapClientDelegate {
 
         metalView.autoresizingMask = [.flexibleHeight, .flexibleWidth]
         metalView.autoResizeDrawable = true
-        metalView.contentScaleFactor = UIScreen.main.scale
-        metalView.contentMode = .center
+        metalView.contentScaleFactor = pixelRatio
+        metalView.contentMode = .scaleToFill
```
alexshalamov, line 350: I think contentMode has to remain `.center`. What would be the case where content has to be scaled, especially non-proportionally?

Sources/MapboxMaps/Foundation/MapView.swift
```diff
@@ -69,6 +69,8 @@ open class MapView: UIView {
     private var displayCallback: (() -> Void)?
     @objc dynamic internal var displayLink: CADisplayLink?
 
+    private var pixelRatio: CGFloat = 3.0 /// Same default value as found in `MapInitOptions`
```
alexshalamov, line 72: Could it be `private var pixelRatio: CGFloat = UIScreen.main.scale` or simply `0.0`? The `pixelRatio` will be always overwritten by `self.pixelRatio = CGFloat(resolvedMapInitOptions.mapOptions.pixelRatio)`, right?

Sources/MapboxMaps/Foundation/MapView.swift
```diff
@@ -135,6 +137,9 @@ open class MapView: UIView {
         } else {
             resolvedMapInitOptions = mapInitOptions
         }
+
+        self.pixelRatio = CGFloat(resolvedMapInitOptions.mapOptions.pixelRatio)
```
alexshalamov, line 141: What would be the best place to document that for externally connected display, e.g., when [CarPlay window is connected](https://developer.apple.com/documentation/carplay/cptemplateapplicationscenedelegate/3340482-templateapplicationscene?language=objc), developer has to provide `MapOptions.pixelRatio` of a connected screen? For example `CPWindow.screen.scale`.

Sources/MapboxMaps/Foundation/MapView.swift
neelmistry94: Without changing the content mode, it actually does not consistently fix the issue. `.scaleToFile` guarantees the fix

Sources/MapboxMaps/Foundation/MapView.swift
neelmistry94: No, a user can initialize a MapView without the `MapOptions` so we need to provide a default

Sources/MapboxMaps/Foundation/MapView.swift
neelmistry94: I'd say the Migration Guide. @knov thoughts about this?

Sources/MapboxMaps/Foundation/MapView.swift
alexshalamov: > No, a user can initialize a MapView without the MapOptions

Could you please tell what is the codepath where `self.pixelRatio = CGFloat(resolvedMapInitOptions.mapOptions.pixelRatio)` will not be executed for constructed MapView and default will be used?

Sources/MapboxMaps/Foundation/MapView.swift
alexshalamov: Thanks, but that doesn't answer my question. What is the case where non-proportional scaling of rendered content is required?

neelmistry94: I see what you are saying, but I think when using `.scaleToFill` that accounts for that and properly scales the proportions. Especially when we are providing a scale factor

Sources/MapboxMaps/Foundation/MapView.swift
neelmistry94: You are correct, I was just checking it and it will always be initialized by the init. 

I thought we put it inside an if condition

Sources/MapboxMaps/Foundation/MapView.swift
alexshalamov: Why would we want to stretch view? Could you explain why `.scaleToFill` is used instead of, for example `.scaleAspectFit`or `.scaleAspectFill` that preserve aspect ratio?

neelmistry94: The issues with the other two is that `.scaleAspectFit` will not fill the screen and `.scaleAspectFill` will result in some clipping. Which in theory could be accounted for by zooming.

alexshalamov: How could map not fit the screen? We are not rendering static content, like an image. When view is laid-out, map could always render content preserving aspect ratio.

neelmistry94: I mean in my testing, they all produced similar results and you do bring up a good point. I'm good with moving it to scaleAspectFill

alexshalamov (approved):

Sources/MapboxMaps/Foundation/MapView.swift
neelmistry94: @alexshalamov and I discussed this and went over some examples. Keeping content mode to `.center` and specifically passing the right seale factor in fixes the issue

PR #428, opened 2021-06-02
support CLLocationDirection/course for bearing

neelmistry94:

Fixes: An issue where location puck was not tracking course

## Pull request checklist:

### Summary of changes
This change brings the support for `course` when updating the `locationPuck`. There is a new option that the developer can specify whether the bearing should be updated via `heading: CLHeading` or via `course: CLLocationDirection`

### DIFF
--- a/Sources/MapboxMaps/Location/LocationManager.swift
+++ b/Sources/MapboxMaps/Location/LocationManager.swift
@@ -56,7 +56,7 @@ public class LocationManager: NSObject {
             }
 
             if options.puckBearingSource != oldValue.puckBearingSource {
-                locationPuckManager?.changePuckBearingSource(to: options.puckBearingSource)
+                locationPuckManager?.puckBearingSource = options.puckBearingSource
             }
         }
     }
--- a/Sources/MapboxMaps/Location/LocationManager.swift
+++ b/Sources/MapboxMaps/Location/LocationManager.swift
@@ -54,6 +54,10 @@ public class LocationManager: NSObject {
             if let puckType = options.puckType, puckType != oldValue.puckType {
                 locationPuckManager?.changePuckType(to: puckType)
             }
+
+            if options.puckBearingSource != oldValue.puckBearingSource {
+                locationPuckManager?.changePuckBearingSource(to: options.puckBearingSource)
+            }
         }
     }
 
@@ -196,13 +200,13 @@ private extension LocationManager {
             locationProvider.startUpdatingHeading()
 
             if let locationPuckManager = locationPuckManager {
-                // This serves as a reset and handles the case if permissions were changed for accuracy
+                // This serves as a reset and handles the case if permissions were changed for accuracy 
                 locationPuckManager.changePuckStyle(to: currentPuckStyle)
             } else {
-                let locationPuckManager = LocationPuckManager(
-                    locationSupportableMapView: locationSupportableMapView,
-                    style: style,
-                    puckType: puckType)
+                let locationPuckManager = LocationPuckManager(locationSupportableMapView: locationSupportableMapView,
+                                                              style: style,
+                                                              puckType: puckType,
+                                                              puckBearingSource: options.puckBearingSource)
                 consumers.add(locationPuckManager)
                 self.locationPuckManager = locationPuckManager
             }
--- a/Sources/MapboxMaps/Location/LocationManager.swift
+++ b/Sources/MapboxMaps/Location/LocationManager.swift
@@ -1,10 +1,6 @@
 import CoreLocation
 import UIKit
 
-#if canImport(MapboxMapsFoundation)
-import MapboxMapsFoundation
-#endif
-
 /// An object responsible for notifying the map view about location-related events,
 /// such as a change in the device’s location.
 public class LocationManager: NSObject {
--- a/Sources/MapboxMaps/Location/LocationOptions.swift
+++ b/Sources/MapboxMaps/Location/LocationOptions.swift
@@ -24,6 +24,10 @@ public struct LocationOptions: Equatable {
     /// Sets the type of puck that should be used
     public var puckType: PuckType?
 
+    /// Specifies if a `Puck` should use `Heading` or `Course` for the bearing
+    /// This is an experimental option
+    public var puckBearingSource: PuckBearingSource = .heading
+
     public init() {}
 
 }
--- a/Sources/MapboxMaps/Location/LocationOptions.swift
+++ b/Sources/MapboxMaps/Location/LocationOptions.swift
@@ -1,7 +1,4 @@
 import Foundation
-#if canImport(MapboxMapsFoundation)
-import MapboxMapsFoundation
-#endif
 import CoreLocation
 
 /// A struct to configure a `LocationManager`
--- a/Sources/MapboxMaps/Location/Pucks/LocationPuckManager.swift
+++ b/Sources/MapboxMaps/Location/Pucks/LocationPuckManager.swift
@@ -2,6 +2,8 @@ import CoreLocation
 import MapboxCoreMaps
 import UIKit
 
+@_implementationOnly import MapboxCommon_Private
+
 #if canImport(MapboxMapsFoundation)
 import MapboxMapsFoundation
 #endif
@@ -27,6 +29,13 @@ public enum PuckType: Equatable {
     case puck3D(Puck3DConfiguration)
 }
 
+// MARK: PuckBearingSource
+/// This enum controls how the puck is oriented
+public enum PuckBearingSource: Equatable {
+    case heading
+    case course
+}
+
 // MARK: LocationPuckManager
 /// An object that is responsible for managing the location indicator which can be view based, or layer based
 internal class LocationPuckManager: LocationConsumer {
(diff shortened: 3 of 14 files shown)


julianrex (approved):

Sources/MapboxMaps/Location/Pucks/Puck3D.swift
```diff
@@ -95,7 +95,7 @@ internal class Puck3D: Puck {
     }
 
     // MARK: Protocol Implementation
-    internal func updateLocation(location: Location) {
+    internal func updateLocation(location: Location, bearingType: BearingType) {
```
nishant-karajgikar, line 98: shouldn't you be honoring bearing in this method?

Sources/MapboxMaps/Location/Pucks/LocationPuckManager.swift
```diff
@@ -49,19 +56,24 @@ internal class LocationPuckManager: LocationConsumer {
...
+        self.bearingType = bearingType
         self.locationSupportableMapView = locationSupportableMapView
         self.style = style
     }
 
     /// LocationConsumer protocol method that will handle location updates
     internal func locationUpdate(newLocation: Location) {
         if let puck = self.puck {
-            puck.updateLocation(location: newLocation)
+            puck.updateLocation(location: newLocation, bearingType: bearingType)
```
nishant-karajgikar, line 76: feels like overkill that you're passing in the `bearingType` on every invocation of `updateLocation`. Instead `bearingType` should be  a stored property in `Puck2D` / `Puck3D` . The thing to take care of in that situation is that updating the `bearingType` results in an update of the store property in the `Puck`.

Sources/MapboxMaps/Location/LocationOptions.swift
```diff
@@ -24,6 +24,10 @@ public struct LocationOptions: Equatable {
     /// Sets the type of puck that should be used
     public var puckType: PuckType?
 
+    /// Specifies if `Pucks` should use `Heading` or `Course` for the bearing
+    /// This is an experimental option
+    public var bearingType: BearingType = .heading
```
nishant-karajgikar, line 29: could this be renamed to something a bit different? `puckOrientationSource` ?

Sources/MapboxMaps/Location/Pucks/Puck3D.swift
neelmistry94: There was no existing use case of it which is why I decided not to make changes.

Sources/MapboxMaps/Location/LocationOptions.swift
neelmistry94: I feel like bearingType is more descriptive

Sources/MapboxMaps/Location/Pucks/LocationPuckManager.swift
neelmistry94: I see your point, but that feels like it might add overhead as opposed to just passing it through. At the end the day, the switch statement won't change

Sources/MapboxMaps/Location/Pucks/Puck3D.swift
nishant-karajgikar: It's used below as part of headingDirection

Sources/MapboxMaps/Location/Pucks/LocationPuckManager.swift
nishant-karajgikar: Hmm. It's not typical to pass this kind of configuration as part of doing some work.

Sources/MapboxMaps/Location/LocationOptions.swift
nishant-karajgikar: It feels quite vague to be honest.

neelmistry94: Are you okay with `puckBearingSource`

Sources/MapboxMaps/Location/Pucks/LocationPuckManager.swift
```diff
@@ -27,6 +27,13 @@ public enum PuckType: Equatable {
     case puck3D(Puck3DConfiguration)
 }
 
+// MARK: PuckBearingSource
+/// This enum represents the how the bearing should be tracked
```
nishant-karajgikar, line 31: Change to something like "Controls how the puck is oriented"

Tests/MapboxMapsTests/Location/LocationManagerTests.swift
```diff
@@ -86,4 +86,20 @@ internal class LocationManagerTests: XCTestCase {
         XCTAssertEqual(locationManager.options, locationOptions2)
         XCTAssertEqual(locationManager.locationPuckManager?.puckType, locationOptions2.puckType)
     }
+
+    func testUpdateLocationOptionsWithCoursePuckBearingSource() {
```
nishant-karajgikar, line 90: nice!

Sources/MapboxMaps/Location/Pucks/Puck.swift
```diff
@@ -8,6 +8,9 @@ internal protocol Puck {
     /// Property that stores the current `PuckStyle` of the puck
     var puckStyle: PuckStyle { get set }
 
+    /// Property that stores the current `PuckBearingSource` of the puck
```
nishant-karajgikar, line 11: We don't need both of these right? I think we should just use the `puckBearingSource` property.. We can get rid of the `updateBearingSource` method.

Sources/MapboxMaps/Location/LocationManager.swift
```diff
@@ -54,6 +54,10 @@ public class LocationManager: NSObject {
             if let puckType = options.puckType, puckType != oldValue.puckType {
                 locationPuckManager?.changePuckType(to: puckType)
             }
+
+            if options.puckBearingSource != oldValue.puckBearingSource {
+                locationPuckManager?.changePuckBearingSource(to: options.puckBearingSource)
```
nishant-karajgikar, line 59: this should be updated.

neelmistry94: To what? Unless you are saying to omit the `changePuckBearingSource` function on `LocationPuckManager` as well. if that is the case, I would disagree. I think one level of encapsulation here is nice to read

nishant-karajgikar: hmm let's address as part of the coming refactor.

Sources/MapboxMaps/Location/Pucks/LocationPuckManager.swift
```diff
@@ -114,4 +128,14 @@ internal class LocationPuckManager: LocationConsumer {
...
+
+    internal func changePuckBearingSource(to newPuckBearingSource: PuckBearingSource) {
+        puckBearingSource = newPuckBearingSource
+
+        if var puck = self.puck {
+            puck.puckBearingSource = newPuckBearingSource
+        } else {
+            createPuck()
+        }
+    }
```
nishant-karajgikar, line 140: Would this not result in a puck always showing?

neelmistry94: yes that would, I think since puckBearingSource is a nonOptional, we should just omit the else statement here

Sources/MapboxMaps/Location/Pucks/LocationPuckManager.swift
```diff
@@ -114,4 +130,14 @@ internal class LocationPuckManager: LocationConsumer {
             createPuck()
         }
     }
+
+    internal func changePuckBearingSource(to newPuckBearingSource: PuckBearingSource) {
+        guard var puck = self.puck else {
+            Log.warning(forMessage: "Puck must exist to change a PuckBearingSource", category: "Location")
+            return
+        }
```
nishant-karajgikar, line 138: why do you want to do this? if we're storing the `newPuckBearingSource` below, then we can just use that to configure the puck if/when the puck is created right?

nishant-karajgikar: Doing this will mean that the puckBearingSource must only be set after the style has finished loading, which is a bit of an arbitrary constraint.

Sources/MapboxMaps/Location/Pucks/LocationPuckManager.swift
```diff
@@ -27,6 +19,13 @@ public enum PuckType: Equatable {
     case puck3D(Puck3DConfiguration)
 }
 
+// MARK: PuckBearingSource
+/// This enum controls how the puck is oriented
+public enum PuckBearingSource: Equatable {
+    case heading
+    case course
```
nishant-karajgikar, line 26: Could you add some documentation here that goes into more detail about these individual cases?

Sources/MapboxMaps/Location/Pucks/LocationPuckManager.swift
```diff
@@ -49,11 +48,22 @@ internal class LocationPuckManager: LocationConsumer {
     /// The current puck type
     internal private(set) var puckType: PuckType
 
+    /// The type of value that should be passed for bearing
+    internal var puckBearingSource: PuckBearingSource = .heading {
+        didSet{
+            if puck != nil {
+                puck?.puckBearingSource = puckBearingSource
```
nishant-karajgikar, line 55: You don't need the if check here, would be cleaner to leave it out.

Sources/MapboxMaps/Location/Pucks/LocationPuckManager.swift
```diff
@@ -78,14 +88,17 @@ internal class LocationPuckManager: LocationConsumer {
 
         var puck: Puck
 
+
```
nishant-karajgikar, line 91: nit: spacing

nishant-karajgikar (approved):

Issue #357, opened 2021-05-15
fly(to: CameraOptions... does nothing from time to time


roboqo:
## Environment

- Xcode version: `12.5`
- iOS version: `14.5`
- Devices affected: `All including simulator`
- Maps SDK Version: `v10.0.0-beta.19.1`

## Observed behavior and steps to reproduce

MapView's method:
```swift
fly(to camera: CameraOptions, duration: TimeInterval? = nil, completion: AnimationCompletion? = nil) -> CameraAnimator?
```

sometimes does simply nothing. Camera does not fly to new location, just nothing. It depends on zoom level. When fly method stops working, after zooming in a bit more - it works again. Cannot find out what causes this behaviour. This only started to happen with `v10.0.0-beta.19.1` SDK

## Expected behavior

Camera should always fly to specified `center`, `zoom`, etc.

## Notes / preliminary analysis

As a workaround I tried to create custom camera animator, like:

```swift
flyAnimator?.stopAnimation()

let animator = mapView?.camera.makeAnimator(duration: 0.25, curve: .easeOut) { (transition) in
    transition.center.toValue = camera.center
    transition.padding.toValue = camera.padding
    transition.anchor.toValue = camera.anchor
    transition.zoom.toValue = camera.zoom
    transition.bearing.toValue = camera.bearing
    transition.pitch.toValue = camera.pitch
}

// Nil out the `internalAnimator` once the "ease to" finishes
animator?.addCompletion { [weak self] (position) in
    self?.flyAnimator = nil
}

animator?.startAnimation()
flyAnimator = animator
```

But this also does not work as expected with `v10.0.0-beta.19.1` SDK. Given the same center coordinates, same zoom etc, sometimes it animates correctly to desired location, but sometimes it stops in the middle or so. See GIF below:

![animator_bug](https://user-images.githubusercontent.com/66432/118363201-ec109480-b59b-11eb-95e8-aaff3cddfa89.gif)

PR #443, opened 2021-06-08
Ensure that animations / completion blocks are honored even when scheduleRepaint() isn't called

nishant-karajgikar:
…

Fixes: https://github.com/mapbox/mapbox-maps-ios/issues/357

## Pull request checklist:
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-ios` changelog: `<changelog>Fix issue where animations are not honored sometimes</changelog>`.

### Summary of changes

This PR fixes a subtle bug that happens with animations where they don't always get honored.

#### Root cause

The core issue is that animations are contrary to the regular rendering loop. Let's first consider the regular case of how a setCamera() call results in a render call.

If a developer calls `setCamera(..)` with some valid camera option, the renderer processes that as a request and eventually calls `scheduleRepaint()`. In [`scheduleRepaint()`, the platform SDK sets the `needsDisplayRefresh` flag.](https://github.com/mapbox/mapbox-maps-ios/blob/24e43136030cddda830817548007a5e3fbbb5633/Sources/MapboxMaps/Foundation/MapView.swift#L343-L345) This flag is [eventually referenced](https://github.com/mapbox/mapbox-maps-ios/blob/24e43136030cddda830817548007a5e3fbbb5633/Sources/MapboxMaps/Foundation/MapView.swift#L270) in `updateFromDisplayLink()`. Only if the flag is true (which it will be in the scenario where a `setCamera` had been previously called), does the `metalView` invoke `setNeedsDisplay` .

In the case of animations however, animators don't call `setCamera` ever(!).. Rather they return the `CameraOptions` that must be set for that particular time in the duration of the animation ( i.e. the interpolated camera options value).

So in the scenario, where the developer loads the mapView and immediately makes a call to something like `fly(to:)`, the `needsDisplayRefresh` flag is never set since the renderer will never call `scheduleRepaint()` (which in turn doesn't happen since setCamera is never called by an animator!).

#### Fix

This flow definitely needs some work so we can make this easier to work with in the future. I'll be opening up a ticket for it.

For now, however, this PR point fixes the issue by checking for the presence of a camera animator in the `cameraAnimatorsSet` in `updateFromDisplayLink()`.

### DIFF
--- a/Sources/MapboxMaps/Foundation/MapView.swift
+++ b/Sources/MapboxMaps/Foundation/MapView.swift
@@ -267,7 +267,9 @@ open class MapView: UIView {
             return
         }
 
-        if needsDisplayRefresh {
+        if needsDisplayRefresh
+            || cameraAnimatorsSet.allObjects.count > 0
+            || !pendingAnimatorCompletionBlocks.isEmpty {
             needsDisplayRefresh = false
 
             for animator in cameraAnimatorsSet.allObjects {


Sources/MapboxMaps/Foundation/MapView.swift
```diff
@@ -267,7 +267,9 @@ open class MapView: UIView {
             return
         }
 
-        if needsDisplayRefresh {
+        if needsDisplayRefresh
+            || cameraAnimatorsSet.allObjects.count > 0
+            || !pendingAnimatorCompletionBlocks.isEmpty {
```
julianrex, line 272: Good catch! We had a similar issue with pending completion blocks < 10.0.0, and required a similar fix IIRC.

julianrex (approved):

Sources/MapboxMaps/Foundation/MapView.swift
```diff
@@ -75,7 +75,7 @@ open class MapView: UIView {
         return cameraAnimatorsSet.allObjects
     }
 
-    private var needsDisplayRefresh: Bool = false
+    internal var needsDisplayRefresh: Bool = false
```
macdrevx, line 78: Can this remain private?

nishant-karajgikar: whoops, yes it can

neelmistry94 (approved):

PR #454, opened 2021-06-09
Added DownloadStatus convenience

julianrex:

## Pull request checklist:

### DIFF
--- a/Sources/MapboxMaps/Foundation/Extensions/Core/DownloadStatus.swift
+++ b/Sources/MapboxMaps/Foundation/Extensions/Core/DownloadStatus.swift
@@ -1,0 +1,122 @@
+@_exported import MapboxCommon
+@_implementationOnly import MapboxCommon_Private
+
+extension DownloadError: LocalizedError {
+    /// Standardized error message
+    public var errorDescription: String? {
+        return "\(code): \(message)"
+    }
+}
+
+extension DownloadStatus {
+
+    /// Initialize a `DownloadStatus`
+    ///
+    /// - Parameters:
+    ///   - downloadId: Download id which was created by download request.
+    ///   - state: State of download request.
+    ///   - error: Contains error information in case of failure when state is
+    ///         set to DownloadState::Failed.
+    ///   - totalBytes: Total amount of bytes to receive. In some cases this
+    ///         value is unknown until we get final part of the file.
+    ///   - receivedBytes: Amount of bytes already received and saved on the disk.
+    ///         Includes previous download attempts for a resumed download.
+    ///   - transferredBytes: Amount of bytes received during the current resume
+    ///         attempt. For downloads that weren't resumed, this value will be
+    ///         the same as receivedBytes.
+    ///   - downloadOptions: Download options used to send the download request.
+    ///   - httpResult: HTTP result. This field is only set for `DownloadState.failed`
+    ///         and `DownloadState.finished`.
+    ///         For `.failed` expect `HttpRequestError` to be provided for cases when
+    ///         DownloadErrorCode` is `NetworkError`. For `.finished` `HttpResponseData`
+    ///         is set, but with empty data field (since all the data was written
+    ///         to the disk).
+    public convenience init(downloadId: UInt64,
+                            state: DownloadState,
+                            error: DownloadError?,
+                            totalBytes: UInt64?,
+                            receivedBytes: UInt64,
+                            transferredBytes: UInt64,
+                            downloadOptions: DownloadOptions,
+                            httpResult: Result<HttpResponseData, HttpRequestError>) {
+
+        let expected: Expected<AnyObject, AnyObject>
+        switch httpResult {
+        case let .success(response):
+            expected = Expected(value: response)
+        case let .failure(error):
+            expected = Expected(error: error)
+        }
+
+        self.init(downloadId: downloadId,
+                  state: state,
+                  error: error,
+                  totalBytes: totalBytes?.NSNumber,
+                  receivedBytes: receivedBytes,
+                  transferredBytes: transferredBytes,
+                  downloadOptions: downloadOptions,
+                  httpResult: expected)
+    }
+
+    /// Convenice to initialize a `DownloadStatus` when the download state is `.pending`
+    ///
+    /// - Parameters:
+    ///   - error: Contains error information in case of failure when state is
+    ///         set to DownloadState::Failed.
+    ///   - totalBytes: Total amount of bytes to receive. In some cases this
+    ///         value is unknown until we get final part of the file.
+    ///   - downloadOptions: Download options used to send the download request.
+    ///   - httpResult: HTTP result. This field is only set for `DownloadState.failed`
+    ///         and `DownloadState.finished`.
+    ///         For `.failed` expect `HttpRequestError` to be provided for cases when
+    ///         DownloadErrorCode` is `NetworkError`. For `.finished` `HttpResponseData`
+    ///         is set, but with empty data field (since all the data was written
+    ///         to the disk).
+    public convenience init(error: DownloadError?,
+                            totalBytes: UInt64?,
+                            downloadOptions: DownloadOptions,
+                            httpResult: Result<HttpResponseData, HttpRequestError>) {
+
+        let expected: Expected<AnyObject, AnyObject>
+        switch httpResult {
+        case let .success(response):
+            expected = Expected(value: response)
+        case let .failure(error):
+            expected = Expected(error: error)
+        }
+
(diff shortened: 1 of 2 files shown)


Sources/MapboxMaps/Foundation/Extensions/Core/DownloadStatus.swift
```diff
@@ -0,0 +1,122 @@
...
+                  transferredBytes: transferredBytes,
+                  downloadOptions: downloadOptions,
+                  httpResult: expected)
+    }
+
+    /// Convenience to initialize a `DownloadStatus` when the download state is `.pending`
+    ///
+    /// - Parameters:
+    ///   - error: Contains error information in case of failure when state is
+    ///         set to `DownloadState.failed`.
```
macdrevx, line 65: This is confusing. If this initializer is for use with download state pending, why would you ever provide an error if that's only for the failed state?

Sources/MapboxMaps/Foundation/Extensions/Core/DownloadStatus.swift
```diff
@@ -0,0 +1,122 @@
...
+@_implementationOnly import MapboxCommon_Private
+
+extension DownloadError: LocalizedError {
+    /// Standardized error message
+    public var errorDescription: String? {
+        return "\(code): \(message)"
+    }
+}
+
+extension DownloadStatus {
```
macdrevx, line 11: Could this concept be modeled more cleanly using an enum with associated values? I'd love to avoid needing to lean so heavily on documentation to say in which states things will be nil or not.

Sources/MapboxMaps/Foundation/Extensions/Core/DownloadStatus.swift
```diff
@@ -0,0 +1,122 @@
...
+    /// `DownloadErrorCode` is `NetworkError`.
+    ///  And for `.finished` `HttpResponseData` is set, but with empty data field
+    ///  (since all the data was written to the disk).
+    public var httpResult: Result<HttpResponseData, HttpRequestError>? {
+        guard let httpExpected = __httpResult else {
+            return nil
+        }
+
+        guard let expected = httpExpected as? Expected<HttpResponseData, HttpRequestError>  else {
+            fatalError("Invalid Expected types or none.")
```
macdrevx, line 107: I think this would only happen due to `Invalid Expected types`. The `none` case would be caught be the previous guard statement, right?

Tests/MapboxMapsTests/Foundation/Extensions/Core/DownloadStatusTests.swift
```diff
@@ -0,0 +1,89 @@
+import XCTest
+@testable import MapboxMaps
+
+final class DownloadStatusTests: XCTestCase {
+
+    func testSuccessRoundtrip1() throws {
```
macdrevx, line 6: Is this meant to test the httpResult handling specifically? I'm thinking about whether we could come up with a more descriptive name because it's not clear to me what the focus of this test is.

Tests/MapboxMapsTests/Foundation/Extensions/Core/DownloadStatusTests.swift
```diff
@@ -0,0 +1,89 @@
+import XCTest
+@testable import MapboxMaps
+
+final class DownloadStatusTests: XCTestCase {
```
macdrevx, line 4: I'm thinking there are maybe 9 tests to do here…

1. Test the first convenience initializer where httpResult is success. Use `__httpResult`, not `httpExpected` to verify the behavior so that we can isolate the initializer
2. Like 1, but where httpResult is failure
3. Like 1, but for the second convenience initializer
4. Like 2, but for the second convenience initializer
5. Test httpResult in the case where `__httpResult` is `nil`. Don't use the convenience constructors in this test so that we can better isolate httpResult's implementation
6. Test httpResult for the case where __httpResult is a value
7. Test httpResult for the case where __httpResult is an error
8. Test totalBytes when it's nil
9. Test totalBytes when it's non-nil

macdrevx (changes requested):
Pausing my review. I think the test improvements are the most important. Remodeling DownloadStatus as an enum can be deferred, since that seems outside of the scope here (refining an existing construct vs wrapping it in a new one).

Tests/MapboxMapsTests/Foundation/Extensions/Core/DownloadStatusTests.swift
```diff
@@ -0,0 +1,206 @@
...
+            XCTFail("The result was not a failure.")
+        }
+    }
+
+    func testConvenienceInitializerSuccess() throws {
+        let status = DownloadStatus(downloadId: 1,
+                                            state: .finished, error: .none, totalBytes: 1234,
+                                            receivedBytes: 1234, transferredBytes: 1234,
+                                            downloadOptions: downloadOptions,
+                                            httpResult: .success(httpResponseData))
```
macdrevx, line 125: indentation

macdrevx: also, can we put each param on its own line?

Tests/MapboxMapsTests/Foundation/Extensions/Core/DownloadStatusTests.swift
```diff
@@ -0,0 +1,206 @@
...
+            XCTFail("httpResult should be a value.")
+        }
+    }
+
+    func testConvenienceInitFailure() throws {
+        let status = DownloadStatus(downloadId: 1,
+                                            state: .finished, error: downloadError, totalBytes: 1234,
+                                            receivedBytes: 1234, transferredBytes: 1234,
+                                            downloadOptions: downloadOptions,
+                                            httpResult: .failure(httpRequestError))
```
macdrevx, line 141: same here

Tests/MapboxMapsTests/Foundation/Extensions/Core/DownloadStatusTests.swift
```diff
@@ -0,0 +1,206 @@
...
+        XCTAssertNil(httpResult.value, "The value should be nil.")
+
+        let httpErrorWrapped: Result<HttpResponseData, HttpRequestError> = .failure(httpRequestError)
+        guard case .failure(let error) = httpErrorWrapped else {
+            XCTFail("Not a failure")
+            return
+        }
+
+        XCTAssertEqual(error, httpRequestError, "The two errors should be equal.")
+    }
```
macdrevx, line 186: These tests should be covering the behavior of the DownloadStatus.httpResult property. Sorry that wasn't clear in. my earlier comment.

Tests/MapboxMapsTests/Foundation/Extensions/Core/DownloadStatusTests.swift
```diff
@@ -0,0 +1,184 @@
...
+        if let result = status.__httpResult, result.isError(), let error = result.error as? HttpRequestError {
+            XCTAssertEqual(error, httpRequestError, "__httpResult should be \(httpRequestError.debugDescription), got \(error)")
+        } else {
+            XCTFail("__httpResult should be an error, got \(String(describing: status.__httpResult))")
+        }
+    }
+
+    func testNilShortInitializer() {
+        let status = DownloadStatus(error: nil,
+                                    totalBytes: UInt64(1234),
```
macdrevx, line 122:
```suggestion
                                    totalBytes: NSNumber(value: 1234), // forces using the unrefined initializer
```

Tests/MapboxMapsTests/Foundation/Extensions/Core/DownloadStatusTests.swift
```diff
@@ -0,0 +1,184 @@
...
+        XCTAssertEqual(status.__totalBytes, 1234, "The value for __totalBytes should be 1234, got \(status.__totalBytes.debugDescription)")
+        XCTAssertEqual(status.downloadOptions, downloadOptions, "The value for downloadOptions should be \(downloadOptions.debugDescription), got \(status.downloadOptions.debugDescription)")
+        XCTAssertNil(status.__httpResult, "_httpResult should be nil, got \(status.__httpResult.debugDescription)")
+    }
+
+    func testNilHttpResult() {
+        let status = DownloadStatus(downloadId: 1,
+                                    state: .finished,
+                                    error: .none,
+                                    totalBytes: UInt64(2),
```
macdrevx, line 136:
```suggestion
                                    totalBytes: NSNumber(value: 2), // forces using the unrefined initializer
```

macdrevx (approved):

PR #467, opened 2021-06-11
Fix SwiftUI Example

### LINKED ISSUE #425: SwiftUIExample.swift example does not behave as stated in comment
## Environment

- Xcode version: 12.5
- iOS version: 14.5
- Devices affected: all
- Maps SDK Version: 10.0.0-beta.20

In `SwiftUIExample.swift`, the binding between the custom `Camera` struct and `mapView.cameraState` is only used once at map load, contrary to the comments around the call.

## Expected behavior

According to the comment here

    ...
    /// The coordinator observes the `.cameraChanged` event, and
    /// whenever the camera changes, it updates the camera binding
    mapView?.mapboxMap.onNext(.cameraChanged, handler: notify(for:))
    ...

The Slider view position and mapView zoom value should be bound - the handler should be called many times, whenever the camera changes.

It looks like this is caused by using `.onNext` to detect the `.cameraChanged` event, instead of `.onEvery`

macdrevx:
- Changes onNext to onEvery (Fixes #425)
- Updates to avoid modifying state during view update. (This bug was a
  regression introduced by a recent change to make it so that the camera
  changed event is delivered synchronously instead of asynchronously.)

## Pull request checklist:

### DIFF
--- a/Apps/Examples/Examples/All Examples/SwiftUIExample.swift
+++ b/Apps/Examples/Examples/All Examples/SwiftUIExample.swift
@@ -89,14 +89,19 @@ internal struct SwiftUIMapView: UIViewRepresentable {
     /// If your `SwiftUIMapView` is reconfigured externally, SwiftUI will invoke `updateUIView(_:context:)`
     /// to give you an opportunity to re-sync the state of the underlying map view.
     func updateUIView(_ mapView: MapView, context: Context) {
-        mapView.mapboxMap.setCamera(to: CameraOptions(center: camera.center, zoom: camera.zoom))
+        /// When setting the camera, we need to temporarily disable the coordinator's observers.
+        /// If we didn't do this, the SwiftUI state would be modified during view update, which
+        /// causes undefined behavior.
+        context.coordinator.performWithoutObservation {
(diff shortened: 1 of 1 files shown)


Apps/Examples/Examples/All Examples/SwiftUIExample.swift
```diff
@@ -89,14 +89,19 @@ internal struct SwiftUIMapView: UIViewRepresentable {
     /// If your `SwiftUIMapView` is reconfigured externally, SwiftUI will invoke `updateUIView(_:context:)`
     /// to give you an opportunity to re-sync the state of the underlying map view.
     func updateUIView(_ mapView: MapView, context: Context) {
-        mapView.mapboxMap.setCamera(to: CameraOptions(center: camera.center, zoom: camera.zoom))
+        /// When setting the camera, we need to temporarily disable the coordinator's observers.
+        /// If we didn't do this, the SwiftUI state would be modified during view update, which
+        /// causes undefined behavior.
+        context.coordinator.performWithoutObservation {
```
julianrex, line 95: FYI @jmkiley

julianrex (approved):

PR #475, opened 2021-06-15
[annotations] Update generated code for layer persistence

jmkiley:

## Pull request checklist:

### Summary of changes

This PR updates annotations to add support for layer persistence. By default, annotation layers will exist after the map's style changes as part of the new style. This can be disabled by passing `false` into the `shouldPersist` argument when creating an annotation manager. The annotation manager still needs to exist as long as the annotations should be visible.

This is the iOS equivalent to the annotations work on https://github.com/mapbox/mapbox-maps-android/pull/368


CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ Mapbox welcomes participation and contributions from everyone.
 
 ## main
 
+### Breaking changes ⚠️
+
+- Annotations now will persist across style changes. ([#475](https://github.com/mapbox/mapbox-maps-ios/pull/475))
+
```
nishant-karajgikar, line 10: Isn't this a feature?

Sources/MapboxMaps/Annotations/Generated/CircleAnnotationManager.swift
```diff
@@ -41,13 +41,17 @@ public class CircleAnnotationManager: AnnotationManager {
     /// Dependency required to add gesture recognizer to the MapView
     private weak var view: UIView?
 
-    internal init(id: String, style: Style, view: UIView, mapFeatureQueryable: MapFeatureQueryable, layerPosition: LayerPosition?) {
+    /// Indicates whether the style layer exists after style changes. Default value is `true`.
+    public private(set) var isPersistent: Bool
```
macdrevx, line 45: Are we sure we want this to be optional behavior? In what situation would you want this to be false?

Sources/MapboxMaps/Annotations/AnnotationOrchestrator.swift
```diff
@@ -77,6 +79,7 @@ public class AnnotationOrchestrator {
     ///   - layerPosition: Optionally set the `LayerPosition` of the layer managed.
     /// - Returns: An instance of `PolygonAnnotationManager`
     public func makePolygonAnnotationManager(id: String = String(UUID().uuidString.prefix(5)),
+                                             isPersistent: Bool = true,
```
nishant-karajgikar, line 82: should this be `shouldPersist` instead of `isPersistent`. Wonder if it reads better..

Tests/MapboxMapsTests/Annotations/Generated/PointAnnotationIntegrationTests.swift
```diff
@@ -139,7 +139,33 @@ class PointAnnotationIntegrationTests: MapViewIntegrationTestCase {
         XCTAssertEqual(featureProperties["text-halo-width"] as? Double, annotation.textHaloWidth)
         XCTAssertEqual(featureProperties["text-opacity"] as? Double, annotation.textOpacity)
     }
+
+  func testAnnotationPersistence() throws {
```
nishant-karajgikar, line 143: nice!

Sources/MapboxMaps/Annotations/AnnotationOrchestrator.swift
jmkiley: Fixed in code gen so it's consistent.

jmkiley:
Updated generated code based on reviews.

Sources/MapboxMaps/Annotations/AnnotationOrchestrator.swift
```diff
@@ -106,15 +114,18 @@ public class AnnotationOrchestrator {
...
     }
 
     /// Creates a `CircleAnnotationManager` which is used to manage a collection of `CircleAnnotation`s.
     /// - Parameters:
     ///   - id: Optional string identifier for this manager.
+    ///   - shouldPersist: Indicates whether the annotation layer should exist in after changing the map's style. Default value is `true`.
     ///   - layerPosition: Optionally set the `LayerPosition` of the layer managed.
     /// - Returns: An instance of `CircleAnnotationManager`
     public func makeCircleAnnotationManager(id: String = String(UUID().uuidString.prefix(5)),
+                                            shouldPersist: Bool = true,
```
pengdev, line 128: is this a public API? In Android we always use persistent layer, user can still remove the annotation even using the persistent layer.

pengdev: We didn't introduce any API breaking changes in Android, if we decide to introduce this flag, we need to align and add on both platforms.

Sources/MapboxMaps/Annotations/Generated/CircleAnnotationManager.swift
pengdev: fyi, we didn't expose this option to users on Android, we will always the persistent layer for annotations.

Sources/MapboxMaps/Annotations/AnnotationOrchestrator.swift
nishant-karajgikar: So this doesn't qualify as a breaking change on iOS since `shouldPersist` is defaulted to true.. Having said that, I'm having second thoughts on whether we should allow users the option of setting this to `false`.

@jmkiley : how do you feel about removing it from the public API (i.e. the `make*AnnotationManager` factory methods in `AnnotationOrchestrator`) ? We can leave them in the internal constructors of each of the manaagers..

jmkiley: Because we provide a default value, I'm not sure that this would be a breaking change. Developers can still call `mapView.annotations.makePointAnnotationManager()` without passing in any values, for example, there is just an option to pass a value in for `shouldPersist` now.

jmkiley: That said, we can remove this option to align with Android, but it seems like developers should have the option to disable this behavior for annotations.

pengdev: > So this doesn't qualify as a breaking change on iOS since shouldPersist is defaulted to true.. Having said that, I'm having second thoughts on whether we should allow users the option of setting this to false.

Sorry, yes, it is not a breaking change, but it's a new API that user could use to configure the annotations, and we don't expose such option on Android.

I'm not sure if we have a use case to set it to false, and considering API consistency, we don't have such options for LocationComponentPlugin as well, the default behaviour is that the location puck will persist. I think we should do the same for the Annotations.

pengdev: We can still add it to both platforms when customers request it, let's iterate on them 👍

Sources/MapboxMaps/Annotations/Generated/CircleAnnotationManager.swift
```diff
@@ -41,13 +41,17 @@ public class CircleAnnotationManager: AnnotationManager {
     /// Dependency required to add gesture recognizer to the MapView
     private weak var view: UIView?
 
-    internal init(id: String, style: Style, view: UIView, mapFeatureQueryable: MapFeatureQueryable, layerPosition: LayerPosition?) {
+    /// Indicates whether the style layer exists after style changes. Default value is `true`.
+    public private(set) var shouldPersist: Bool
```
nishant-karajgikar, line 45: I thought you'd changed this to be a `let` and `internal`?

jmkiley: Oh, thanks! Switched it back.

nishant-karajgikar (approved):
LGTM!

PR #480, opened 2021-06-17
Feature/localization

neelmistry94:

Fixes: Localization Feature

## Pull request checklist:
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-ios` changelog: `<changelog>Adds localization support for v10 Maps SDK. This can be used by setting the `mapView.locale`. There is a `SupportedLanguages` enum that will allow you to create a supported `Locale`</changelog>`.

### Summary of changes
This is the introduction for Localization Support in Maps SDK v10. We will take the existing expression found on a `SymbolLayer.textField` and convert it to a JSON String. Replace all instances of the expression `["get","name_xx"]` with the new locale that can be set on the `mapView`. It will then be converted back to an expression type and the layer will be updated.

### DIFF
--- a/Sources/MapboxMaps/Foundation/MapView.swift
+++ b/Sources/MapboxMaps/Foundation/MapView.swift
@@ -35,9 +35,10 @@ open class MapView: UIView {
     /// Controls the addition/removal of annotations to the map.
     public internal(set) var annotations: AnnotationOrchestrator!
 
+    /// Property that describes the current language for `SymbolLayer.textField`
     public var locale: Locale = Locale.current {
         didSet {
-            mapboxMap.localizeLabels(into: locale)
+            mapboxMap.style.localizeLabels(into: locale)
         }
     }
 
--- a/Sources/MapboxMaps/Foundation/MapView.swift
+++ b/Sources/MapboxMaps/Foundation/MapView.swift
@@ -37,7 +37,11 @@ open class MapView: UIView {
     /// Property that describes the current language for `SymbolLayer.textField`
     public var locale: Locale = Locale.current {
         didSet {
-            mapboxMap.style.localizeLabels(into: locale)
+            do {
+                try mapboxMap.style.localizeLabels(into: locale)
+            } catch {
+                Log.warning(forMessage: "Error when localizing labels", category: "Style")
+            }
         }
     }
 
--- a/Sources/MapboxMaps/Foundation/MapView.swift
+++ b/Sources/MapboxMaps/Foundation/MapView.swift
@@ -1,3 +1,4 @@
+// swiftlint:disable file_length
 @_exported import MapboxCoreMaps
 @_exported import MapboxCommon
 @_implementationOnly import MapboxCoreMaps_Private
--- a/Sources/MapboxMaps/Foundation/MapView.swift
+++ b/Sources/MapboxMaps/Foundation/MapView.swift
@@ -2,6 +2,7 @@
 @_exported import MapboxCommon
 @_implementationOnly import MapboxCoreMaps_Private
 @_implementationOnly import MapboxCommon_Private
+import Foundation
 import UIKit
 import Turf
 
@@ -33,6 +34,12 @@ open class MapView: UIView {
 
     /// Controls the addition/removal of annotations to the map.
     public internal(set) var annotations: AnnotationOrchestrator!
+
+    public var locale: Locale = Locale.current {
+        didSet {
+            mapboxMap.localizeLabels(into: locale)
+        }
+    }
 
     /// A reference to the `EventsManager` used for dispatching telemetry.
     internal var eventsListener: EventsListener!
--- a/Sources/MapboxMaps/Foundation/MapView.swift
+++ b/Sources/MapboxMaps/Foundation/MapView.swift
@@ -2,7 +2,6 @@
 @_exported import MapboxCommon
 @_implementationOnly import MapboxCoreMaps_Private
 @_implementationOnly import MapboxCommon_Private
-import Foundation
 import UIKit
 import Turf
 
--- a/Sources/MapboxMaps/Style/Style+Localization.swift
+++ b/Sources/MapboxMaps/Style/Style+Localization.swift
@@ -1,0 +1,40 @@
+@_implementationOnly import MapboxCommon_Private
+
+/// Support for localization of labels
+extension Style {
+
+    /// This function creates an expression that will localize the `textField` property of a `SymbolLayer`
+    /// - Parameter locale: A `SupportedLanguage` based `Locale`
+    internal func localizeLabels(into locale: Locale) {
+
+        /// Get all symbol layers that are currently on the map
+        let symbolLayers = allLayerIdentifiers.filter { layer in
+            return layer.type == .symbol
+        }
+
+        /// Expression to be applied to the `SymbolLayer.textField`to localize the language
+        /// Sample Expression JSON: `["format",["coalesce",["get","name_en"],["get","name"]],{}]`
+        let expression = Exp(.format) {
+            Exp(.coalesce) {
+                Exp(.get) {
+                    "name_\(locale.identifier)"
+                }
+                Exp(.get) {
+                    "name"
+                }
+            }
+        }
+
+        for layerInfo in symbolLayers {
+            do {
+                let tempLayer = try layer(withId: layerInfo.id) as SymbolLayer
+
+                try updateLayer(withId: tempLayer.id) { (layer: inout SymbolLayer) throws in
+                    layer.textField = .expression(expression)
+                }
+            } catch {
+                Log.error(forMessage: "Error localizing textField for Symbol Layer with ID: \(layerInfo.id)", category: "Style")
+            }
+        }
+    }
+}
--- a/Sources/MapboxMaps/Style/Style+Localization.swift
+++ b/Sources/MapboxMaps/Style/Style+Localization.swift
@@ -14,35 +14,28 @@ extension Style {
 
         /// Expression to be applied to the `SymbolLayer.textField`to localize the language
         /// Sample Expression JSON: `["format",["coalesce",["get","name_en"],["get","name"]],{}]`
-        let newVal = "[\"get\",\"name_\(getLocaleValue(locale: locale))\"]"
+        let replacement = "[\"get\",\"name_\(getLocaleValue(locale: locale))\"]"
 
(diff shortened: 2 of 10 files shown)


Sources/MapboxMaps/Foundation/MapView.swift
```diff
@@ -2,6 +2,7 @@
 @_exported import MapboxCommon
 @_implementationOnly import MapboxCoreMaps_Private
 @_implementationOnly import MapboxCommon_Private
+import Foundation
```
nishant-karajgikar, line 5: unintentional?

Sources/MapboxMaps/Foundation/MapView.swift
```diff
@@ -34,6 +35,13 @@ open class MapView: UIView {
     /// Controls the addition/removal of annotations to the map.
     public internal(set) var annotations: AnnotationOrchestrator!
 
+    /// Property that describes the current language for `SymbolLayer.textField`
+    public var locale: Locale = Locale.current {
+        didSet {
+            mapboxMap.style.localizeLabels(into: locale)
+        }
+    }
```
nishant-karajgikar, line 43: Could this be an unsupported Locale?

Sources/MapboxMaps/Foundation/MapView.swift
neelmistry94: Thought I needed it for `Locale`

Sources/MapboxMaps/Foundation/MapView.swift
neelmistry94: It can, but in that case, we will default to the local case if there is no data that exists

nishant-karajgikar: We may want to make that behavior more explicit.

neelmistry94: I could add a check when passed in to see if it is under one of our supported languages. If so, then continue, otherwise don't do anything

Sources/MapboxMaps/Style/Style+Localization.swift
```diff
@@ -0,0 +1,88 @@
...
+                    let data = stringExpression.data(using: .utf8)
+                    let convertedExpression = try JSONDecoder().decode(Expression.self, from: data!)
+
+                    try updateLayer(withId: tempLayer.id) { (layer: inout SymbolLayer) throws in
+                        layer.textField = .expression(convertedExpression)
+                    }
+                }
+            } catch {
+                Log.warning(forMessage: "Error localizing textField for Symbol Layer with ID: \(layerInfo.id)", category: "Style")
+            }
```
nishant-karajgikar, line 43: Considering that this block of code is going to be executed many times, we should think about the performance aspect of it..

Using `layer(withId:)` and `updateLayer(withId:)` would result in quite a few roundtrips to JSON and back. Instead could we drop down to the lower level api (i.e. `style.setStyleLayerProperty(..)`) ?

Sources/MapboxMaps/Style/Style+Localization.swift
```diff
@@ -0,0 +1,88 @@
+@_implementationOnly import MapboxCommon_Private
+
+/// Support for localization of labels
+extension Style {
+
+    /// This function creates an expression that will localize the `textField` property of a `SymbolLayer`
+    /// - Parameter locale: A `SupportedLanguage` based `Locale`
+    internal func localizeLabels(into locale: Locale) {
```
nishant-karajgikar, line 8: Nav SDK has a requirement that not all labels are localized I think.. They want to exempt road labels for example from any localization.

To help them do this, it's probably best to break this function up into some functions like this:

```swift
public func localizeAllLabels(into locale: Locale) { .. }

public func localizeLabel(withId id: String, into locale: Locale) { .. } 

```

Sources/MapboxMaps/Style/SupportedLanguage.swift
```diff
@@ -0,0 +1,45 @@
...
+    /// Support for Russian (if available)
+    case russian = "ru"
+
+    /// Support for Simplified Chinese (if available)
+    case simplifiedChinese = "zh-Hans"
+
+    /// Support for Spanish (if available)
+    case spanish = "es"
+
+    /// Support for Traditional Chinese (if available)
```
julianrex, line 40: Can we make a note that this is not supported in mapbox.mapbox-streets-v7?

Sources/MapboxMaps/Style/Style+Localization.swift
```diff
@@ -0,0 +1,88 @@
...
+        let expressionCoalesce = try! NSRegularExpression(pattern: "\\[\"get\",\\s*\"(name_.{2,7})\"\\]",
+                                                          options: .caseInsensitive)
+        let expressionAbbr = try! NSRegularExpression(pattern: "\\[\"get\",\\s*\"abbr\"\\]",
+                                                      options: .caseInsensitive)
+
+        for layerInfo in symbolLayers {
+            do {
+                let tempLayer = try layer(withId: layerInfo.id) as SymbolLayer
+
+                if var stringExpression = String(data: try! JSONEncoder().encode(tempLayer.textField), encoding: .utf8),
```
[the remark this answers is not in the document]
neelmistry94: I think that is fine. Falls in line with the rest of style

Sources/MapboxMaps/Style/SupportedLanguage.swift
neelmistry94: sure

julianrex:
Noting that this PR does not support [`Locale.preferredLanguages`](https://developer.apple.com/documentation/foundation/nslocale/1415614-preferredlanguages) (which < 10.0.0 did). This is something that we should address in a later PR @tobrun @sbma44

Sources/MapboxMaps/Style/Style+Localization.swift
neelmistry94: @julianrex @tobrun This feels like something we need to align on. This functionality does not exist on Android

Sources/MapboxMaps/Style/Style+Localization.swift
```diff
@@ -0,0 +1,83 @@
...
+extension Style {
+
+    /// This function creates an expression that will localize the `textField` property of a `SymbolLayer`
+    /// - Parameter locale: A `SupportedLanguage` based `Locale`
+    internal func localizeLabels(into locale: Locale) throws {
+
+        /// Get all symbol layers that are currently on the map
+        let symbolLayers = allLayerIdentifiers.filter { layer in
+            return layer.type == .symbol
+        }
```
1ec5, line 13: Not every symbol layer should be forced to have a `name`-based `text-field`, which is specific to the Mapbox Streets source. For example, consider a symbol layer backed by a vector tileset or GeoJSON source with the labelable text in `NAME` instead of `name`. Now localizing the rest of the map will blow away any labeling of this non-Streets source.

Sources/MapboxMaps/Style/Style+Localization.swift
```diff
@@ -0,0 +1,83 @@
...
+
+        for sourceInfo in vectorSources {
+            do {
+                let tempSource = try source(withId: sourceInfo.id) as VectorSource
+
+                if tempSource.url?.contains("mapbox.mapbox-streets-v7") == true {
+                    if locale.identifier.contains("zh") {
+                        // v7 styles does not support value of "name_zh-Hant"
+                        if locale.identifier == SupportedLanguage.traditionalChinese.rawValue {
```
1ec5, line 55: Replace SupportedLanguage with an array of `Locale`s to avoid unchecked usage of raw values.

Sources/MapboxMaps/Style/Style+Localization.swift
```diff
@@ -0,0 +1,83 @@
...
+        }
+
+        /// Expression to be applied to the `SymbolLayer.textField`to localize the language
+        /// Sample Expression JSON: `["format",["coalesce",["get","name_en"],["get","name"]],{}]`
+        let replacement = "[\"get\",\"name_\(getLocaleValue(locale: locale))\"]"
+
+        let expressionCoalesce = try NSRegularExpression(pattern: "\\[\"get\",\\s*\"(name_.{2,7})\"\\]",
+                                                         options: .caseInsensitive)
+        let expressionAbbr = try NSRegularExpression(pattern: "\\[\"get\",\\s*\"abbr\"\\]",
+                                                     options: .caseInsensitive)
```
1ec5, line 22: It’s unnecessary to convert the entire expression to JSON, then to a string, then monkeypatch the string. That defeats the purpose of having a structured, type-safe representation of an expression.

Consider adapting the implementation in mapbox/mapbox-navigation-ios#2933 that recurses into the expression, surgically replacing occurrences of `name` or `name_*` with the coalescing expression. That approach is much more reliable and maintainable than trying to account for the variety of possible `text-field` values with finding-and-replacing in source code.

Sources/MapboxMaps/Foundation/MapView.swift
```diff
@@ -34,6 +35,17 @@ open class MapView: UIView {
     /// Controls the addition/removal of annotations to the map.
     public internal(set) var annotations: AnnotationOrchestrator!
 
+    /// Property that describes the current language for `SymbolLayer.textField`
+    public var locale: Locale = Locale.current {
```
1ec5, line 39: `Locale.current` only matches the language that the iOS UI is currently displayed in, which can differ from both the surrounding application and the user’s preference. `Bundle.preferredLocalizations` has the advantage of matching the language that the rest of the application is currently displayed in. `Locale.preferredLanguages` has the advantage of matching the user’s preferred content language settings, even if the application is unavailable in the preferred language.

Sources/MapboxMaps/Style/Style+Localization.swift
```diff
@@ -0,0 +1,83 @@
...
+    internal func getLocaleValue(locale: Locale) -> String {
+        let vectorSources = allSourceIdentifiers.filter { source in
+            return source.type == .vector
+        }
+
+        for sourceInfo in vectorSources {
+            do {
+                let tempSource = try source(withId: sourceInfo.id) as VectorSource
+
+                if tempSource.url?.contains("mapbox.mapbox-streets-v7") == true {
```
1ec5, line 52: The tileset ID always occurs in the hostname part of the URL, so compare against `URL.host` for additional robustness.

1ec5 (commented):
Thank you for implementing this important feature. Here’s some belated feedback based on lessons learned in the past.

PR #511, opened 2021-06-30
Add docs for v10.0.0-rc.3 [production]

macdrevx:


colleenmcginnis (approved):
Looks good to me @macdrevx. Here's what I check for:

✅ new version folder name matches the pattern of other version folders
✅ opened index.html locally and it looks as expected
✅ didn't see any methods i know to be private being documented
✅ didn't see any unintentional files added/removed/modified outside the new version folder

PR #512, opened 2021-06-30
Allow localization by Layer ID

neelmistry94:
## Pull request checklist:

### Summary of changes
This PR removes unneeded abstraction for the Localization feature by removing state in mapView and also removing the Supported Language Enum since they were not being used. Furthermore, this change introduces the ability to localize only a subset of layers by passing in an array of layer Ids to the localize function.


Sources/MapboxMaps/Style/Style+Localization.swift
```diff
@@ -47,13 +59,18 @@ extension Style {
...
+                    if tempSource.url?.contains("mapbox.mapbox-streets-v7") == true {
                         // v7 styles does not support value of "name_zh-Hant"
-                        if locale.identifier == SupportedLanguage.traditionalChinese.rawValue {
-                            return SupportedLanguage.chinese.rawValue
+                        if locale.identifier == "zh-Hant" {
+                            return "zh"
+                        }
+                    } else if tempSource.url?.contains("mapbox.mapbox-streets-v8") == true {
+                        /// Return simplified chinese if the Locale is Taiwan
+                        if locale.identifier == "zh_Hant_TW" {
```
Chaoba, line 72: Taiwan is using Traditional Chinese

Apps/Examples/Examples/All Examples/LocalizationExample.swift
```diff
@@ -49,27 +49,32 @@ public class LocalizationExample: UIViewController, ExampleProtocol {
...
 
         alert.addAction(UIAlertAction(title: "English", style: .default, handler: { [weak self] _ in
             guard let self = self else { return }
-            self.mapView.locale = Locale(identifier: SupportedLanguage.english.rawValue)
+            try! self.mapView.mapboxMap.style.localizeLabels(into: Locale(identifier: "en"))
+        }))
+
+        alert.addAction(UIAlertAction(title: "Japanese - Country's Only", style: .default, handler: { [weak self] _ in
+            guard let self = self else { return }
+            try! self.mapView.mapboxMap.style.localizeLabels(into: Locale(identifier: "ja"), for: ["country-label"])
```
macdrevx, line 77:
```suggestion
            try! self?.mapView.mapboxMap.style.localizeLabels(into: Locale(identifier: "ja"), for: ["country-label"])
```

Use optional chaining to achieve the same behavior more succinctly.

CHANGELOG.md
```diff
@@ -4,6 +4,13 @@ Mapbox welcomes participation and contributions from everyone.
 
 ## main
 
+### Breaking changes ⚠️
+* Removed the `SupportedLanguage` enum. You may now use `Locale(identifier: String)` as intended. ([512](https://github.com/mapbox/mapbox-maps-ios/pull/512))
+* Removed the `MapView.locale` property. Now, in order to localize values, you must call `mapView.mapboxMap.style.localizeLabels(into: Locale)`. ([512](https://github.com/mapbox/mapbox-maps-ios/pull/512))
+
+### Features ✨ and improvements 🏁
+* Expanded `localizeLabels(into: Locale)` to accept a `[String]`. This array will contain a list of layer Ids that you will want to localize. ([512](https://github.com/mapbox/mapbox-maps-ios/pull/512))
```
macdrevx, line 12:
```suggestion
* Removed the `SupportedLanguage` enum. You may now use `Locale(identifier: String)` as intended. ([#512](https://github.com/mapbox/mapbox-maps-ios/pull/512))
* Removed the `MapView.locale` property. Now, in order to localize values, you must call `mapView.mapboxMap.style.localizeLabels(into: Locale)`. ([#512](https://github.com/mapbox/mapbox-maps-ios/pull/512))

### Features ✨ and improvements 🏁
* Expanded `localizeLabels(into: Locale)` to accept a `[String]`. This array will contain a list of layer ids that you want to localize. ([#512](https://github.com/mapbox/mapbox-maps-ios/pull/512))
```

Sources/MapboxMaps/Foundation/MapView.swift
```diff
@@ -33,16 +33,16 @@ open class MapView: UIView {
...
+//    /// Property that describes the current language for `SymbolLayer.textField`
+//    public var locale: Locale = Locale.current {
+//        didSet {
+//            do {
+//                try mapboxMap.style.localizeLabels(into: locale)
+//            } catch {
+//                Log.warning(forMessage: "Error when localizing labels", category: "Style")
+//            }
+//        }
+//    }
```
macdrevx, line 45:
```suggestion
```

Sources/MapboxMaps/Style/Style+Localization.swift
```diff
@@ -5,13 +5,25 @@ extension Style {
 
     /// This function creates an expression that will localize the `textField` property of a `SymbolLayer`
     /// - Parameter locale: A `SupportedLanguage` based `Locale`
```
macdrevx, line 7: I think this needs to be updated

Sources/MapboxMaps/Style/Style+Localization.swift
```diff
@@ -5,13 +5,25 @@ extension Style {
...
-    internal func localizeLabels(into locale: Locale) throws {
+    public func localizeLabels(into locale: Locale, for layerIds: [String]? = nil) throws {
+
+        /// Do nothing if we do not support the locale
+        if !supportedLocales.contains(locale) {
+            return
```
macdrevx, line 12: Would it be better to throw?

Also, `///` is for documentation comments. Normal comments can use `//`

Sources/MapboxMaps/Style/Style+Localization.swift
```diff
@@ -5,13 +5,25 @@ extension Style {
 
     /// This function creates an expression that will localize the `textField` property of a `SymbolLayer`
     /// - Parameter locale: A `SupportedLanguage` based `Locale`
-    internal func localizeLabels(into locale: Locale) throws {
+    public func localizeLabels(into locale: Locale, for layerIds: [String]? = nil) throws {
+
+        /// Do nothing if we do not support the locale
+        if !supportedLocales.contains(locale) {
```
macdrevx, line 11: This appears to be the only place where `supportedLocales` is referenced. Rather than declaring it as a global constant, declare it in this method as a local constant to avoid polluting the global namespace.

macdrevx: Also, what do you think of making `supportedLocales` an array of locale identifiers instead of an array of locales? Then this line would become
```suggestion
        if !supportedLocaleIdentifiers.contains(locale.identifier) {
```

macdrevx: also, we should think carefully about whether we mean locale identifier or [language code](https://developer.apple.com/documentation/foundation/locale/2293346-languagecode).

macdrevx (changes requested):
Take a close look at the docs on Locale. This PR is using Locale.identifier in situations where some of the other properties may be more appropriate.

Sources/MapboxMaps/Style/Style+Localization.swift
```diff
@@ -4,16 +4,29 @@
 extension Style {
 
     /// This function creates an expression that will localize the `textField` property of a `SymbolLayer`
-    /// - Parameter locale: A `SupportedLanguage` based `Locale`
-    internal func localizeLabels(into locale: Locale) throws {
+    /// - Parameter locale: The`Locale` to update the map to
+    /// - Parameter layerIds: An optional list of ids that need to be localized. If nothing is provided, all layers will be updated
```
macdrevx, line 8:
```suggestion
    /// - Parameter layerIds: An optional list of ids that need to be localized. If `nil` is provided, all layers will be updated
```
This may be more clear so that folks are less likely to think they can pass `[]` to update all layers

Sources/MapboxMaps/Style/Style+Localization.swift
```diff
@@ -4,16 +4,29 @@
...
+    public func localizeLabels(into locale: Locale, for layerIds: [String]? = nil) throws {
```
macdrevx, line 9:
```suggestion
    public func localizeLabels(into locale: Locale, forLayerIds layerIds: [String]? = nil) throws {
```

See "Compensate for weak type information" in the [Swift API Design Guidelines](https://swift.org/documentation/api-design-guidelines/)

Sources/MapboxMaps/Style/Style+Localization.swift
cmilack: Is there any value in exposing the supported locale identifiers publicly? It seems like it could be beneficial to provide clients with strongly-typed values to use rather than having them manually specify locale ids (strings) and handle exception (or undefined behavior) if/when they aren't supported. I just realized that I am essentially making an argument in favor of the enum that was removed below. I have no insight into what prompted these changes, or some of the specific use cases, so take this with a grain of salt

macdrevx: Also, couldn't folks be using custom tile sources that support locales outside of our pre-defined set? (How) Do we handle that case?

Sources/MapboxMaps/Style/Style+Localization.swift
cmilack: Is there a style guide or best practice we're using to determine if/when we use argument labels? I guess I'm trying to understand how I should be choosing/evaluating these two options:

`public func localizeLabels(into locale: Locale, forLayerIds layerIds: [String]? = nil) throws`

`public func localizeLabels(locale: Locale, layerIds: [String]? = nil) throws`

Apps/Examples/Examples/All Examples/LocalizationExample.swift
```diff
@@ -48,28 +48,27 @@ public class LocalizationExample: UIViewController, ExampleProtocol {
...
         alert.addAction(UIAlertAction(title: "Arabic", style: .default, handler: { [weak self] _ in
-            guard let self = self else { return }
-            self.mapView.locale = Locale(identifier: SupportedLanguage.arabic.rawValue)
+            try! self?.mapView.mapboxMap.style.localizeLabels(into: Locale(identifier: "ar"))
         }))
 
         alert.addAction(UIAlertAction(title: "English", style: .default, handler: { [weak self] _ in
-            guard let self = self else { return }
-            self.mapView.locale = Locale(identifier: SupportedLanguage.english.rawValue)
+            try! self?.mapView.mapboxMap.style.localizeLabels(into: Locale(identifier: "en-us"))
```
macdrevx, line 67: Why `en-us` here?

Sources/MapboxMaps/Style/Style+Localization.swift
```diff
@@ -40,20 +51,33 @@ extension Style {
 
     /// Filters through source to determine supported locale styles.
     /// This is needed for v7 support
-    internal func getLocaleValue(locale: Locale) -> String {
+    internal func getLocaleValue(locale: Locale) -> String? {
+        // Reference for supported languages: https://docs.mapbox.com/ios/navigation/guides/localization-and-internationalization/#supported-languages
```
macdrevx, line 55: this links into the docs for navigation. is maps limited to these languages as well?

Sources/MapboxMaps/Style/Style+Localization.swift
```diff
@@ -40,20 +51,33 @@ extension Style {
...
+        let supportedLocaleIdentifiers = ["ar", "zh", "en", "fr", "de", "it", "ja", "ko", "pt", "ru", "zh-Hans", "es", "zh-Hant", "vi"]
+
+        // Do nothing if we do not support the locale
+        if !supportedLocaleIdentifiers.contains(locale.languageCode!) {
```
macdrevx, line 59: What will happen for locale identifiers starting in `zh`?

Sources/MapboxMaps/Style/Style+Localization.swift
```diff
@@ -40,20 +51,33 @@ extension Style {
...
+
         let vectorSources = allSourceIdentifiers.filter { source in
             return source.type == .vector
         }
 
         for sourceInfo in vectorSources {
             do {
-                let tempSource = try source(withId: sourceInfo.id) as VectorSource
+                if locale.identifier.starts(with: "zh") {
+                    let tempSource = try source(withId: sourceInfo.id) as VectorSource
```
macdrevx, line 70: why is this variable named `tempSource`? What's `temp` about it?

Sources/MapboxMaps/Style/Style+Localization.swift
```diff
@@ -40,20 +51,33 @@ extension Style {
...
-                        if locale.identifier == SupportedLanguage.traditionalChinese.rawValue {
-                            return SupportedLanguage.chinese.rawValue
+                        if locale.identifier == "zh-Hant" {
+                            return "zh"
+                        }
+                    } else if tempSource.url?.contains("mapbox.mapbox-streets-v8") == true {
+                        // Return traditional chinese if the Locale is Taiwan
+                        if locale.identifier == "zh-Hant-TW" {
+                            return "zh-Hant"
                         }
```
macdrevx, line 81: Why is TW the only region code that we need to strip?

macdrevx: Do we support other region codes?

Sources/MapboxMaps/Style/Style+Localization.swift
```diff
@@ -62,6 +86,11 @@ extension Style {
             }
         }
 
+        if supportedLocaleIdentifiers.contains(locale.languageCode!) &&
+            !supportedLocaleIdentifiers.contains(locale.identifier) {
+            return locale.languageCode!
+        }
```
macdrevx, line 92: The condition on line 89 is redundant with the one on line 59

Tests/MapboxMapsTests/Style/StyleIntegrationTests.swift
```diff
@@ -129,4 +129,23 @@ internal class StyleIntegrationTests: MapViewIntegrationTestCase {
...
+        var localeValue = style.getLocaleValue(locale: locale)
+
+        XCTAssertEqual(localeValue, "es")
+
+        // Case: when script and region codes does not exist
+        locale = Locale(identifier: "en-US")
+        localeValue = style.getLocaleValue(locale: locale)
+
+        XCTAssertEqual(localeValue, "en")
+    }
```
macdrevx, line 150: other cases to check:
* handling of zh-Hans
* handling of zh-Hant
* handling of the special cases for streets v7 and streets v8

macdrevx: * handling of unsupported locales

macdrevx:
Are there any tests for `localizeLabels` method itself?

Sources/MapboxMaps/Style/Style+Localization.swift
neelmistry94: I'm not sure what the requirements are around that. But that would be another ticket entirely given that is different than Android and hydrogen

Apps/Examples/Examples/All Examples/LocalizationExample.swift
neelmistry94: Whoops, forgot to remove test code

Sources/MapboxMaps/Style/Style+Localization.swift
neelmistry94: This is the only online docs that exist as a list of supported language.

Sources/MapboxMaps/Style/Style+Localization.swift
neelmistry94: I've always used 'temp` for throw away vars.

Sources/MapboxMaps/Style/Style+Localization.swift
neelmistry94: not as of now

neelmistry94:
> Are there any tests for `localizeLabels` method itself?

No, I feel like there isn't a lot of value in testing that function. `localizeLabels` leans on style gets/updates, and the another localization function which has been tested

Apps/Examples/Examples/All Examples/LocalizationExample.swift
```diff
@@ -48,28 +48,27 @@ public class LocalizationExample: UIViewController, ExampleProtocol {
...
+            try! self?.mapView.mapboxMap.style.localizeLabels(into: Locale(identifier: "ar"))
         }))
 
         alert.addAction(UIAlertAction(title: "English", style: .default, handler: { [weak self] _ in
-            guard let self = self else { return }
-            self.mapView.locale = Locale(identifier: SupportedLanguage.english.rawValue)
+            try! self?.mapView.mapboxMap.style.localizeLabels(into: Locale(identifier: "en"))
+        }))
+
+        alert.addAction(UIAlertAction(title: "Japanese - Country's Only", style: .default, handler: { [weak self] _ in
```
macdrevx, line 70: What does `Country's Only` mean?

macdrevx: ah, maybe you mean `Countries Only`

macdrevx:
```suggestion
        alert.addAction(UIAlertAction(title: "Japanese - Countries Only", style: .default, handler: { [weak self] _ in
```

CHANGELOG.md
```diff
@@ -10,12 +10,15 @@ Mapbox welcomes participation and contributions from everyone.
...
 
 ### Breaking changes ⚠️
 
 * `TileRegionError` has a new case `tileCountExceeded(String)`. ([#522](https://github.com/mapbox/mapbox-maps-ios/pull/522))
 * FlyToCameraAnimator.state will now be `.inactive` after it completes or is stopped. This change makes its behavior consistent with the behavior of `BasicCameraAnimator`. ([#519](https://github.com/mapbox/mapbox-maps-ios/pull/519))
 * Completion blocks added to `BasicCameraAnimator` will no longer be invoked as a side-effect of deinitialization. ([#519](https://github.com/mapbox/mapbox-maps-ios/pull/519))
+* Removed the `SupportedLanguage` enum. You may now use `Locale(identifier: String)` as intended. ([512](https://github.com/mapbox/mapbox-maps-ios/pull/512))
+* Removed the `MapView.locale` property. Now, in order to localize values, you must call `mapView.mapboxMap.style.localizeLabels(into: Locale)`. ([512](https://github.com/mapbox/mapbox-maps-ios/pull/512))
```
macdrevx, line 21: These are missing the `#` before the issue number.

CHANGELOG.md
```diff
@@ -10,12 +10,15 @@ Mapbox welcomes participation and contributions from everyone.
   Note: This change will bring following changes for CJK text block:
   1. For vertical CJK text, all the characters including Latin and Numbers will be vertically placed now. Previously, Latin and Numbers are horizontally placed.
   2. For horizontal CJK text, it may have a slight horizontal shift due to the anchor shift.
+* Expanded `localizeLabels(into: Locale)` to accept a `[String]`. This array will contain a list of layer Ids that you will want to localize. ([512](https://github.com/mapbox/mapbox-maps-ios/pull/512))
```
macdrevx, line 13: I'd lowercase the I in Ids.

Tests/MapboxMapsTests/Style/StyleIntegrationTests.swift
```diff
@@ -129,4 +129,48 @@ internal class StyleIntegrationTests: MapViewIntegrationTestCase {
 
         wait(for: [expectation], timeout: 5.0)
     }
+
+    func testGetLocaleValue() {
```
macdrevx, line 133: Let's split this up into multiple test methods. Each test should focus on one clearly defined behavior that you want to validate.

macdrevx: You can even name the test methods in a way that describes the behavior you're validating

macdrevx: Doing so makes the tests easier to maintain if we need to refactor them later, because it provides extra context that helps a future developer understand why we're asserting what we're asserting.

Tests/MapboxMapsTests/Style/StyleIntegrationTests.swift
macdrevx: Would still love to see a test for `zh-Hans`

Sources/MapboxMaps/Style/Style+Localization.swift
```diff
@@ -62,6 +86,10 @@ extension Style {
             }
         }
 
+        if !supportedLocaleIdentifiers.contains(locale.identifier) {
+            return locale.languageCode!
+        }
+
         return locale.identifier
```
macdrevx, line 93: I think this would be more clear if you negate the logic and write it with a ternary operator:

```suggestion
    return supportedLocaleIdentifiers.contains(locale.identifier) ? locale.identifier : locale.languageCode!
```

Sources/MapboxMaps/Style/Style+Localization.swift
```diff
@@ -40,20 +51,33 @@ extension Style {
...
+
         let vectorSources = allSourceIdentifiers.filter { source in
             return source.type == .vector
         }
 
         for sourceInfo in vectorSources {
             do {
-                let tempSource = try source(withId: sourceInfo.id) as VectorSource
+                if locale.identifier.starts(with: "zh") {
+                    let tempSource = try source(withId: sourceInfo.id) as VectorSource
```
macdrevx, line 70: If this throws, we log a warning. I don't think that's always necessarily correct. Isn't it possible for the operation to succeed even if this throws?

macdrevx: Also, under what circumstances does this method throw? Is it only if there is no source with the specified ID? If that's the only case, is that even possible here?

Sources/MapboxMaps/Style/Style+Localization.swift
```diff
@@ -40,20 +51,33 @@ extension Style {
 
     /// Filters through source to determine supported locale styles.
     /// This is needed for v7 support
-    internal func getLocaleValue(locale: Locale) -> String {
+    internal func getLocaleValue(locale: Locale) -> String? {
+        // Reference for supported languages: https://docs.mapbox.com/ios/navigation/guides/localization-and-internationalization/#supported-languages
+        let supportedLocaleIdentifiers = ["ar", "zh", "en", "fr", "de", "it", "ja", "ko", "pt", "ru", "zh-Hans", "es", "zh-Hant", "vi", "zh-Hant-TW"]
```
macdrevx, line 56: Would love to see this array be sorted for readability.

```suggestion
        let supportedLocaleIdentifiers = ["ar", "de", "en", "es", "fr", "it", "ja", "ko", "pt", "ru", "vi", "zh", "zh-Hans", "zh-Hant", "zh-Hant-TW"]
```

Sources/MapboxMaps/Style/Style+Localization.swift
macdrevx: I thought the goal was just to link to some generic reference documentation in case someone wanted to look up the definition of one of these locales. https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPInternational/LanguageandLocaleIDs/LanguageandLocaleIDs.html points to http://www.loc.gov/standards/iso639-2/php/English_list.php for language codes, https://www.iso.org/obp/ui/#search for country codes, and http://www.unicode.org/iso15924/iso15924-codes.html for the script designators. Perhaps just a link to the apple doc would be sufficient to help someone head in the right direction.

macdrevx:
> > Are there any tests for localizeLabels method itself?

> No, I feel like there isn't a lot of value in testing that function. localizeLabels leans on style gets/updates, and the another localization function which has been tested

I feel strongly that it should have some tests. It's a 40-line method that does some pretty complex logic including the use of regular expressions.

Sources/MapboxMaps/Style/Style+Localization.swift
neelmistry94: But then they might pass in something that is not supported which would result in unexpected behavior. I think documenting based on guaranteed Mapbox support would be the better experience

neelmistry94: This is also an internal function

macdrevx: > This is also an internal function

Exactly. I think the point of this documentation is to help someone look up the definition of one of the identifiers in the array, not provide a canonical list of supported locales.

Tests/MapboxMapsTests/Style/StyleIntegrationTests.swift
```diff
@@ -129,4 +129,72 @@ internal class StyleIntegrationTests: MapViewIntegrationTestCase {
 
         wait(for: [expectation], timeout: 5.0)
     }
+
+    func testGetLocaleValueBaseCase() {
+        guard let style = style else {
+            XCTFail("There should be valid MapView and Style objects created by setUp.")
+            return
+        }
```
macdrevx, line 137: I'd remove these guards and just optional chain on style.

macdrevx: we control what happens in setup, so we can depend on that here.

Sources/MapboxMaps/Style/Style+Localization.swift
neelmistry94: fair, that makes sense. Maybe our guides should have a list of supported languages

macdrevx (approved):

Issue #474, opened 2021-06-15
Panning while zooming not working as expected


georgbachmann:
## Environment

- Xcode version: 12.5.0
- iOS version: 14.6
- Devices affected: All
- Maps SDK Version: 10.0.rc1

## Observed behavior and steps to reproduce

I just tried out the new V10 and immediatly noticed that I can move around in the map any more as I am used to. I can't zoom (spread two fingers apart) while panning (move my two fingers across the screen).
The old version of Mapbox could to that. Live every scrollview on iOS does.

I noticed it somehow works when I start with one finger to pan and then add the second to zoom, but it still feels very strange to me... Not like other iOS scrollviews react.
Or am I just missing something?

## Expected behavior

Map pans AND zooms at the same time

patrickkempff:
I noticed this as well

cmilack:
I think this is happening because `maximumNumberOfTouches == 1 ` on the underlying `UIPanGestureRecognizer` in `PanGestureHandler.swift`.  Removing that setting (and using the default of UINT_MAX) seems to resolve the issue, however, I think this will require additional investigation to determine whether there will be any side effects from this change.

cmilack:
> I think this is happening because `maximumNumberOfTouches == 1 ` on the underlying `UIPanGestureRecognizer` in `PanGestureHandler.swift`. Removing that setting (and using the default of UINT_MAX) seems to resolve the issue, however, I think it will require additional investigation to determine whether there will be any side effects from this change.

This turned out to be incorrect and only partially resolved this issue. The real problem was that we were not shifting the map appropriately during zoom/pinch.

PR #517, opened 2021-07-02
Enable panning while zooming

cmilack:

Fixes: https://github.com/mapbox/mapbox-maps-ios/issues/474

## Pull request checklist:

### Summary of changes

This PR enables panning while zooming, which is something that existed in the legacy SDK. The underlying issue was that the map was not being recentered as part of pinch (zoom) gesture handling.


Sources/MapboxMaps/Gestures/GestureManager.swift
```diff
@@ -97,6 +97,10 @@ public final class GestureManager: NSObject {
         if let pitchHandler = gestureHandlers[.pitch], let panHandler = gestureHandlers[.pan] {
             requireGestureToFail(allowedGesture: pitchHandler, failableGesture: panHandler)
         }
+        
+        if let pinchHandler = gestureHandlers[.pinch], let panHandler = gestureHandlers[.pan] {
```
cmilack, line 101: Need to add a test for this...

Sources/MapboxMaps/Gestures/GestureHandlers/PinchGestureHandler.swift
```diff
@@ -34,12 +37,20 @@ internal class PinchGestureHandler: GestureHandler {
              TODO: Handle a concurrent rotate gesture here.
              Prioritize the correct gesture by comparing the velocity of competing gestures.
              */
-
+            previousPinchCenterPoint = pinchCenterPoint
         } else if pinchGestureRecognizer.state == .changed {
-
+
```
neelmistry94, line 42: nit: whitespace

cmilack: Good catch. Linter wasn't set up correctly and I miss several of these.

Sources/MapboxMaps/Gestures/GestureManager+GestureHandlerDelegate.swift
```diff
@@ -86,6 +86,20 @@ extension GestureManager: GestureHandlerDelegate {
...
+            return
+        }
+
+        let currentCenterCoordinate = mapView.mapboxMap.cameraState.center
+        let currentCenterScreenPoint = mapView.mapboxMap.point(for: currentCenterCoordinate)
+        let newCenterScreenPoint = currentCenterScreenPoint.shifted(xOffset: -offset.width,
+                                                                    yOffset: -offset.height)
+        let newCenterCoordinate = mapView.mapboxMap.coordinate(for: newCenterScreenPoint)
+        mapView.mapboxMap.setCamera(to: CameraOptions(center: newCenterCoordinate))
+    }
```
neelmistry94, line 101: Instead of using a setCamera, I'd like to see us use the Drag API for all panning on the map. @cmilack Let me know if you need help with this. But there should be a function called drag that you can use

cmilack: I updated this to use the drag API but it's creating some awkward behavior. It seems to over-compensate when panning which causes the map to shift much more than expected.  I've tried a variety of things to get this right and the implementation I had (w/o drag API) seems to work the best.  From what I've seen, I think that the DragAPI is just not behaving nicely when we're zooming simultaneously. I'm happy to try to dig deeper but not sure how much time/effort I should be putting in here.


1. Two setCamera calls (panning w/o drag API):  works as expected w/ smooth panning and scrolling

2. Two setCamera calls (panning w/ drag API): map pans more than expected and gets worse as touch points are further apart.

3. Single Camera update using dragOptions (drag API) and updating zoom:  similar result to example 2 but with more jitter

CHANGELOG.md
```diff
@@ -24,6 +24,10 @@ Mapbox welcomes participation and contributions from everyone.
 * Fix threading issues in HTTP file source. ([#522](https://github.com/mapbox/mapbox-maps-ios/pull/522))
 * Fixed an issue that could cause flickering during ease to and basic animations ([#519](https://github.com/mapbox/mapbox-maps-ios/pull/519))
 * Fixed an issue that could result in ease to and basic animations never reaching their final values ([#519](https://github.com/mapbox/mapbox-maps-ios/pull/519))
+* Fixed an issue where panning was not enabled while zooming. ([#474](https://github.com/mapbox/mapbox-maps-ios/pull/474))
+
+
+
```
macdrevx, line 30:
```suggestion
```

Sources/MapboxMaps/Foundation/Extensions/CoreGraphics.swift
```diff
@@ -24,6 +24,18 @@ extension CGPoint {
         return CGPoint(x: origin.x + fraction * (destination.x - origin.x),
                        y: origin.y + fraction * (destination.y - origin.y))
     }
+
+    /// Returns a new 'CGPoint' adjusted by the input offsets
+    /// - Parameters:
+    ///   - xOffset: The offset in the x direction
+    ///   - yOffset: The offset in the y direction
+    /// - Returns: A new 'CGPoint' with the offets applied
+    internal func shifted(xOffset: CGFloat, yOffset: CGFloat) -> Self {
```
macdrevx, line 33:
```suggestion
    internal func shiftedBy(xOffset: CGFloat, yOffset: CGFloat) -> Self {
```

See https://swift.org/documentation/api-design-guidelines/#argument-labels

Screen Shot 2021-07-09 at 3 00 55 PM

Sources/MapboxMaps/Gestures/GestureHandlerDelegate.swift
```diff
@@ -74,7 +74,7 @@ internal extension GestureHandlerDelegate {
 
     func scaleForZoom() -> CGFloat { return 0.0 }
 
-    func pinchScaleChanged(with newScale: CGFloat, andAnchor anchor: CGPoint) {}
+    func pinchChanged(with zoom: CGFloat, anchor: CGPoint, offset: CGSize) { }
```
macdrevx, line 77: We should delete this entire extension on GestureHandlerDelegate. I don't see a reason for it given that it's an internal construct.

Sources/MapboxMaps/Gestures/GestureHandlers/PinchGestureHandler.swift
```diff
@@ -30,16 +33,22 @@ internal class PinchGestureHandler: GestureHandler {
...
-
         } else if pinchGestureRecognizer.state == .changed {
 
             let newScale = scale * pinchGestureRecognizer.scale
-            delegate.pinchScaleChanged(with: log2(newScale), andAnchor: pinchCenterPoint)
+            let newZoom = log2(newScale)
 
+            let offset = CGSize(width: pinchCenterPoint.x - previousPinchCenterPoint.x,
+                                height: pinchCenterPoint.y - previousPinchCenterPoint.y)
+            previousPinchCenterPoint = pinchCenterPoint
```
macdrevx, line 49: This class uses a pinch gesture, but I think the principle here is the same as the [recommendation](https://developer.apple.com/documentation/uikit/uipangesturerecognizer/1621207-translation) when using a pan gesture:

> The x and y values report the total translation over time. They are not delta values from the last time that the translation was reported. Apply the translation value to the state of the view when the gesture is first recognized—do not concatenate the value each time the handler is called.

Would it be possible to refactor this so that we always calculate the new camera options relative to the camera options when the gesture began, not relative to the camera options at the previous frame?

Sources/MapboxMaps/Gestures/GestureHandlers/PinchGestureHandler.swift
```diff
@@ -30,16 +33,22 @@ internal class PinchGestureHandler: GestureHandler {
...
+
+            delegate.pinchChanged(with: newZoom, anchor: pinchCenterPoint, offset: offset)
         } else if pinchGestureRecognizer.state == .ended
             || pinchGestureRecognizer.state == .cancelled {
```
macdrevx, line 53: I would imagine that the drift logic below also needs to be updated

Sources/MapboxMaps/Gestures/GestureHandlerDelegate.swift
cmilack: thats fine with me. To be honest, I was simply following the pre-established pattern here. I think a few of these default implementations are actually used so I will move them into GestureManager.

cmilack:
On hold pending discussion with GL Native

Sources/MapboxMaps/Gestures/GestureHandlers/PinchGestureHandler.swift
cmilack: There is a desire to move this implementation to use the dragAPI, instead of manually calculating a new camera center position. That API has a method to retrieve cameraOptions for the drag operation and it needs the center from the previous frame along with the new center point. That said, I think we may end up wanting to stick with something similar to what we have here. I will circle back to this though as things become clearer. I'll also be adding some additional notes below to describe the challenges related to the DragAPI.

`func dragCameraOptions(from: CGPoint, to: CGPoint) -> CameraOptions`

cmilack:
A few more details on the current state of this PR:

Ideally we would be using the drag API as part of this implementation, but during testing, I found that I wasn't getting the behavior I was expecting. After digging deeper, I found that the underlying issue is that the drag API depends on a specific "pivot" point that is supplied with the `dragStart` method.  The problem with this is, the concurrent zoom operation is going to alter the map which makes that original pivot point invalid. I will be working with the GL team to come up with appropriate solutions for this problem.

Sources/MapboxMaps/Gestures/GestureHandlerDelegate.swift
```diff
@@ -28,7 +28,7 @@ internal protocol GestureHandlerDelegate: AnyObject {
     func pinchChanged(with zoom: CGFloat, anchor: CGPoint, previousAnchor: CGPoint)
 
     // Pinch has completed with a final scale and possible drift
-    func pinchEnded(with finalScale: CGFloat, andDrift possibleDrift: Bool, andAnchor anchor: CGPoint)
+    func pinchEnded(with finalScale: CGFloat, andAnchor anchor: CGPoint)
```
neelmistry94, line 31: We should update the comment if we are removing this drift var. 

Which to that point, why are we removing the drift param

cmilack: Good Call. I removed drift because I found that it isn't actually used downstream (in the GestureManager).

cmilack:
> On hold pending discussion with GL Native

I'll be doing my own final review of this tomorrow AM and marking as 'Ready For Review'

Sources/MapboxMaps/Gestures/GestureHandlerDelegate.swift
neelmistry94: Nice! just to make sure, I want to tag in @nishant-karajgikar is there any impact with removing this drift variable? I assume not since it's not being used

cmilack: I can definitely add a new ticket for follow-up work but I'm not totally clear what the expected behavior is. Some clarification on that would be helpful too.

Sources/MapboxMaps/Gestures/GestureHandlers/PinchGestureHandler.swift
cmilack: The approach changed and I'm not sure the original comment still applies. The DragAPI requires the to/from screen points in order to handle change from the previous frame and drag location. We're no longer needing to calculate offsets either.

CHANGELOG.md
```diff
@@ -32,6 +33,9 @@ Mapbox welcomes participation and contributions from everyone.
 * Fixed an issue that could cause flickering during ease to and basic animations ([#519](https://github.com/mapbox/mapbox-maps-ios/pull/519))
 * Fixed an issue that could result in ease to and basic animations never reaching their final values ([#519](https://github.com/mapbox/mapbox-maps-ios/pull/519))
 
+
+
+
```
neelmistry94, line 38: remove whitespaces

Sources/MapboxMaps/Gestures/GestureHandlers/PinchGestureHandler.swift
neelmistry94: I actually believe none of this drift logic is needed at all anymore. Drag API handles drift for us.

neelmistry94 (dismissed):

Sources/MapboxMaps/Gestures/GestureManager.swift
```diff
@@ -101,6 +101,10 @@ public final class GestureManager: NSObject {
             requireGestureToFail(allowedGesture: pitchHandler, failableGesture: panHandler)
         }
 
+        if let pinchHandler = gestureHandlers[.pinch], let panHandler = gestureHandlers[.pan] {
+            requireGestureToFail(allowedGesture: pinchHandler, failableGesture: panHandler)
```
macdrevx, line 105: I always have to look up the implementation of this method to remember what it does. Here's the implementation:

```
    internal func requireGestureToFail(allowedGesture: GestureHandler, failableGesture: GestureHandler) {
        guard let failableGesture = failableGesture.gestureRecognizer else { return }
        allowedGesture.gestureRecognizer?.require(toFail: failableGesture)
    }
```

A couple thoughts:
1. This should be a method on GestureHandler `func require(toFail other: GestureHandler)`
2. Is it necessary for `GestureHandler.gestureRecognizer` to be optional? If not, let's just delete this method entirely and use the underlying UIGestureRecognizer method directly.

Sources/MapboxMaps/Gestures/GestureManager.swift
```diff
@@ -255,12 +257,17 @@ extension GestureManager: GestureHandlerDelegate {
         return mapboxMap.cameraState.zoom
     }
 
-    internal func pinchScaleChanged(with newScale: CGFloat, andAnchor anchor: CGPoint) {
-        mapboxMap.setCamera(to: CameraOptions(anchor: anchor, zoom: newScale))
+    internal func pinchChanged(with zoom: CGFloat, anchor: CGPoint, previousAnchor: CGPoint) {
```
macdrevx, line 260:
```suggestion
    internal func pinchChanged(withZoom zoom: CGFloat, anchor: CGPoint, previousAnchor: CGPoint) {
```

In this case, CGFloat doesn't provide sufficient context, so you're supposed to include the name…

Screen Shot 2021-07-22 at 2 47 49 PM

macdrevx: also applies to pinchEnded below

macdrevx:
does this implementation support post-gesture drift?

Sources/MapboxMaps/Gestures/GestureManager.swift
```diff
@@ -255,12 +257,17 @@ extension GestureManager: GestureHandlerDelegate {
...
+        mapboxMap.dragStart(for: previousAnchor)
+        let dragOptions = mapboxMap.dragCameraOptions(from: previousAnchor, to: anchor)
+        mapboxMap.setCamera(to: dragOptions)
+        mapboxMap.dragEnd()
+
+        mapboxMap.setCamera(to: CameraOptions(anchor: anchor, zoom: zoom + mapboxMap.cameraState.zoom))
```
macdrevx, line 266: I'm concerned that `zoom + mapboxMap.cameraState.zoom` will accumulate error. I'd like to explore a version of this that bases everything off of the zoom when the gesture began.

Tests/MapboxMapsTests/Gestures/GestureManagerTests.swift
```diff
@@ -91,10 +91,36 @@ final class GestureManagerTests: XCTestCase {
...
+                                                       shouldRecognizeSimultaneouslyWith: rotateGestureRecognizer))
+    }
+
     func testSimultaneousRotationAndPanGestures() {
         let panGestureRecognizer = UIPanGestureRecognizer()
         let rotateGestureRecognizer = UIRotationGestureRecognizer()
-        XCTAssertTrue(gestureManager.gestureRecognizer(panGestureRecognizer,
+        XCTAssertFalse(gestureManager.gestureRecognizer(panGestureRecognizer,
                                                        shouldRecognizeSimultaneouslyWith: rotateGestureRecognizer))
     }
```
macdrevx, line 125: This is now redundant with the previous test

Tests/MapboxMapsTests/Gestures/GestureManagerTests.swift
```diff
@@ -105,20 +131,27 @@ final class GestureManagerTests: XCTestCase {
                                                         shouldRecognizeSimultaneouslyWith: tapGestureRecognizer))
     }
 
-    func testPinchScaleChanged_SetsCamera() {
+    func testSimultaneousPinchAndPanGestures() {
+        let pinchGestureRecognizer = UIPinchGestureRecognizer()
+        let panGestureRecognizer = UIPanGestureRecognizer()
+        XCTAssertFalse(gestureManager.gestureRecognizer(pinchGestureRecognizer,
+                                                        shouldRecognizeSimultaneouslyWith: panGestureRecognizer))
+    }
```
macdrevx, line 139: Also redundant

Tests/MapboxMapsTests/Gestures/GestureSupportableViewMock.swift
```diff
@@ -8,15 +8,14 @@ import UIKit
...
     var tapCalled = false
     var tapCalledWithNumberOfTaps = 0
     var tapCalledWithNumberOfTouches = 0
 
     var pannedCalled = false
 
-    var pinchScaleChangedMethod: (wasCalled: Bool, newScale: CGFloat?, anchor: CGPoint?) = (false, nil, nil)
-    var pinchEndedMethod: (wasCalled: Bool, drift: Bool?, anchor: CGPoint?) = (false, nil, nil)
+    var pinchChangedMethod: (wasCalled: Bool, newZoom: CGFloat?, anchor: CGPoint?, previousAnchor: CGPoint?) = (false, nil, nil, nil)
+    var pinchEndedMethod: (wasCalled: Bool, anchor: CGPoint?) = (false, nil)
```
macdrevx, line 18: Let's continually refactor towards using `Stub`

macdrevx (dismissed):
Main thing I'd like to explore is whether we can avoid the error accumulation issue. Other things are more minor.

Sources/MapboxMaps/Gestures/GestureManager.swift
cmilack: I will make this change but I want to point out that this is pretty inconsistent across the codebase. I've made this comment before but it would be great to have clarity on why/when we want to use argument labels. For example, this change requested above results in this:

`func pinchChanged(withZoom targetZoom: CGFloat, ...)`

but that feels redundant to me and I would typically prefer this:

`func pinchChanged(targetZoom: CGFloat, ...)`

In this case, does the argument label provide any value?  I think the primary reason people use argument labels is to avoid ambiguity but in this case, I feel that the argument label isn't necessary for that. Again, I will make this change but it feels like maybe the team needs to DDC on these kind of conventions.

Sources/MapboxMaps/Gestures/GestureManager.swift
cmilack: I will circle around with folks to try and understand why these were made optional in the first place. I don't want to make that change without understanding the reason. I agree with this assessment but I'm reluctant to make the change without knowing the original intention here.

Also, this feels like it's starting to creep into some fundamental API changes around gestures that several people have acknowledged. How far do we want to take this under this PR? Is there an opportunity for follow up work here to address the big picture?

cmilack:
> does this implementation support post-gesture drift?

It does not but that was never implemented before. Is that a requirement? If so, we'll need a new ticket.

neelmistry94 (approved):

Issue #493, opened 2021-06-24
Screen flash when panning by "flicking" the map.


julianrex:
## Environment

- Xcode version: 12.4
- iOS version: 14.6
- Devices affected: All
- Maps SDK Version: 10.0.0-rc.1 / 10.0.0-rc.2

## Observed behavior and steps to reproduce

There is a known issue in 10.0.0-rc.2 that causes a momentary camera jump when "flicking" the map (pan followed by a drift). There is currently no work-around for this issue, but the underlying issue is understood and a solution is being developed.

PR #519, opened 2021-07-06
Fix animator bugs

macdrevx:
Fixes: #493

## Pull request checklist:
 - [x] Update changelog

### Summary of changes

- No breaking API changes
- Fixes an issue that could cause flickering during ease to and basic animations
    - This would happen when an animator was started and then the display link ran before the the animations added to the CameraView were flushed to the presentation tree at the end of the run loop. When the values were read out of the CameraView, the presentation layer was nil, so the to values were read from the layer tree instead, resulting in a single-frame jump to the end of the animation before jumping back to near the beginning during the next frame.
- Fixes an issue that could result in ease to and basic animations never reaching their final values
    - This was possible because BasicCameraAnimator always returned nil for its current camera options after its internal animator completed. This meant that the final rendered frame might be an intermediate interpolated value rather than the final to value.
- Includes significant refactoring to support these fixes. Notable changes include:
    - Animator completion blocks are no longer scheduled for the display link; they are invoked immediately
    - FlyToCameraAnimator's final state is now `.inactive` for consistency with `BasicCameraAnimator`
    - CameraAnimationsManager was previously split across 2 files, but has now been combined into 1
    - GestureManager was previously split across 2 files, but has now been combined into 1
    - GestureManager now depends on MapboxMap instead of on MapView
    - CameraAnimationsManager now depends on MapboxMap instead of on MapView
    - CameraViews are now pooled for reuse and added to a subview of MapView

### For discussion

- [x] Should we revert https://github.com/mapbox/mapbox-maps-ios/pull/369 which was a previous attempt to fix the flickering issues? No.
- [x] This PR uses dependency inversion when injecting MapboxMap into the animators. This results in some duplication. Do we want to use dependency inversion, or would we rather just program to interfaces? Program to interfaces per subsystem.

### Not Included

- Does not test that `CameraAnimationsManager.update()` is called during the display link. Accomplishing this would entail additional refactoring to `MapView` to allow supplying a mock implementation of `CameraAnimationsManager`. We should discuss how best to approach this.


Sources/MapboxMaps/Foundation/Camera/AnimationCompletion.swift
```diff
@@ -0,0 +1,3 @@
+import UIKit
+
+public typealias AnimationCompletion = (UIViewAnimatingPosition) -> Void
```
neelmistry94, line 3: Why does this get its own file?

Sources/MapboxMaps/Foundation/Camera/BasicCameraAnimator.swift
```diff
@@ -3,31 +3,46 @@ import CoreLocation
 
 // MARK: CameraAnimator Class
 public class BasicCameraAnimator: NSObject, CameraAnimator, CameraAnimatorInterface {
+    private enum InternalState: Equatable {
+        case initial
+        case inProgress(CameraTransition)
+        case final
```
neelmistry94, line 9: this should be called `finished`

Sources/MapboxMaps/Foundation/Camera/FlyToCameraAnimator.swift
```diff
@@ -91,3 +96,10 @@ public class FlyToCameraAnimator: NSObject, CameraAnimator, CameraAnimatorInterf
         stopAnimation()
     }
 }
+
+internal protocol FlyToCameraAnimatorMapboxMap: AnyObject {
+    func setCamera(to cameraOptions: CameraOptions)
+}
```
macdrevx, line 96: Refactor to a single shared protocol

Sources/MapboxMaps/Foundation/Camera/BasicCameraAnimator.swift
```diff
@@ -49,20 +46,19 @@ public class BasicCameraAnimator: NSObject, CameraAnimator, CameraAnimatorInterf
     }
 
     // MARK: Initializer
-    internal init(delegate: CameraAnimatorDelegate,
-                  propertyAnimator: UIViewPropertyAnimator,
+    internal init(propertyAnimator: UIViewPropertyAnimator,
                   owner: AnimationOwner,
                   mapboxMap: BasicCameraAnimatorMapboxMap,
-                  cameraView: CameraView = CameraView()) {
-        self.delegate = delegate
```
macdrevx, line 54: consider adding a precondition that the cameraView needs to be in the window.

Sources/MapboxMaps/Foundation/Camera/FlyToCameraAnimator.swift
```diff
@@ -45,7 +45,7 @@ public class FlyToCameraAnimator: NSObject, CameraAnimator, CameraAnimatorInterf
     }
 
     public func stopAnimation() {
-        state = .stopped
+        state = .inactive
```
macdrevx: Note this behavior change in the changelog;

Create a ticket about divorcing our public animator interfaces from the UIKit types

Sources/MapboxMaps/Foundation/Camera/BasicCameraAnimator.swift
```diff
@@ -64,119 +80,151 @@ public class BasicCameraAnimator: NSObject, CameraAnimator, CameraAnimatorInterf
...
-        propertyAnimator.stopAnimation(false)
-        propertyAnimator.finishAnimation(at: .current)
+        switch internalState {
+        case .initial:
+            fatalError("Attempt to stop an animation that has not started.")
+        case .inProgress:
+            propertyAnimator.stopAnimation(false)
+            propertyAnimator.finishAnimation(at: .current)
+        case .final:
+            fatalError("Attempt to stop an animation that has already completed.")
```
macdrevx, line 124: Should this just be a no-op?

Sources/MapboxMaps/Foundation/Camera/BasicCameraAnimator.swift
```diff
@@ -3,6 +3,11 @@ import CoreLocation
 
 // MARK: CameraAnimator Class
 public class BasicCameraAnimator: NSObject, CameraAnimator, CameraAnimatorInterface {
```
macdrevx, line 5: Ticket: make `isReversed` settable

macdrevx:
From discussion: finish exploring whether we can remove the requirement for callers to keep strong references to animators.

Sources/MapboxMaps/Foundation/Camera/AnimationCompletion.swift
macdrevx: There used to be a delegate protocol declared in this file. The file was named after that delegate protocol. The delegate protocol has been deleted, so I renamed the file after the `typealias` instead.

Sources/MapboxMaps/Foundation/Camera/BasicCameraAnimator.swift
macdrevx: Can you explain your rationale?

Sources/MapboxMaps/Foundation/Camera/BasicCameraAnimator.swift
macdrevx: I've decided not to do this. It makes the unit tests worse, and provides limited value.

Sources/MapboxMaps/Foundation/Camera/BasicCameraAnimator.swift
```diff
@@ -47,139 +62,170 @@ public class BasicCameraAnimator: NSObject, CameraAnimator, CameraAnimatorInterf
...
+            if let presentationCameraOptions = cameraView.presentationCameraOptions {
+                mapboxMap.setCamera(
+                    to: cameraOptions(
+                        with: transition,
+                        cameraViewCameraOptions: presentationCameraOptions))
+            }
         }
+    }
 
+    private func cameraOptions(with transition: CameraTransition, cameraViewCameraOptions: CameraOptions) -> CameraOptions {
```
nishant-karajgikar, line 169: should we rename this method? This may be just me, but i'm not sure i fully understand what it's doing from the signature?

Alternatively, we could maybe add a comment here?

Sources/MapboxMaps/Foundation/Camera/BasicCameraAnimator.swift
```diff
@@ -47,139 +62,170 @@ public class BasicCameraAnimator: NSObject, CameraAnimator, CameraAnimatorInterf
...
 
         if transition.pitch.toValue != nil {
-            cameraOptions.pitch = interpolatedCamera.pitch
+            cameraOptions.pitch = cameraViewCameraOptions.pitch
         }
 
         return cameraOptions
     }
 
+    private func createTransition() {
```
nishant-karajgikar, line 199: nit: do we want to use `makeTransition` instead of `createTransition` ?

Sources/MapboxMaps/Foundation/Camera/CameraAnimationsManager.swift
```diff
@@ -148,18 +154,123 @@ public class CameraAnimationsManager {
...
+    // MARK: Animator Functions
 
-fileprivate extension CoordinateBounds {
-    func contains(_ coordinates: [CLLocationCoordinate2D]) -> Bool {
-        let latitudeRange = southwest.latitude...northeast.latitude
-        let longitudeRange = southwest.longitude...northeast.longitude
+    /// Convenience to create a `CameraAnimator` and will add it to a list of `CameraAnimators` to track the lifecycle of that animation.
+    ///
+    /// NOTE: Keep in mind the lifecycle of a `CameraAnimator`. If a `CameraAnimator` is destroyed, before the animation is finished,
+    /// the animation will be interrupted and completion handlers will be called.
```
nishant-karajgikar, line 162: This comment needs an update since completion handlers are no longer called.

nishant-karajgikar: Same for the rest of the factory methods in this class.

Sources/MapboxMaps/Foundation/Camera/CameraAnimationsManager.swift
```diff
@@ -148,18 +154,123 @@ public class CameraAnimationsManager {
...
+            propertyAnimator: propertyAnimator,
+            owner: animationOwner,
+            mapboxMap: mapboxMap,
+            cameraView: getCameraView())
+        cameraAnimator.addAnimations(animations)
+        cameraAnimatorsSet.add(cameraAnimator)
+        return cameraAnimator
+    }
+
+    private func getCameraView() -> CameraView {
```
nishant-karajgikar, line 271: nit: should this be `makeCameraView` ?

Sources/MapboxMaps/Foundation/Camera/BasicCameraAnimator.swift
macdrevx: typically `make` prefix is for factory methods, but this method returns Void.

Sources/MapboxMaps/Foundation/Camera/BasicCameraAnimator.swift
macdrevx: Will add a comment

Sources/MapboxMaps/Foundation/Camera/BasicCameraAnimator.swift
nishant-karajgikar: ah i see, maybe we should name this something like `addTransitionToPropertyAnimator` ?

nishant-karajgikar: or we could add a comment..

nishant-karajgikar (approved):

PR #539, opened 2021-07-15
Migrate device farm to xcodegen

macdrevx:
## Pull request checklist:

### Summary of changes

Migrates device farm scripts to xcodegen.

- Introduce project.yml for use with Device Farm test scripts
  - HTTP tests are now run in Device Farm in addition to the normal unit test suite
- Doc generation
  - Combine doc generation scripts
  - Remove cocoapods installation from generate-docs
  - Update jazzy invocation to build the SDK as a Swift package instead of using the Xcode project
- Clean up framework versioning build settings. We require that client apps are rebuilt when upgrading Maps SDK versions, so these settings are superfluous.
- Make update-version.sh update the Info.plist instead of using agvtool
- Reorganization & Deletions
  - Delete legacy Xcode project
  - Reorganize other contents of the Mapbox directory
  - Delete Mapbox directory
  - Updates the .gitattributes file
- Makefile
  - Update symbolication and coverage
    - Since each test run now has 2 profraw files, our coverage scripts now use the profdata file which is the result of merging the profraw files.
    - Updates script to log the symbolicated reports for easier analysis of CI failures
  - Update device farm rules
  - remove xcframework from makefile
- Update CI configs
- depsvalidator
  - Add support for user-defined manifest types
  - Update dependencies
  - Remove built-in JSON manifest support
  - Update config to use a user-defined manifest type for packager's versions.json


scripts/doc-generation/generate-docs.sh
```diff
@@ -37,10 +37,8 @@ function finish { >&2 echo -e "\033[1m\033[32m✔ $@\033[0m"; }
 
 SCRIPT_DIR=`dirname "$0"`
 
-# Search for Jazzy and install documentation dependencies if necessary.
-if [ -z `which jazzy` ]; then
-    $SCRIPT_DIR/install-documentation-dependencies.sh
-fi
+# Install documentation dependencies
+$SCRIPT_DIR/install-documentation-dependencies.sh
```
macdrevx, line 41: `install-documentation-dependencies.sh` contains appropriate checks for whether jazzy already exists, so I updated this to just invoke it unconditionally.

scripts/doc-generation/install-documentation-dependencies.sh
```diff
@@ -66,4 +67,18 @@ else
...
+    brew install xcodegen
+
+    if [ -z `which xcodegen` ]; then
+        error "Unable to install xcodegen ($XCODEGEN_VERSION)."
+        exit 1
+    fi
+else
+    step "Found xcodegen (${XCODEGEN_VERSION})"
+fi
+
```
macdrevx, line 83: Need to test this

scripts/release/update-version.sh
```diff
@@ -24,15 +24,6 @@ SHORT_VERSION=${SEM_VERSION%-*}
...
 
-cd Mapbox
-
-step "Updating Xcode targets to version ${SHORT_VERSION}…"
-
-xcrun agvtool bump -all
-xcrun agvtool new-marketing-version "${SHORT_VERSION}"
-
-cd ..
-
```
macdrevx, line 35: Need to discuss this with @neelmistry94

macdrevx:
Discuss the unusual framework embedding result with @julianrex

scripts/doc-generation/install-documentation-dependencies.sh
```diff
@@ -66,4 +67,18 @@ else
     step "Found Jazzy (${JAZZY_VERSION})"
 fi
 
+## Install xcodegen
```
neelmistry94, line 70: I almost feel like this entire script can be removed and we should just move it to generate-docs.sh

macdrevx: having it separate was actually nice for testing it when I made this change

neelmistry94: I can see that use case, but why would this install script need to be reused? When would we specifically need to install these dependencies but not run the generate docs

macdrevx: we don't have any other use cases for reusing the script at the moment. happy to combine it since it seems you feel strongly about it

neelmistry94: That would be great. I'd love to reduce unneeded overhead

.gitattributes
```diff
@@ -2,4 +2,4 @@
 *.pbxproj merge=union
 
 # Mark files as generated in Github PRs
-Mapbox/MapboxMapsStyle/Generated/** linguist-generated=true
\ No newline at end of file
+Sources/MapboxMaps/Style/Generated/** linguist-generated=true
```
macdrevx: Noticed this had gone out of sync a while back

scripts/doc-generation/generate-docs.sh
```diff
@@ -37,11 +37,65 @@ function finish { >&2 echo -e "\033[1m\033[32m✔ $@\033[0m"; }
...
-    $SCRIPT_DIR/install-documentation-dependencies.sh
+# Install documentation dependencies
+
+# VARIABLES
+COCOAPODS_VERSION="1.7.5"
+JAZZY_VERSION="0.13.6"
+CIRCLECI=${CIRCLECI:-false}
+
+## Install CocoaPods
+if [[ -z `which pod` || $(pod --version) != "${COCOAPODS_VERSION}" ]]; then
```
neelmistry94, line 48: why is cocoapods needed for docs generation?

macdrevx: hmm, i can't see a reason. I'll remove it

.circleci/config.yml
```diff
@@ -430,6 +432,7 @@ jobs:
       - run:
           name: Install Device Farm Dependencies
           command: make install-devicefarm-dependencies
+      - install-dependencies
```
macdrevx, line 435: Is this necessary?

macdrevx: Yes, because this causes xcodegen to be installed, and on line 450, `make test-with-device-farm` is invoked, which runs xcodegen unconditionally since it's shared between unit tests and examples tests.

julianrex (approved):
Thank you!

PR #190, opened 2021-03-17
Add Examples app entry for a data-driven, anchored annotation example

avi-c:
### Summary of changes

This PR adds a new top-level example to the Examples app. This example shows how to add a custom SymbolLayer that supports adding data-driven image annotations that support anchoring on either the left, center, or right side of the annotation image. Annotations support stretchable images that will dynamically resized based on their text contents.

Custom Symbols

Issue #381, opened 2021-05-21
Stretchable UIImage support seems to be broken


avi-c:
## Environment

- Maps SDK Version: v10 beta

## Observed behavior and steps to reproduce

The Nav SDK started seeing some issues with the route duration annotations being implemented here: https://github.com/mapbox/mapbox-navigation-ios/issues/2996. These issues are also seen in the simpler PR for adding a new example to the Maps SDK sample app: https://github.com/mapbox/mapbox-maps-ios/pull/190.

The issues are around the `icon-anchor` and `text-anchor` properties seeming to regress somewhere between when I first put up the sample PR and now. @nishant-karajgikar and I spent a lot of time debugging this. Our shared assumption was that the issue lay either in my use of the runtime styling expressions or in the SDK expression logic itself.

After looking at it for a while and even eliminating the use of data-driven styling, the main issues remained. We came to the conclusion that the issues are likely to be in how the SDK handles generating the GL stretchable images from the UIImage interface provided in the iOS Maps SDK.

https://github.com/mapbox/mapbox-maps-ios/pull/190 contains an animated GIF of how this should look, mainly that the annotations are clearly visually anchored properly with the tip of the annotation staying attached to the map at the same spot. If you build and try the same sample in this branch: `ac-AnnotationStylingRedux` you will see that the visual behavior is now incorrect and the anchors are not maintained properly.

avi-c:
Also, the `icon-offset` and `text-offset` properties seem to be broken. it could be that only one or the other is the issue or both could have an issue.

ZiZasaurus:
I wanted to provide an image depicting the behavior @avi-c mentioned above:

![navRouteImage](https://user-images.githubusercontent.com/44972592/119384237-62934d80-bc92-11eb-9488-cec2f286473d.png)

avi-c:
On the latest versions of the Maps SDK beta we are actually seeing different issues than that one, but both point to errors in the image stretching. Right now we see incorrect anchoring of the annotations. If you look at the image below you will see that the pin should be anchored on the route line but instead it is anchored further up the base/stem of the annotation. This used to work as expected in earlier betas.

![Simulator Screen Shot - iPhone 11 - 2021-05-24 at 10 55 25](https://user-images.githubusercontent.com/10037825/119388189-cad83400-bc7e-11eb-9ba4-7ec4a7760ba2.png)

avi-c:
[...]

PR #548, opened 2021-07-16
[Example] Data-Driven, Anchored Styling

cmilack:
## Pull request checklist:

### Summary of changes

This PR adds a new top-level example to the Examples app. This example shows how to add a custom SymbolLayer that supports adding data-driven image annotations that support anchoring on either the left, center, or right side of the annotation image. Annotations support stretchable images that will dynamically resize based on their text contents.

This was originally submitted another PR which has been closed: https://github.com/mapbox/mapbox-maps-ios/pull/190. All feedback from that PR has been addressed here.

This also addresses the issue described here: https://github.com/mapbox/mapbox-maps-ios/issues/381. This turned out to be expected behavior when using both icon-anchor and icon-text-fit.

### User impact (optional)

Custom Symbols


avi-c (dismissed):
I'm approving although it would be good to get another reviewer too since I wrote a bunch of the code I am approving. That said, most of the code was already approved in the earlier, abandoned PR so I think it should be approved by other reviewers without any major issues.

Sources/MapboxMaps/Annotations/Generated/PointAnnotation.swift
```diff
@@ -311,4 +311,4 @@ public struct PointAnnotation: Annotation {
 }
 
 // End of generated file.
-// swiftlint:enable all
\ No newline at end of file
+// swiftlint:enable all
```
macdrevx: probably shouldn't be touching generated files

Apps/Examples/Examples/All Examples/CustomAnchoredSymbolExample.swift
```diff
@@ -0,0 +1,321 @@
...
+    var coordinate: CLLocationCoordinate2D
+    var highlighted: Bool
+    var sortOrder: Int
+    var tailPosition: SymbolTailPosition
+    var label: String
+    var imageName: String
+}
+
+@objc(CustomAnchoredSymbolExample)
+public class CustomAnchoredSymbolExample: UIViewController, ExampleProtocol {
```
macdrevx, line 21: App code never needs to be public since it won't be consumed by other modules. Keeping it internal unlocks additional compile-time optimizations that are not possible for public constructs.

Apps/Examples/Examples/All Examples/CustomAnchoredSymbolExample.swift
```diff
@@ -0,0 +1,321 @@
...
+import MapboxMaps
+import Turf
+
+private enum SymbolTailPosition: Int {
+    case left
+    case right
+    case center
+}
+
+private struct DebugFeature {
```
macdrevx, line 11: Can we choose a different name here? What's Debug about this feature?

avi-c: Another name is fun. It was originally named debug since it was for debugging the visualization of the annotations and "Feature" was a reserved word. Maybe "AnchoredFeature" or similar is better now.

Apps/Examples/Examples/All Examples/CustomAnchoredSymbolExample.swift
```diff
@@ -0,0 +1,321 @@
...
+                                                                  zoom: 15.0))
+
+        mapView = MapView(frame: view.bounds, mapInitOptions: options)
+        mapView.autoresizingMask = [.flexibleWidth, .flexibleHeight]
+        view.addSubview(mapView)
+
+        let tapGestureRecognizer = UITapGestureRecognizer(target: self, action: #selector((mapSymbolTap(sender:))))
+        mapView.addGestureRecognizer(tapGestureRecognizer)
+
+        // Allows the delegate to receive information about map events.
```
macdrevx, line 53: Use of the word `delegate` in this comment is confusing given that this doesn't really fit the typical delegation pattern on iOS

Apps/Examples/Examples/All Examples/CustomAnchoredSymbolExample.swift
```diff
@@ -0,0 +1,321 @@
...
+            try? style.addImage(image.tint(highlightTintColor),
+                                id: "imageLeftHanded-Highlighted",
+                                sdf: false,
+                                stretchX: stretchX,
+                                stretchY: stretchY,
+                                content: imageContent)
+        }
+    }
+
+    private func addFeatures() -> FeatureCollection {
```
macdrevx, line 166: The naming of this method implies side effects, but there aren't any. How about `makeFeatures` instead, following the Swift API naming conventions?

Screen Shot 2021-07-20 at 11 36 01 AM

https://swift.org/documentation/api-design-guidelines/

Apps/Examples/Examples/All Examples/CustomAnchoredSymbolExample.swift
```diff
@@ -0,0 +1,321 @@
...
+
+            self.updateSymbolImages()
+            let features = self.addFeatures()
+            self.addSymbolLayer(features: features)
+
+            // The below line is used for internal testing purposes only.
+            self.finish()
+        }
+
+        // Add the label on top of the map view controller.
```
macdrevx, line 65:
```suggestion
        // Add the label on top of the map view.
```

Apps/Examples/Examples/All Examples/CustomAnchoredSymbolExample.swift
```diff
@@ -0,0 +1,321 @@
...
+        shapeLayer.textOffset = .expression(offsetExpression)
+
+        try? mapView.mapboxMap.style.addLayer(shapeLayer)
+    }
+}
+
+extension UIImage {
+    // Produce a copy of the image with tint color applied.
+    // Useful for deployment to iOS versions prior to 13 where tinting support was added to UIImage natively.
+    func tint(_ tintColor: UIColor) -> UIImage {
```
macdrevx, line 297:
```suggestion
    func tinted(with tintColor: UIColor) -> UIImage {
```

* `tinted` vs `tint`: The latter implies mutation. The former implies the creation of a new object.
* `_` vs `with`: without the argument label, at the point of use, it reads as if you're tinting the color rather than the image.

Apps/Examples/Examples/All Examples/CustomAnchoredSymbolExample.swift
```diff
@@ -0,0 +1,321 @@
...
+        }
+        shapeLayer.textOffset = .expression(offsetExpression)
+
+        try? mapView.mapboxMap.style.addLayer(shapeLayer)
+    }
+}
+
+extension UIImage {
+    // Produce a copy of the image with tint color applied.
+    // Useful for deployment to iOS versions prior to 13 where tinting support was added to UIImage natively.
```
macdrevx, line 296: Could we do some availability checks inside this method and use the built-in method when available? Then when we drop support for iOS 12 and lower someday, Xcode should lead us here with a warning and we can remove this extension entirely

Apps/Examples/Examples/All Examples/CustomAnchoredSymbolExample.swift
```diff
@@ -0,0 +1,321 @@
...
+                                       "imageName": feature.imageName,
+                                       "sortOrder": feature.highlighted == true ? index : -index]
+
+            features.append(pointFeature)
+        }
+
+        return FeatureCollection(features: features)
+    }
+
+    private func addSymbolLayer(features: FeatureCollection) {
```
macdrevx, line 196: nit: features implies [Feature] which is only 1 aspect of a [FeatureCollection](https://github.com/mapbox/turf-swift/blob/main/Sources/Turf/FeatureCollection.swift).

```suggestion
    private func addSymbolLayer(with featureCollection: FeatureCollection) {
```

Apps/Examples/Examples/All Examples/CustomAnchoredSymbolExample.swift
```diff
@@ -0,0 +1,321 @@
...
+        NSLayoutConstraint.activate([
+            label.leadingAnchor.constraint(equalTo: view.leadingAnchor),
+            label.trailingAnchor.constraint(equalTo: view.trailingAnchor),
+            label.topAnchor.constraint(equalTo: view.safeAreaLayoutGuide.topAnchor),
+            label.heightAnchor.constraint(equalToConstant: 60.0)
+        ])
+    }
+
+    @objc private func mapSymbolTap(sender: UITapGestureRecognizer) {
+        if sender.state == .recognized {
```
macdrevx, line 82: Consider converting to a guard

Apps/Examples/Examples/All Examples/CustomAnchoredSymbolExample.swift
```diff
@@ -0,0 +1,321 @@
...
+    var highlighted: Bool
+    var sortOrder: Int
+    var tailPosition: SymbolTailPosition
+    var label: String
+    var imageName: String
+}
+
+@objc(CustomAnchoredSymbolExample)
+public class CustomAnchoredSymbolExample: UIViewController, ExampleProtocol {
+    static let symbols = "custom_symbols"
```
macdrevx, line 22: Could we improve the naming of this constant? Is it an identifier?

Apps/Examples/Examples/All Examples/CustomAnchoredSymbolExample.swift
```diff
@@ -0,0 +1,321 @@
...
+            DebugFeature(coordinate: CLLocationCoordinate2DMake(40.707918, -74.006008), highlighted: false, sortOrder: 0, tailPosition: .right, label: "Cliff & John - Righthand Stem", imageName: "imageRightHanded"),
+            DebugFeature(coordinate: CLLocationCoordinate2DMake(40.716281, -74.004526), highlighted: true, sortOrder: 1, tailPosition: .right, label: "Broadway & Worth - Right Highlighted", imageName: "imageRightHanded-Highlighted"),
+            DebugFeature(coordinate: CLLocationCoordinate2DMake(40.710194, -74.004248), highlighted: true, sortOrder: 1, tailPosition: .left, label: "Spruce & Gold - Left Highlighted", imageName: "imageLeftHanded-Highlighted"),
+            DebugFeature(coordinate: CLLocationCoordinate2DMake(40.7128, -74.0060), highlighted: true, sortOrder: 2, tailPosition: .center, label: "City Hall - Centered Highlighted", imageName: "imageCentered-Highlighted"),
+            DebugFeature(coordinate: CLLocationCoordinate2DMake(40.711427, -74.008614), highlighted: false, sortOrder: 3, tailPosition: .center, label: "Broadway & Vesey - Centered Stem", imageName: "imageCentered")
+        ]
```
macdrevx, line 177: Calling the lower array featureList is confusing given that now we have Turf.Feature and DebugFeature. Maybe we could rename DebugFeature in terms of just some generic data entity and avoid using the term feature in its name

Apps/Examples/Examples/All Examples/CustomAnchoredSymbolExample.swift
```diff
@@ -0,0 +1,321 @@
...
+
+        for (index, feature) in featureList.enumerated() {
+            let point = Turf.Point(feature.coordinate)
```
macdrevx, line 180: With https://github.com/mapbox/mapbox-maps-ios/pull/548/files#r673294649, in mind, maybe this would become something like…

```suggestion
        for (index, dataItem) in dataItems.enumerated() {
            let point = Turf.Point(dataItem.coordinate)
```

Apps/Examples/Examples/All Examples/CustomAnchoredSymbolExample.swift
```diff
@@ -0,0 +1,321 @@
...
+            var pointFeature = Feature(geometry: .point(point))
+
+            // Set the feature attributes which will be used in styling the symbol style layer.
+            pointFeature.properties = ["highlighted": feature.highlighted,
+                                       "tailPosition": feature.tailPosition.rawValue,
+                                       "text": feature.label,
+                                       "imageName": feature.imageName,
+                                       "sortOrder": feature.highlighted == true ? index : -index]
+
+            features.append(pointFeature)
```
macdrevx, line 190: And then down here things get a lot more clear because we don't have to parse feature vs pointFeature

Apps/Examples/Examples/All Examples/CustomAnchoredSymbolExample.swift
macdrevx: It also seems unnecessary for this to be static. Letting it be an instance property would clean things up at the point of use.

Apps/Examples/Examples/All Examples/CustomAnchoredSymbolExample.swift
```diff
@@ -0,0 +1,321 @@
...
+
+    @objc private func mapSymbolTap(sender: UITapGestureRecognizer) {
+        if sender.state == .recognized {
+            let layers: [String] = [CustomAnchoredSymbolExample.symbols]
+
+            mapView.mapboxMap.queryRenderedFeatures(at: sender.location(in: mapView),
+                                                    options: RenderedQueryOptions(layerIds: layers, filter: nil)) { result in
+                switch result {
+                case .success(let features):
+                    if features.count > 0 {
```
macdrevx, line 89:
```suggestion
                    if !features.isEmpty {
```

See docs on isEmpty for the recommended best practice

Apps/Examples/Examples/All Examples/CustomAnchoredSymbolExample.swift
```diff
@@ -0,0 +1,321 @@
...
+                        guard  let featureText = features[0].feature.properties["text"] as? String else { return }
```
macdrevx, line 90:
```suggestion
                    if let featureText = features.first?.feature.properties["text"] as? String {
```

macdrevx (commented):
Done with comments. Will look again after updates are in. Also happy to discuss feedback if desired.

macdrevx (changes requested):
Meant to request changes

Apps/Examples/Examples/All Examples/CustomAnchoredSymbolExample.swift
macdrevx: See https://github.com/mapbox/mapbox-maps-ios/pull/548#discussion_r673294649. I think avoiding the term feature in this name would be advantageous.

macdrevx:
Closing this PR for now. We can revisit when we prioritize adding support for stretchable UIImages.

Issue #140, opened 2021-02-24
GestureManager does not notify delegate for tap or quickZoom


johnkassebaum:
## Environment

- Xcode version: 12.4
- iOS version: 14.4
- Devices affected: any
- Maps SDK Version: beta.13

## Observed behavior and steps to reproduce

- Assign an instance of a `GestureManagerDelegate` adopting type to a `MapView`'s `gestureManager.delegate` property
- note that `GestureManagerDelegate` method `gestureBegan(for:)` is not called for `GestureType` of `tap` or `quickZoom`, but that it is called for `pinch`, `pan`, `rotate` and `pitch`.

## Expected behavior

A tap gesture should invoke `gestureBegan(for:)`

## Notes / preliminary analysis

Tried assigning the `gestureManager.delegate` at different points in a `MapView`'s lifecycle to no avail.

johnkassebaum:
After further consideration on the `GestureManager`, is there possibly not a `UITapGestureRecognizer` for single taps (because a single tap does nothing) and that we add one via `gestureRecognizerShouldBegin(_:)`? Still, single touch double tap and double finger single tap, which do exist, would qualify as taps -- but how are they different from `quickZoom`?

jmkiley:
Noting that I do see `gestureBegan` being called for quick zoom, but not for single tap. The single tap gesture recognizer is managed by the `AnnotationManager`s and doesn't seem to be connected to the `GestureManagerDelegate`.

landyrev:
@jmkiley looks like it is still the issue for a global MapView GestureManagerDelegate

Following delegate not receiving tap gestures in RC4:
```swift
mapView.gestures.delegate = self
```

PR #549, opened 2021-07-19
[gestures]: GestureManager singleTap gesture handler

landyrev:
Fixes: https://github.com/mapbox/mapbox-maps-ios/issues/140

## Pull request checklist:
### Summary of changes

This PR adds missing sintleTap event handler to GestureManager

### DIFF
--- a/Sources/MapboxMaps/Gestures/GestureManager.swift
+++ b/Sources/MapboxMaps/Gestures/GestureManager.swift
@@ -24,6 +24,10 @@ public final class GestureManager: NSObject {
         return gestureHandlers[.pan]?.gestureRecognizer
     }
 
+    public var tapGestureRecognizer: UIGestureRecognizer? {
+        return gestureHandlers[.tap(numberOfTaps: 1, numberOfTouches: 1)]?.gestureRecognizer
+    }
+
     /// The underlying gesture recognizer for the "double tap to zoom in" gesture
     public var doubleTapToZoomInGestureRecognizer: UIGestureRecognizer? {
         return gestureHandlers[.tap(numberOfTaps: 2, numberOfTouches: 1)]?.gestureRecognizer
--- a/Sources/MapboxMaps/Gestures/GestureOptions.swift
+++ b/Sources/MapboxMaps/Gestures/GestureOptions.swift
@@ -117,6 +117,8 @@ public struct GestureOptions: Equatable {
     internal func supportedGestureTypes() -> Set<GestureType> {
         var supportedGestureTypes = Set<GestureType>()
 
+        supportedGestureTypes.insert(.tap(numberOfTaps: 1, numberOfTouches: 1))
+
         if scrollEnabled {
             supportedGestureTypes.insert(.pan)
         }


macdrevx:
Hi @landyrev, thanks for contributing to Mapbox. I took a look at issue #140 and retested the implementation. I'm seeing the GestureManager's delegate called as expected:

* panning with a single touch: pan
* double tap with a single touch: zoom in
* double tap with two touches: zoom out
* double tap and drag up or down: quick zoom
* two touches side-by-side that move up or down: pitch
* two touches that rotate about their midpoint: rotate
* two touches that move toward or away from their midpoint: pinch

The gesture manager's delegate isn't called for a single tap because there's no corresponding manipulation of the map camera. As @jmkiley pointed out our annotations components do use single tap gesture recognizers to implement annotation selection and deselection. These events get sent out through [their own delegate properties](https://github.com/mapbox/mapbox-maps-ios/blob/e3cc234b086963526cb2784f610e6cbc7268bd61/Sources/MapboxMaps/Annotations/Generated/PointAnnotationManager.swift#L494).

Could you tell us more about your use case for wanting the GestureManagerDelegate method to be called following a single tap with one touch? What are you looking to build on top of that?

landyrev:
Hi @macdrevx!

Actually, I was trying to implement a simple feature — dropping pins on a map at random locations with user touch ([like this](https://cln.sh/qHwyv4)). 

I've already managed to do this using a native mapView single tap gesture recognizer and subsequent point to geolocation conversion, but I think that iOS SDK needs singleTap gesture delegate similar to Android [OnMapClickListener](https://docs.mapbox.com/android/maps/api/10.0.0-rc.4/-mapbox%20-maps%20-android/com.mapbox.maps.plugin.gestures/-on-map-click-listener/) to implement those kind of feature.

landyrev:
Current PR actually not solves the problem too, because GestureManager delegates do not provide gesture view/map location

macdrevx:
Hi @landyrev thanks for the explanation.

> I've already managed to do this using a native mapView single tap gesture recognizer and subsequent point to geolocation conversion

This is our recommended approach right now. For anyone else following along, here's what that looks like:

    // Add a gesture recognizer, perhaps in viewDidLoad()
    mapView.addGestureRecognizer(UITapGestureRecognizer(target: self, action: #selector(handleTap(_:))))

    // Implement the handler
    @objc func handleTap(_ gestureRecognizer: UITapGestureRecognizer) {
        let point = gestureRecognizer.location(in: mapView)
        let coordinate = mapView.mapboxMap.coordinate(for: point)
        // do something with the coordinate
    }

As you can imagine, there are many different types of gestures that apps using the SDK might want to support, and since `MapView` has all the capabilities of `UIView`, including support for adding arbitrary gesture recognizers, we decided to provide the building blocks that let developers implement their own behaviors rather than making the SDK implement gestures that are unused internally.

Comparing this approach to Android's OnMapClickListener, I see what you're saying about that interface doing the conversion from a point in a view to a coordinate on a map for you. I'll bring this feedback to the team, and we'll discuss which direction to take.

Just imagining what this might look like if we did have some delegate where the converted coordinate was delivered directly, the example above might look like:

    // Assign a delegate
    mapView.gestures.delegate = self

    // Implement the delegate method
    func gestureManager(_ gestureManager: GestureManager, didDetectTappedCoordinate coordinate: CLLocationCoordinate2D) {
        // do something with the coordinate
    }

This saves a couple lines of code but would likely require more than a couple lines to implement in the SDK. Additionally, should the interface accept the point in the view as well, or would we require folks who want the point in the view instead of the coordinate to use a gesture recognizer or to the opposite conversion?

Thanks again for sharing your use case.

landyrev:
Thank you for detailed reply, @macdrevx 

Closing this PR for now

PR #579, opened 2021-08-05
Test using CocoaImageHashing for image comparisons

julianrex:

## Pull request checklist:

### Summary of changes

This PR uses [cocoaimagehashing](https://github.com/ameingast/cocoaimagehashing) for image comparison tests, where [pixelmatch-cpp](https://github.com/mapbox/pixelmatch-cpp) needed a unusably high tolerance factor.

More understanding of the meaning of the distance returned by `hashDistance` is needed before this PR can really be merged. The tests currently uses the maximum distance (2) returned by diffing the same rendered scheme on different iOS versions - the main difference were in the CG context drawing and text rendering; map rendering did not change. 

This change was tested by added some variation to the source reference images, and appeared to behave well.

### DIFF
--- a/Tests/MapboxMapsTests/Helpers/XCTestCase+MapboxAccessToken.swift
+++ b/Tests/MapboxMapsTests/Helpers/XCTestCase+MapboxAccessToken.swift
@@ -1,5 +1,6 @@
 import XCTest
 import MapboxMaps
+import CocoaImageHashing
 
 extension XCTestCase {
     func guardForMetalDevice() throws {
@@ -70,10 +71,18 @@ extension XCTestCase {
             return false
         }
 
-        // This comparison will NOT work for images embedded in an xcassets
-        // package. Images appear to be "optimized" modifying the RGB colors.
-        // In future, this should be replaced by a `pixelmatch` comparison.
-        equal = (expectedImage.pngData() == observedImage.pngData())
+        // See https://github.com/ameingast/cocoaimagehashing, http://phash.org
+        // and https://github.com/aetilius/pHash
+        let start = CACurrentMediaTime()
+        let imageHashing = OSImageHashing.sharedInstance()
+        let observedHash = imageHashing.hashImage(observedImage, with: .pHash)
+        let expectedHash = imageHashing.hashImage(expectedImage, with: .pHash)
+        let imageDistance = imageHashing.hashDistance(observedHash, to: expectedHash, with: .pHash)
+        let end = CACurrentMediaTime()
+
+        equal = (imageDistance <= 2)
+
+        print("Image comparison took \(end-start) seconds, distance = \(imageDistance)")
 
         return equal
     }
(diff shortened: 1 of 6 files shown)


julianrex:
Adding that we may want to evaluate https://github.com/google/butteraugli

Tests/MapboxMapsTests/Helpers/XCTestCase+MapboxAccessToken.swift
```diff
@@ -70,10 +71,18 @@ extension XCTestCase {
...
+        let start = CACurrentMediaTime()
+        let imageHashing = OSImageHashing.sharedInstance()
+        let observedHash = imageHashing.hashImage(observedImage, with: .pHash)
+        let expectedHash = imageHashing.hashImage(expectedImage, with: .pHash)
+        let imageDistance = imageHashing.hashDistance(observedHash, to: expectedHash, with: .pHash)
+        let end = CACurrentMediaTime()
+
+        equal = (imageDistance <= 2)
+
+        print("Image comparison took \(end-start) seconds, distance = \(imageDistance)")
```
macdrevx, line 85: How long do the comparisons take? Do we need to leave this logging in place?

macdrevx:
Could we move the extension method on XCTestCase to another file or rename the file that it's in to not be about access tokens?

project.yml
```diff
@@ -100,3 +100,6 @@ targets:
       - path: MapboxTestHost
     dependencies:
       - package: MapboxMaps
+        product: MapboxMaps
+      - package: MapboxMaps
+        product: CocoaImageHashing
```
macdrevx, line 105: Shouldn't this just be…

```suggestion
      - package: CocoaImageHashing
```

macdrevx:
If we do this, should we move all of our test assets into a catalog and reenable image optimization in our tests?

Alternatively, should we just have separate expected images for different iOS versions when needed? We could probably come up with a naming convention to make this easy.

Tests/MapboxMapsTests/Helpers/XCTestCase+MapboxAccessToken.swift
julianrex: No, they can be removed. Very fast for tests.

julianrex:
> should we move all of our test assets into a catalog

Agreed. If this lands, let's do that as a follow up.

> Alternatively, should we just have separate expected images for different iOS versions when needed? We could probably come up with a naming convention to make this easy.

Yes, @neelmistry94 and I discussed this - our concern was maintainability/scalability.

project.yml
julianrex: It doesn't like that. We get `Spec validation error: Target "MapboxTestHost" has an invalid package dependency "CocoaImageHashing"`

macdrevx: Ah, you also need to add an entry to the top-level packages key. See the xcodegen docs for more info and feel free to reach out to me if you want to walk through it.

PR #581, opened 2021-08-10
Add snapshot (last render) to MapView

julianrex:

## Pull request checklist:
 - [X] Include before/after visuals or gifs if this PR includes visual changes.

### Summary of changes

This PR migrates the `snapshot` code from MapViewIntegrationTestCase.swift to MapView, and makes it asynchronous. The API is marked as experimental.

There are currently a number of issues (some that we may wish to address later):

1. You need to disable Metal API Validation during debugging, otherwise the call to `CIContext().createCGImage(...)` will assert. If validation is enabled, no image will be returned.
2. On simulator sometimes the returned image can be nil; it's not clear why the image from the metal texture is blank. This may happen when the method is called after map rendering has finished. Scheduling a render call, and triggering the closure to happen after the render may solve this (and is likely something we'd want to look at anyway).
3. The texture is grabbed on a background thread, i.e. there could be timing issues if this API is used while the map is moving
4. The color of the returned image is washed out compared with the map view. This SO post may be useful when digging into colorspaces: https://stackoverflow.com/questions/45295689/mtkview-displaying-wide-gamut-p3-colorspace/49578887

    IMG_6655

5. There is currently no attribution.

### DIFF
--- a/Apps/DebugApp/DebugApp.xcodeproj/xcshareddata/xcschemes/DebugApp.xcscheme
+++ b/Apps/DebugApp/DebugApp.xcodeproj/xcshareddata/xcschemes/DebugApp.xcscheme
@@ -92,6 +92,7 @@
       ignoresPersistentStateOnLaunch = "NO"
       debugDocumentVersioning = "YES"
       debugServiceExtension = "internal"
+      enableGPUValidationMode = "1"
       allowLocationSimulation = "YES">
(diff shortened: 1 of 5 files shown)


nishant-karajgikar:
Closing in favor of https://github.com/mapbox/mapbox-maps-ios/pull/693

PR #642, opened 2021-09-01
Disambiguate Feature in public API

### LINKED ISSUE #634: Conflicting definitions of `Feature` type
## Environment

- Maps SDK Version: `ccd01605a571e6f39acea4e56fbe00057e28a103`

A recent change (#622) means that both `Turf` and `MapboxCommon` are exporting a type called `Feature` (at least when imported in Swift).

This means the following code is ambiguous:

```swift
import MapboxMaps

func foo(feature: Feature) {} // `Feature` is ambiguous for type lookup in this context
```

Of course, I can use `Turf.Feature` or `MapboxCommon.Feature` but this could get verbose.

## Expected behavior

Only one `Feature` type should be exported.

macdrevx:
Fixes #634

## Pull request checklist:
 - [x] Update the changelog

### Summary of changes

Adds a public typealias defining Feature as Turf.Feature. This fixes an issue where consuming code that referenced the type Feature was ambiguous due to the presence of both MapboxCommon.Feature and Turf.Feature.

### DIFF
--- a/Sources/MapboxMaps/Foundation/Extensions/Turf/Feature.swift
+++ b/Sources/MapboxMaps/Foundation/Extensions/Turf/Feature.swift
@@ -1,7 +1,7 @@
 import Foundation
 import MapboxCommon
 
-// MARK: - Feature
+public typealias Feature = Turf.Feature
 
 extension Turf.Feature {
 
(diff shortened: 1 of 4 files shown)


Sources/MapboxMaps/Foundation/Extensions/Turf/Feature.swift
```diff
@@ -1,7 +1,7 @@
 import Foundation
 import MapboxCommon
 
-// MARK: - Feature
+public typealias Feature = Turf.Feature
 
 extension Turf.Feature {
```
neelmistry94, line 6: since you define `Feature` above, can we make this `extension Feature`?

macdrevx: Yes, we could clean this up throughout the codebase for Feature and Geometry. We have a bunch of unnecessary `Turf.` say this point. Could you make a ticket to clean them all up at once?

neelmistry94 (approved):

PR #669, opened 2021-09-09
Internalize UIGestureRecognizerDelegate

neelmistry94:
## Pull request checklist:

### Summary of changes
This creates a new class that is a `NSObject` and conforms to `UIGestureRecognizerDelegate`. The `GestureManager` will hold on to this delegate as a single instance. This work was done so that `UIGestureRecognizerDelegate` can be internal instead of public.

### DIFF
--- a/Sources/MapboxMaps/Gestures/GestureManager.swift
+++ b/Sources/MapboxMaps/Gestures/GestureManager.swift
@@ -65,13 +65,12 @@ public final class GestureManager {
     /// Set this delegate to be called back if a gesture begins
     public weak var delegate: GestureManagerDelegate?
 
-    internal let gestureRecognizerDelegate: GestureRecognizerDelegate
+    internal weak var gestureRecognizerDelegate: GestureRecognizerDelegate?
 
     internal init(view: UIView, cameraAnimationsManager: CameraAnimationsManagerProtocol, mapboxMap: MapboxMap) {
         self.view = view
         self.cameraAnimationsManager = cameraAnimationsManager
         self.mapboxMap = mapboxMap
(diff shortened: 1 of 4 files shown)


Sources/MapboxMaps/Gestures/GestureManager.swift
```diff
@@ -65,11 +65,13 @@ public final class GestureManager: NSObject {
     /// Set this delegate to be called back if a gesture begins
     public weak var delegate: GestureManagerDelegate?
 
+    /// Internal delegate for gesture recognizers
+    internal weak var gestureRecognizerDelegate: GestureRecognizerDelegate?
```
macdrevx, line 69: Can this be a `let` reference to a non-optional delegate instead of a `weak var`?

Sources/MapboxMaps/Gestures/GestureRecognizerDelegate.swift
```diff
@@ -0,0 +1,41 @@
+import UIKit
+
+internal class GestureRecognizerDelegate: NSObject, UIGestureRecognizerDelegate {
+
+    /// The view that all gestures operate on
+    private weak var view: UIView?
```
macdrevx, line 6: Let's remove this weak ref to view and just grab the view from the gesture recognizer in `gestureRecognizerShouldBegin`

Sources/MapboxMaps/Gestures/GestureManager.swift
neelmistry94: I pushed a change to address this but it means silencing a swift lint warning

macdrevx (approved):

PR #677, opened 2021-09-14
Refactor gestures

macdrevx:
## Pull request checklist:
 - [x] Update the changelog

### Summary of changes

Refactors gestures for better modularity, maintainability, and testability.

#### Breaking Changes

- Pan deceleration has been temporarily removed
- `TapGestureHandler.init` was previously public by mistake and is now internal
- The behavior of `GestureManager.options` has been updated to better reflect the `isEnabled` state of the associated gesture recognizers
- The gesture recognizer properties of `GestureManager` are no longer `Optional`
- `GestureType` has been redesigned so that its cases have a 1-1 relationship with the built-in gestures

#### Public API Additions

- `CameraState`'s fields are now `var`s instead of `let`s for testing purposes, and a public, memberwise initializer has been added.
- `PanScrollingMode` now conforms to `CaseIterable`
- `GestureType` now conforms to `CaseIterable`

#### Bug fixes

- GestureManager no longer sets itself as the delegate of all gestures in MapView when its options change

#### Internal Refactoring

- Generalizes `CameraAnimatorMapboxMap` by renaming it to `MapboxMapProtocol` so that it can be used throughout the SDK. SDK seems small enough that per-component dependency inversion feels unnecessary and might negatively impact binary size.
- Removes old `GestureHandlerDelegate` by injecting `MapboxMap` and `CameraAnimationsManager` into each gesture handler to allow each handler to manipulate the camera directly.
- Removes unnecessary `GestureContextProvider`
- Refactors handlers and `GestureManager` for dependency injection and more thorough tests
- Tidies up handlers and `GestureManager` and associated tests for greater consistency


Sources/MapboxMaps/Foundation/Camera/CameraAnimationsManager.swift
```diff
@@ -15,8 +15,17 @@ internal protocol CameraAnimatorInterface: CameraAnimator {
     func update()
 }
 
+internal protocol CameraAnimationsManagerProtocol: AnyObject {
+    func ease(to camera: CameraOptions,
```
neelmistry94, line 19: what is the advantage of this being protocoled? Can we just have the methods in the `CameraAnimationsManager`

Sources/MapboxMaps/Gestures/GestureHandlers/DoubleTapGestureHandler.swift
```diff
@@ -0,0 +1,55 @@
...
+                return
+            }
+            cameraAnimationsManager.cancelAnimations()
+            delegate?.gestureBegan(for: gestureType)
+            _ = cameraAnimationsManager.ease(to: CameraOptions(zoom: mapboxMap.cameraState.zoom + zoomDelta),
+                                             duration: 0.3,
+                                             curve: .easeOut,
+                                             completion: nil)
+        default:
+            break
```
neelmistry94, line 52: The idea of a switch inside of a switch looks a little gross. Maybe the code instead the `.ended` case can be moved to a separate private function, or internal if it is something that can be unit tested.

Also, switch is the convention when dealing with these states, but since you are only using one case, would this be a better case for an `if gestureRecognizer.state == .ended` ?

Sources/MapboxMaps/Gestures/GestureHandlers/PinchGestureHandler.swift
```diff
@@ -1,63 +1,70 @@
...
-                                  initialAnchor: self.initialPinchCenterPoint,
-                                  initialCameraState: self.initialCameraState)
+            mapboxMap.setCamera(to: cameraOptions)
 
-            previousScale = pinchGestureRecognizer.scale
-        } else if pinchGestureRecognizer.state == .ended
-            || pinchGestureRecognizer.state == .cancelled {
+            mapboxMap.dragStart(for: initialPinchCenterPoint)
+            let dragOptions = mapboxMap.dragCameraOptions(
+                from: initialPinchCenterPoint,
```
neelmistry94, line 54: nit: indentation feels weird here

Sources/MapboxMaps/Gestures/GestureHandlers/PinchGestureHandler.swift
```diff
@@ -1,63 +1,70 @@
...
+                to: pinchCenterPoint)
+            mapboxMap.setCamera(to: dragOptions)
+            mapboxMap.dragEnd()
 
-            delegate.pinchEnded()
+            mapboxMap.setCamera(
+                to: CameraOptions(
```
neelmistry94, line 60: nit: indentation feels weird here. 
nit: make camera options on a different line

Sources/MapboxMaps/Gestures/GestureHandlers/PinchGestureHandler.swift
```diff
@@ -1,63 +1,70 @@
 import UIKit
 
-/// The PinchGestureHandler is responsible for all `pinch` related infrastructure
-/// Tells the view to update itself when required
-internal class PinchGestureHandler: GestureHandler {
-    private var previousScale: CGFloat = 0.0
+/// `PinchGestureHandler` updates the map camera in response to a 2-touch
+/// gesture that may consist of translation and scaling
+internal final class PinchGestureHandler: GestureHandler {
```
neelmistry94, line 5: Why is this gesture a final class?

Sources/MapboxMaps/Gestures/GestureHandlers/PitchGestureHandler.swift
```diff
@@ -1,88 +1,83 @@
...
     }
 
-    @objc internal func handlePitchGesture(_ gesture: UIPanGestureRecognizer) {
-
-        let horizontalTiltTolerance = delegate.horizontalPitchTiltTolerance()
-
-        if gesture.numberOfTouches != 2 {
-            gesture.state = .ended
-            return
+    private func touchAngleIsLessThanMaximum(for gestureRecognizer: UIGestureRecognizer) -> Bool {
```
neelmistry94, line 21: should this be called isTouchAngleIsLessThanMaximum?

Sources/MapboxMaps/Foundation/Camera/CameraAnimationsManager.swift
macdrevx: The gesture handlers depend on this protocol instead of depending on `CameraAnimationsManager` directly so that we can inject mocks when we test them.

Sources/MapboxMaps/Gestures/GestureHandlers/PinchGestureHandler.swift
macdrevx: Because it is not designed to be subclassed.

Sources/MapboxMaps/Gestures/GestureHandlers/PitchGestureHandler.swift
macdrevx: I don't think so. Typically you look to the place where these are called, so for example, on UIView, there's a property `isHidden`. Grammatically, `is` is the verb and the subject is the object you call the method on (the view) so that it forms a sentence: `view.isHidden`. In the situation you highlighted, we call `touchAngleIsLessThanMaximum` on self, which is a gesture handler. It doesn't really make sense grammatically for there to be a leading `is` on the method name since the function isn't saying anything about the gesture handler itself.

Sources/MapboxMaps/Foundation/Camera/CameraAnimationsManager.swift
neelmistry94: That makes sense, so it sounds like it is solely for testing purposes as opposed to SDK functionality/architecture?

Sources/MapboxMaps/Gestures/GestureHandlers/PinchGestureHandler.swift
macdrevx: I think this indentation is advantageous for a few reasons…

* It keeps the lines from getting too long because Xcode's indention behaviors only increase by 4 spaces per level when you indent this way.
* It helps to avoid needing to reindent subsequent lines if, for example, you rename `setCamera` or `CameraOptions` down the road. (unlikely in this situation, but as a general rule, this is useful)
* It provides a natural visual hierarchy of nested invocations and their associated parameters.

Sources/MapboxMaps/Gestures/GestureHandlers/PinchGestureHandler.swift
neelmistry94: why is this the only gesture designed that way?

Sources/MapboxMaps/Gestures/GestureHandlers/PinchGestureHandler.swift
macdrevx: See https://github.com/mapbox/mapbox-maps-ios/pull/677#discussion_r709487225

Sources/MapboxMaps/Gestures/GestureHandlers/DoubleTapGestureHandler.swift
macdrevx: It doesn't really bother me much, but I see your point. I'm happy to create a helper method here.

macdrevx: Re: `if` vs `switch`, they both work. Let's keep it consistent across all the handlers.

Sources/MapboxMaps/Gestures/GestureHandlers/PinchGestureHandler.swift
neelmistry94: Is this a style that we are choosing to follow in our codebase? I see the arguments for it, but IMO, I think this style of indentation is harder to read and follow. It doesn't feel swifty. 

I think we should standardize these things in our coding style guide

Sources/MapboxMaps/Gestures/GestureHandlers/DoubleTapGestureHandler.swift
neelmistry94: I'm good with keeping a `switch` I agree with the consistency. But yeah a helper method would be nice

Sources/MapboxMaps/Gestures/GestureHandlers/PinchGestureHandler.swift
macdrevx: > I think we should standardize these things in our coding style guide

Agreed. We don't have a standard here, so right now it's anyone's opinion.

Sources/MapboxMaps/Foundation/Camera/CameraAnimationsManager.swift
macdrevx: Yep, it's just an implementation detail for testing.

Sources/MapboxMaps/Gestures/GestureHandlers/PinchGestureHandler.swift
macdrevx: Good catch, they should all be final except for the base class.

Sources/MapboxMaps/Gestures/GestureHandlers/DoubleTapGestureHandler.swift
macdrevx: Updated to inject zoom delta and infer gesture type

Sources/MapboxMaps/Gestures/GestureHandlers/PinchGestureHandler.swift
macdrevx: Updated

Sources/MapboxMaps/Gestures/GestureHandlers/PinchGestureHandler.swift
macdrevx: Resolving. Let's centralize the conversation in the linked thread.

nishant-karajgikar (approved):
🎉

PR #680, opened 2021-09-15
Use @_spi(Experimental) for experimental methods

jmkiley:

## Pull request checklist:

### Summary of changes

This PR:
- Uses the `@_spi(Experimental)` annotation to mark style methods as experimental. This means that in order to use these methods, you will need to annotate your import statement for `MapboxMaps` with `@_spi(Experimental)`. (`@_spi(Experimental) import MapboxMaps`).
- Removes the underscore prefix that previously marked these methods as experimental.

### DIFF
--- a/Sources/MapboxMaps/Style/Style.swift
+++ b/Sources/MapboxMaps/Style/Style.swift
@@ -42,6 +42,7 @@ public class Style {
         let layerJSON = try layer.jsonObject()
         try _addPersistentLayer(with: layerJSON, layerPosition: layerPosition)
     }
+
     /**
      :nodoc:
      Moves a `layer` to a new layer position in the style.
@@ -50,6 +51,7 @@ public class Style {
 
      - Throws: `StyleError` on failure, or `NSError` with a _domain of "com.mapbox.bindgen"
      */
+    @_spi(experimental)
     public func _moveLayer(withId id: String, to position: LayerPosition) throws {
         let properties = try layerProperties(for: id)
         let isPersistent = try _isPersistentLayer(id: id)
@@ -88,6 +90,7 @@ public class Style {
      - Returns: The fully formed `layer` object of type equal to `type`
      - Throws: StyleError or type conversion errors
      */
+    @_spi(experimental)
     public func _layer(withId id: String, type: Layer.Type) throws -> Layer {
         // Get the layer properties from the map
         let properties = try layerProperties(for: id)
@@ -166,6 +169,7 @@ public class Style {
      - Returns: The fully formed `source` object of type equal to `type`.
      - Throws: StyleError or type conversion errors
      */
+    @_spi(experimental)
     public func _source(withId id: String, type: Source.Type) throws  -> Source {
         // Get the source properties for a given identifier
         let sourceProps = try sourceProperties(for: id)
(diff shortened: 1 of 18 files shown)


Sources/MapboxMaps/Style/Style.swift
```diff
@@ -165,25 +170,16 @@ public class Style {
      - Parameter type: The type of the source
      - Returns: The fully formed `source` object of type equal to `type`.
      - Throws: StyleError or type conversion errors
+     - Note: This method is marked as experimental. Annotate the import statement
+     for `MapboxMaps` with `@_spi(Experimental)` in order to use experimental methods.
      */
-    public func _source(withId id: String, type: Source.Type) throws  -> Source {
+    @_spi(Experimental) public func source(withId id: String, type: Source.Type) throws  -> Source {
```
macdrevx, line 176: I don't think this needs to be experimental anymore. (same for the corresponding layer method above). We could actually knock out another issue related to these APIs at the same time by removing the variants that don't take a type parameter.

macdrevx: That said, let's defer this feedback to our audit of all experimental APIs.

Sources/MapboxMaps/Style/Style.swift
```diff
@@ -165,25 +170,16 @@ public class Style {
...
-    ///
-    /// - Parameters:
-    ///   - sourceId: Style source identifier.
-    ///   - property: Style source property name.
-    ///
-    /// - Returns: The value of the property in the source with sourceId.
-    public func sourceProperty(for sourceId: String, property: String) -> Any {
-        return _sourceProperty(for: sourceId, property: property).value
-    }
-
```
macdrevx, line 186: Curious why this was removed

Sources/MapboxMaps/Style/StyleManagerProtocol.swift
```diff
@@ -386,7 +400,10 @@ internal protocol StyleManagerProtocol {
     /// - Parameter property: Style light property name.
     ///
     /// - Returns: Style light property value.
-    func _lightProperty(_ property: String) -> StylePropertyValue
+    ///
+    /// - Note: This method is marked as experimental. Annotate the import statement
+    /// for `MapboxMaps` with `@_spi(Experimental)` in order to use experimental methods.
+    func lightProperty(_ property: String) -> StylePropertyValue
```
macdrevx, line 406: These notes feel out of place in this context.

Sources/MapboxMaps/Style/Style.swift
jmkiley: This was removed because it seems like a [duplicate method](https://github.com/mapbox/mapbox-maps-ios/pull/680/files#diff-3c7e02fb492d91d8b149fb3a5799e4cca459c77771dba75e942fa715bb62d6c8R434-R436) . The difference seems to be that once returns a `StylePropertyValue` while the other returns the `StylePropertyValue.value`. Alternatively, we could rename the deleted method.

macdrevx: I'm open to keeping this change. Since the deleted method was not experimental and the one we kept is, maybe we should make the one we kept non-experimental?

macdrevx (approved):

Sources/MapboxMaps/Style/Style.swift
macdrevx: I think this is another issue we can address when we do the full audit of what is experimental

Issue #293, opened 2021-04-23
Use CADisplayLink.targetTimestamp to drive animations


nishant-karajgikar:
The new flyTo implementation uses `Date()` directly. Which is not the most sophisticated way of handling timing in animations. We should use [`CADisplayLink.targetTimestamp`](https://developer.apple.com/documentation/quartzcore/cadisplaylink/1648422-targettimestamp) instead.

Also applies to `GestureDecelerationCameraAnimator`. See https://github.com/mapbox/mapbox-maps-ios/pull/692#discussion_r712514230

Also applies to `InterpolatedLocationProducer`

_Originally posted by @julianrex in https://github.com/mapbox/mapbox-maps-ios/pull/282#discussion_r618626797_

macdrevx:
https://developer.apple.com/documentation/quartzcore/optimizing_promotion_refresh_rates_for_iphone_13_pro_and_ipad_pro has some good explanation of how to use targetTimestamp

macdrevx:
https://developer.apple.com/videos/play/wwdc2021/10147/ is also helpful.

One key takeaway is that switching to targetTimestamp isn't enough — we likely also need to use something like [MTLCommandBuffer.present(_:atTime:)](https://developer.apple.com/documentation/metal/mtlcommandbuffer/1442989-present) so that our content is drawn at the expected time when our display link is using a lower frame rate than the screen's actual refresh rate.

macdrevx:
Possibly relevant: https://developer.apple.com/documentation/metal/drawable_objects/creating_a_custom_metal_view

OdNairy:
Still actual https://github.com/mapbox/mapbox-maps-ios/blob/fb97a45e3b4c98b220dd89acf7b29a9876f9ab56/Sources/MapboxMaps/Camera/FlyToCameraAnimator.swift#L168

PR #692, opened 2021-09-20
Pan Deceleration

macdrevx:

## Pull request checklist:
 - [x] Update the changelog

### Summary of changes

Reimplements pan deceleration via a custom deceleration simulation.

### DIFF
--- a/Sources/MapboxMaps/Foundation/Camera/CameraAnimationsManager.swift
+++ b/Sources/MapboxMaps/Foundation/Camera/CameraAnimationsManager.swift
@@ -21,6 +21,12 @@ internal protocol CameraAnimationsManagerProtocol: AnyObject {
               curve: UIView.AnimationCurve,
               completion: AnimationCompletion?) -> Cancelable?
 
+    func decelerate(location: CGPoint,
+                    velocity: CGPoint,
+                    decelerationRate: CGFloat,
+                    locationChangeHandler: @escaping (CGPoint) -> Void,
+                    completion: @escaping () -> Void)
+
     func cancelAnimations()
 }
 
@@ -109,9 +115,7 @@ public class CameraAnimationsManager: CameraAnimationsManagerProtocol {
         cameraAnimatorsSet.add(flyToAnimator)
 
         flyToAnimator.addCompletion { [weak self, weak flyToAnimator] (position) in
-            if let internalAnimator = self?.internalAnimator,
-               let animator = flyToAnimator,
-               internalAnimator === animator {
+            if self?.internalAnimator === flyToAnimator {
                 self?.internalAnimator = nil
             }
             // Call the developer-provided completion (if present)
@@ -148,9 +152,7 @@ public class CameraAnimationsManager: CameraAnimationsManagerProtocol {
 
         // Nil out the `internalAnimator` once the "ease to" finishes
         animator.addCompletion { [weak self, weak animator] (position) in
-            if let internalAnimator = self?.internalAnimator,
-               let animator = animator,
-               internalAnimator === animator {
+            if self?.internalAnimator === animator {
                 self?.internalAnimator = nil
             }
             completion?(position)
@@ -280,5 +282,40 @@ public class CameraAnimationsManager: CameraAnimationsManagerProtocol {
         let cameraView = CameraView()
         cameraViewContainerView.addSubview(cameraView)
         return cameraView
+    }
+
+    /// This function will handle the natural decelration of a gesture when there is a velocity provided. A use case for this is the pan gesture.
+    /// - Parameters:
+    ///   - location: Current location of center coordinate
+    ///   - velocity: The speed at which the map should move over time
+    ///   - decelerationRate: A multiplication factor that determines the speed at which the velocity should slow down
+    ///   - locationChangeHandler: Change handler to be passed through to the animator
+    ///   - completion: Completion to be called after animation has finished
+    internal func decelerate(location: CGPoint,
+                             velocity: CGPoint,
+                             decelerationRate: CGFloat,
+                             locationChangeHandler: @escaping (CGPoint) -> Void,
+                             completion: @escaping () -> Void) {
+
+        // Stop the `internalAnimator` before beginning a deceleration
+        internalAnimator?.stopAnimation()
+
+        let decelerateAnimator = GestureDecelerationCameraAnimator(
+            location: location,
+            velocity: velocity,
+            decelerationRate: decelerationRate,
+            locationChangeHandler: locationChangeHandler,
+            dateProvider: DefaultDateProvider())
+
+        decelerateAnimator.completion = { [weak self, weak decelerateAnimator] in
+           if self?.internalAnimator === decelerateAnimator {
+               self?.internalAnimator = nil
+           }
+           completion()
+        }
+
+        cameraAnimatorsSet.add(decelerateAnimator)
+        decelerateAnimator.startAnimation()
+        internalAnimator = decelerateAnimator
     }
 }
--- a/Sources/MapboxMaps/Foundation/Camera/GestureDecelerationCameraAnimator.swift
+++ b/Sources/MapboxMaps/Foundation/Camera/GestureDecelerationCameraAnimator.swift
@@ -1,0 +1,68 @@
+import UIKit
+
+internal final class GestureDecelerationCameraAnimator: NSObject, CameraAnimatorInterface {
+
+    private var location: CGPoint
+    private var velocity: CGPoint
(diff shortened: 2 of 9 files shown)


Sources/MapboxMaps/Foundation/Camera/CameraAnimationsManager.swift
```diff
@@ -281,4 +283,39 @@ public class CameraAnimationsManager: CameraAnimationsManagerProtocol {
...
+
+    /// This function will handle the natural decelration of a gesture when there is a velocity provided. A use case for this is the pan gesture.
+    /// - Parameters:
+    ///   - location: Current location of center coordinate
+    ///   - velocity: The speed at which the map should move over time
+    ///   - decelerationRate: A multiplication factor that determines the speed at which the velocity should slow down
+    ///   - locationChangeHandler: Change handler to be passed through to the animator
+    ///   - completion: Completion to be called after animation has finished
+    internal func decelerate(location: CGPoint,
+                             velocity: CGPoint,
```
nishant-karajgikar, line 295: Would `CGVector` be a better data type to represent velocity?

Sources/MapboxMaps/Foundation/Camera/GestureDecelerationCameraAnimator.swift
```diff
@@ -0,0 +1,68 @@
...
+    private let locationChangeHandler: (CGPoint) -> Void
+    private var previousDate: Date?
+    private let dateProvider: DateProvider
+    internal var completion: (() -> Void)?
+
+    internal init(location: CGPoint,
+                  velocity: CGPoint,
+                  decelerationRate: CGFloat,
+                  locationChangeHandler: @escaping (CGPoint) -> Void,
+                  dateProvider: DateProvider) {
```
nishant-karajgikar, line 17: We can provide `DefaultDateProvider` as a default argument to `dateProvider` here. It'll clean up the call site.

Sources/MapboxMaps/Foundation/Camera/GestureDecelerationCameraAnimator.swift
```diff
@@ -0,0 +1,68 @@
...
+        completion = nil
+    }
+
+    internal func update() {
+        guard state == .active, let previousDate = previousDate else {
+            return
+        }
+
+        let currentDate = dateProvider.now
+        self.previousDate = currentDate
```
nishant-karajgikar, line 48: Might want to investigate moving to `CACurrentMediaTime` here too as part of https://github.com/mapbox/mapbox-maps-ios/issues/293

Sources/MapboxMaps/Foundation/Camera/GestureDecelerationCameraAnimator.swift
nishant-karajgikar: Oh wait -- I see that the constructor for `PanGestureHandler` also takes a `dateProvider`. Is that necessary?

Sources/MapboxMaps/Gestures/GestureHandlers/PanGestureHandler.swift
```diff
@@ -33,38 +39,63 @@ internal final class PanGestureHandler: GestureHandler {
...
-            case .vertical:
-                clampedTouchLocation = CGPoint(x: initialTouchLocation.x, y: touchLocation.y)
-            case .horizontalAndVertical:
-                clampedTouchLocation = touchLocation
+            handleChange(withTouchLocation: touchLocation)
+        case .ended:
+            // decelerate
+            guard let lastChangedDate = lastChangedDate,
+                  // if it's been more than 2 frames at 60 Hz since the last change, don't drift
+                  dateProvider.now.timeIntervalSince(lastChangedDate) < 2.0 / 60.0,
```
nishant-karajgikar, line 49: It's a bit confusing to me to describe this logic in terms of "frames at frequency" (especially since 60Hz doesn't seem to have any real significance). Would it be clearer to just store `2.0 / 60.0` as some timeout constant that is consumed as part of this conditional?

Sources/MapboxMaps/Gestures/GestureHandlers/PanGestureHandler.swift
```diff
@@ -33,38 +39,63 @@ internal final class PanGestureHandler: GestureHandler {
...
+            cameraAnimationsManager.decelerate(
+                location: touchLocation,
+                velocity: gestureRecognizer.velocity(in: view),
+                decelerationRate: decelerationRate,
+                locationChangeHandler: handleChange(withTouchLocation:),
+                completion: {
+                    self.initialTouchLocation = nil
+                    self.initialCameraState = nil
+                    self.lastChangedDate = nil
+                })
```
nishant-karajgikar, line 62: Does setting all these values to `nil` need to be done? I see that we assign these values when the pan gesture starts.

Sources/MapboxMaps/Foundation/Camera/CameraAnimationsManager.swift
macdrevx: Perhaps, but this is the type that UIGestureRecognizer uses, so it avoids any kind of conversion.

Sources/MapboxMaps/Foundation/Camera/GestureDecelerationCameraAnimator.swift
macdrevx: Let's discuss this. I'd like to move away from dependency injection via default args, so while this does make the call site a bit messier, I see it as an intermediate state on the way towards a more clear composition root.

Sources/MapboxMaps/Foundation/Camera/GestureDecelerationCameraAnimator.swift
macdrevx: Yep. Okay if we add this to that ticket and investigate it when we get to that?

Sources/MapboxMaps/Gestures/GestureHandlers/PanGestureHandler.swift
macdrevx: I'm fine with that.

Sources/MapboxMaps/Gestures/GestureHandlers/PanGestureHandler.swift
macdrevx: Probably not strictly necessary. It's a bit of defensive programming b/c it ensures that if you enter .changed without having entered .began, we'll just ignore it. Will that ever actually happen? Probably not.

That said, we'll ultimately need to capture self in this completion block to tell the delegate when the deceleration is complete, and there's no danger in letting that be a strong ref if you work out the dependency graph.

Sources/MapboxMaps/Gestures/GestureHandlers/PanGestureHandler.swift
```diff
@@ -8,10 +8,16 @@ internal final class PanGestureHandler: GestureHandler {
     // The camera state when the gesture began
     private var initialCameraState: CameraState?
 
+    private var lastChangedDate: Date?
+
+    private let dateProvider: DateProvider
+
```
neelmistry94, line 14: We should add some comments here these properties

Sources/MapboxMaps/Foundation/Camera/CameraAnimationsManager.swift
nishant-karajgikar: ooh that's strange.

Sources/MapboxMaps/Foundation/Camera/GestureDecelerationCameraAnimator.swift
nishant-karajgikar: Mhm okay!

Sources/MapboxMaps/Foundation/Camera/GestureDecelerationCameraAnimator.swift
nishant-karajgikar: Yep.

Sources/MapboxMaps/Gestures/GestureHandlers/PanGestureHandler.swift
nishant-karajgikar: Gotcha, thanks for the info.

Sources/MapboxMaps/Gestures/GestureHandlers/PanGestureHandler.swift
macdrevx: Updated

neelmistry94 (approved):

PR #694, opened 2021-09-21
Improve usability of generic methods

neelmistry94:
## Pull request checklist:

### Summary of changes
This updates style generic functions to make them more usable. It updates 3 functions to take a `type` parameter so that it can easily allow the developer to use generics.

### DIFF
--- a/Apps/Examples/Examples/All Examples/ShowHideLayerExample.swift
+++ b/Apps/Examples/Examples/All Examples/ShowHideLayerExample.swift
@@ -95,7 +95,7 @@ class ShowHideLayerExample: UIViewController, ExampleProtocol {
         // Update the museum layer's visibility based on whether the switch
         // is on. `visibility` is `nil` by default.
         do {
-            try style.updateLayer(withId: museumLayerId) { (layer: inout CircleLayer) in
+            try style.updateLayer(withId: museumLayerId, type: CircleLayer.self) { layer in
                 layer.visibility = .constant(sender.isOn ? .visible : .none)
             }
         } catch {
@@ -108,7 +108,7 @@ class ShowHideLayerExample: UIViewController, ExampleProtocol {
         // Update the contour layer's visibility based on whether the switch
         // is on.
         do {
-            try style.updateLayer(withId: contourLayerId) { (layer: inout CircleLayer) in
+            try style.updateLayer(withId: contourLayerId, type: CircleLayer.self) { layer in
                 layer.visibility = .constant(sender.isOn ? .visible : .none)
             }
         } catch {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,9 +7,11 @@ Mapbox welcomes participation and contributions from everyone.
 ### Breaking changes ⚠️
 
 * Removes default parameter values in the `addImage` function. ([#695](https://github.com/mapbox/mapbox-maps-ios/pull/695))
-* `public func layer<T: Layer>(withId id: String) throws -> T` has been updated to `public func layer<T>(withId id: String, type: T.Type) throws -> T where T: Layer`
-* `public func updateLayer<T: Layer>(withId id: String, update: (inout T) throws -> Void) throws` has been updated to `public func updateLayer<T>(withId id: String, type: T.Type, update: (inout T) throws -> Void) throws where T: Layer`
-* `public func source<T: Source>(withId id: String) throws -> T` has been updated to `public func source<T>(withId id: String, type: T.Type) throws -> T where T: Source`
+* `public func layer<T: Layer>(withId id: String) throws -> T` has been updated to `public func layer<T>(withId id: String, type: T.Type) throws -> T where T: Layer`. ([#694](https://github.com/mapbox/mapbox-maps-ios/pull/694))
+* `public func updateLayer<T: Layer>(withId id: String, update: (inout T) throws -> Void) throws` has been updated to `public func updateLayer<T>(withId id: String, type: T.Type, update: (inout T) throws -> Void) throws where T: Layer`. ([#694](https://github.com/mapbox/mapbox-maps-ios/pull/694))
+* `public func source<T: Source>(withId id: String) throws -> T` has been updated to `public func source<T>(withId id: String, type: T.Type) throws -> T where T: Source`. ([#694](https://github.com/mapbox/mapbox-maps-ios/pull/694))
+* `@_spi(Experimental) public func layer(withId id: String, type: Layer.Type) throws -> Layer` is no longer experimental and has been updated to `public func layer(withId id: String) throws -> Layer`. ([#694](https://github.com/mapbox/mapbox-maps-ios/pull/694))
+* `@_spi(Experimental) public func source(withId id: String, type: Source.Type) throws  -> Source` is no longer experimental and has been updated to `public func source(withId id: String) throws  -> Source`. ([#694](https://github.com/mapbox/mapbox-maps-ios/pull/694))
 
 ## 10.0.0-rc.9 - Sept 22, 2021
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ Mapbox welcomes participation and contributions from everyone.
 ### Breaking changes ⚠️
 
 * Removes default parameter values in the `addImage` function. ([#695](https://github.com/mapbox/mapbox-maps-ios/pull/695))
+* `public func layer<T: Layer>(withId id: String) throws -> T` has been updated to `public func layer<T>(withId id: String, type: T.Type) throws -> T where T: Layer`
+* `public func updateLayer<T: Layer>(withId id: String, update: (inout T) throws -> Void) throws` has been updated to `public func updateLayer<T>(withId id: String, type: T.Type, update: (inout T) throws -> Void) throws where T: Layer`
+* `public func source<T: Source>(withId id: String) throws -> T` has been updated to `public func source<T>(withId id: String, type: T.Type) throws -> T where T: Source`
 
 ## 10.0.0-rc.9 - Sept 22, 2021
 
--- a/Sources/MapboxMaps/Style/Generated/Sources/GeoJSONSource.swift
+++ b/Sources/MapboxMaps/Style/Generated/Sources/GeoJSONSource.swift
@@ -9,7 +9,7 @@ import Foundation
  */
 public struct GeoJSONSource: Source {
 
-    public let type: SourceType
+    public var type: SourceType
  
   
     /** 
--- a/Sources/MapboxMaps/Style/Generated/Sources/GeoJSONSource.swift
+++ b/Sources/MapboxMaps/Style/Generated/Sources/GeoJSONSource.swift
@@ -9,7 +9,7 @@ import Foundation
  */
 public struct GeoJSONSource: Source {
 
-    public var type: SourceType
+    public let type: SourceType
  
   
     /** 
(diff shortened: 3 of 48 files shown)


Apps/Examples/Examples/All Examples/ShowHideLayerExample.swift
```diff
@@ -95,7 +95,7 @@ class ShowHideLayerExample: UIViewController, ExampleProtocol {
         // Update the museum layer's visibility based on whether the switch
         // is on. `visibility` is `nil` by default.
         do {
-            try style.updateLayer(withId: museumLayerId) { (layer: inout CircleLayer) in
+            try style.updateLayer(withId: museumLayerId, type: CircleLayer.self) { (layer: inout CircleLayer) in
```
macdrevx, line 98: Is it possible to remove the type from the closure argument list now?

```suggestion
            try style.updateLayer(withId: museumLayerId, type: CircleLayer.self) { (layer) in
```

macdrevx: If so, let's apply that change throughout this PR

Sources/MapboxMaps/Style/Style.swift
```diff
@@ -149,12 +151,12 @@ public class Style {
...
     /**
      Retrieves a source from the map
-     - Parameter identifier: The id of the source to retrieve
+     - Parameter id: The id of the source to retrieve
      - Parameter type: The type of the source
      - Returns: The fully formed `source` object of type equal to `type`.
      - Throws: StyleError or type conversion errors
      */
-    public func source<T: Source>(withId id: String) throws -> T {
+    public func source<T>(withId id: String, type: T.Type) throws -> T where T: Source {
```
macdrevx, line 159: Similar question here as https://github.com/mapbox/mapbox-maps-ios/pull/694/files#r713365278

Tests/MapboxMapsTests/Annotations/Generated/CircleAnnotationIntegrationTests.swift
```diff
@@ -42,7 +42,7 @@ final class CircleAnnotationIntegrationTests: MapViewIntegrationTestCase {
         manager.annotations.append(annotation)
 
         expectation(for: NSPredicate(block: { (_, _) in
-            guard let layer: CircleLayer = try? self.style?.layer(withId: self.manager.layerId) else {
+            guard let layer: CircleLayer = try? self.style?.layer(withId: self.manager.layerId, type: CircleLayer.self) else {
```
macdrevx, line 45: Here you can remove the `: CircleLayer` due to this change. Please apply this throughout this PR.

Sources/MapboxMaps/Style/Style+Localization.swift
```diff
@@ -24,7 +24,7 @@ extension Style {
         }
 
         for layerInfo in symbolLayers {
-            let symbolLayer = try layer(withId: layerInfo.id) as SymbolLayer
+            let symbolLayer = try layer(withId: layerInfo.id, type: SymbolLayer.self) as SymbolLayer
```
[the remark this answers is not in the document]
neelmistry94: I tried that initially but then it gets a build failure saying that it needs to be casted.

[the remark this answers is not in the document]
macdrevx: This should be unnecessary. Let's investigate together because it could be an indication of a problem.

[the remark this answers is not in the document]
macdrevx: I tried this locally, and the issue is that we have 2 variants of this method:

    public func layer<T>(withId id: String, type: T.Type) throws -> T where T: Layer
    @_spi(Experimental) public func layer(withId id: String, type: Layer.Type) throws -> Layer

In this instance, the compiler thinks we mean the second one. But I don't think that's really what we want. Let's eliminate the second method altogether to resolve this. It doesn't provide any benefit over using the first one.

Sources/MapboxMaps/Style/Style.swift
```diff
@@ -67,13 +67,13 @@ public class Style {
...
      Gets a `layer` from the map
-     - Parameter layerID: The id of the layer to be fetched
+     - Parameter id: The id of the layer to be fetched
      - Parameter type: The type of the layer that will be fetched
 
      - Returns: The fully formed `layer` object of type equal to `type`
      - Throws: StyleError or type conversion errors
      */
-    public func layer<T: Layer>(withId id: String) throws -> T {
+    public func layer<T>(withId id: String, type: T.Type) throws -> T where T: Layer {
```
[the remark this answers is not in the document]
macdrevx: See https://github.com/mapbox/mapbox-maps-ios/pull/694#discussion_r714264968 -- we should definitely remove the method below.

Sources/MapboxMaps/Style/Generated/Sources/GeoJSONSource.swift
```diff
@@ -9,7 +9,7 @@ import Foundation
  */
 public struct GeoJSONSource: Source {
 
-    public let type: SourceType
+    public var type: SourceType
```
macdrevx, line 12: This should still be a let, right?

Sources/MapboxMaps/Style/Style+Localization.swift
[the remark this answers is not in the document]
macdrevx: Don't forget to remove the "as"

Sources/MapboxMaps/Style/Style.swift
```diff
@@ -107,8 +109,10 @@ public class Style {
     ///   - update: Closure that mutates a layer passed to it
     ///
     /// - Throws: StyleError or type conversion errors
-    public func updateLayer<T: Layer>(withId id: String, update: (inout T) throws -> Void) throws {
-        var layer: T = try self.layer(withId: id)
+    public func updateLayer<T>(withId id: String,
+                               type: T.Type,
+                               update: (inout T) throws -> Void) throws where T: Layer {
+        var layer: T = try self.layer(withId: id, type: T.self)
```
macdrevx, line 115: ": T" can be removed

Sources/MapboxMaps/Style/Generated/Sources/GeoJSONSource.swift
neelmistry94: I changed it to a var since I had to add it to the protocol. This is so we could follow the same pattern that we did for layer.

Sources/MapboxMaps/Style/Sources.swift
```diff
@@ -30,12 +27,32 @@ public enum SourceType: String, Codable {
...
+        case .model:
+            return ModelSource.self
+        }
+    }
 }
 
-public protocol Source: Codable, StyleEncodable, StyleDecodable { }
+public protocol Source: Codable, StyleEncodable, StyleDecodable {
+    /// Rendering type of this source.
+    var type: SourceType { get set }
```
macdrevx, line 54: If you change this to only requiring a getter, it can be satisfied by a let property. I think that makes more sense. We wouldn't want folks to be setting a type that doesn't match the struct/class.

Sources/MapboxMaps/Style/Generated/Sources/GeoJSONSource.swift
macdrevx: Replied in a new comment.

Sources/MapboxMaps/Style/Sources.swift
macdrevx: I think we should do this for layer too. @nishant-karajgikar would love your thoughts on this as well.

Sources/MapboxMaps/Style/Style+Localization.swift
```diff
@@ -49,7 +49,7 @@ extension Style {
         for sourceInfo in vectorSources {
             if locale.identifier.starts(with: "zh") {
                 // Force unwrapping since `allSourceIdentifiers` is getting a fresh list of valid sources
-                let vectorSource = try! source(withId: sourceInfo.id) as VectorSource
+                let vectorSource = try! source(withId: sourceInfo.id, type: VectorSource.self) as VectorSource
```
macdrevx, line 52: Here's another one where we need to remove "as". Could you do a careful self-review of this PR to find the rest of these cases and cases where the variable being assigned is unnecessarily given an explicit type?

Sources/MapboxMaps/Style/Sources.swift
neelmistry94: Note that this would require a style gen change, which is fine since I already have a matching PR for this work anyways.

CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ Mapbox welcomes participation and contributions from everyone.
 ### Breaking changes ⚠️
 
 * Removes default parameter values in the `addImage` function. ([#695](https://github.com/mapbox/mapbox-maps-ios/pull/695))
+* `public func layer<T: Layer>(withId id: String) throws -> T` has been updated to `public func layer<T>(withId id: String, type: T.Type) throws -> T where T: Layer`. ([#694](https://github.com/mapbox/mapbox-maps-ios/pull/694))
+* `public func updateLayer<T: Layer>(withId id: String, update: (inout T) throws -> Void) throws` has been updated to `public func updateLayer<T>(withId id: String, type: T.Type, update: (inout T) throws -> Void) throws where T: Layer`. ([#694](https://github.com/mapbox/mapbox-maps-ios/pull/694))
```
macdrevx, line 11: let's see whether it would make sense to use `rethrows` here

Sources/MapboxMaps/Style/Sources.swift
```diff
@@ -30,12 +27,32 @@ public enum SourceType: String, Codable {
...
+            return ModelSource.self
+        }
+    }
 }
 
-public protocol Source: Codable, StyleEncodable, StyleDecodable { }
+public protocol Source: Codable, StyleEncodable, StyleDecodable {
+    /// Rendering type of this source.
+    var type: SourceType { get }
+}
```
macdrevx, line 55: Anything else we want to add to this protocol? Doing so post-GA would be a breaking change.

Sources/MapboxMaps/Style/Style.swift
```diff
@@ -107,12 +109,13 @@ public class Style {
     ///   - update: Closure that mutates a layer passed to it
     ///
     /// - Throws: StyleError or type conversion errors
-    public func updateLayer<T: Layer>(withId id: String, update: (inout T) throws -> Void) throws {
-        var layer: T = try self.layer(withId: id)
+    public func updateLayer<T>(withId id: String,
+                               type: T.Type,
+                               update: (inout T) throws -> Void) throws where T: Layer {
+        var layer = try self.layer(withId: id, type: T.self)
```
macdrevx, line 115: I'd like to learn more about this change

Tests/MapboxMapsTests/Style/ModelLayerIntegrationTests.swift
```diff
@@ -42,7 +42,7 @@ class ModelLayerIntegrationTests: MapViewIntegrationTestCase {
 
             // Retrieve the layer
             do {
-                _ = try style.layer(withId: "test-id") as ModelLayer
+                _ = try style.layer(withId: "test-id", type: ModelLayer.self) as ModelLayer
```
macdrevx, line 45: Can remove `as`

macdrevx:
Any docs updates?

Sources/MapboxMaps/Style/ModelSource.swift
```diff
@@ -22,7 +22,7 @@ public struct Model: Equatable, Codable {
 /// A model data source used to power a `ModelLayer`
 internal struct ModelSource: Source {
 
-    internal let type: SourceType
+    internal var type: SourceType
```
macdrevx, line 25: Should be a let

macdrevx (approved):

PR #695, opened 2021-09-21
Enable adding cap insets

neelmistry94:
## Pull request checklist:

### Summary of changes
This PR removes default parameters from the `addImage` function so that it will not be a breaking change post GA.

### DIFF
--- a/Sources/MapboxMaps/Location/LocationStyleProtocol.swift
+++ b/Sources/MapboxMaps/Location/LocationStyleProtocol.swift
@@ -23,7 +23,7 @@ extension LocationStyleProtocol {
 
     internal func addImage(_ image: UIImage,
                            id: String,
-                           sdf: Bool = false,
+                           sdf: Bool,
                            stretchX: [ImageStretches],
                            stretchY: [ImageStretches],
                            content: ImageContent?) throws {
--- a/Sources/MapboxMaps/Location/LocationStyleProtocol.swift
+++ b/Sources/MapboxMaps/Location/LocationStyleProtocol.swift
@@ -16,22 +16,22 @@ internal protocol LocationStyleProtocol: AnyObject {
                   content: ImageContent?) throws
 }
 
-extension LocationStyleProtocol {
-    internal func addPersistentLayer(_ layer: Layer, layerPosition: LayerPosition? = nil) throws {
-        try addPersistentLayer(layer, layerPosition: layerPosition)
-    }
-
-    internal func addImage(_ image: UIImage,
-                           id: String,
-                           sdf: Bool,
-                           stretchX: [ImageStretches],
-                           stretchY: [ImageStretches],
-                           content: ImageContent? = nil) throws {
-        try addImage(image,
-                     id: id,
-                     sdf: sdf,
-                     stretchX: stretchX,
-                     stretchY: stretchY,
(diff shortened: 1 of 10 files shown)


Sources/MapboxMaps/Location/LocationStyleProtocol.swift
```diff
@@ -8,15 +8,30 @@ internal protocol LocationStyleProtocol: AnyObject {
...
+                           sdf: Bool,
+                           stretchX: [ImageStretches],
+                           stretchY: [ImageStretches],
+                           content: ImageContent?) throws {
+        try addImage(image,
+                     id: id,
+                     sdf: sdf,
+                     stretchX: stretchX,
+                     stretchY: stretchY,
+                     content: content)
```
macdrevx, line 35: Trying to understand how this isn't a recursive call

macdrevx: This is in a protocol extension which must have something to do with it. Let's see if we can find a way not to do this.

Sources/MapboxMaps/Style/Style.swift
```diff
@@ -471,7 +471,12 @@ extension Style: StyleManagerProtocol {
 
     // MARK: - Style images
 
-    public func addImage(_ image: UIImage, id: String, sdf: Bool = false, stretchX: [ImageStretches] = [], stretchY: [ImageStretches] = [], content: ImageContent? = nil) throws {
+    public func addImage(_ image: UIImage,
+                         id: String,
+                         sdf: Bool = false,
+                         stretchX: [ImageStretches],
+                         stretchY: [ImageStretches],
+                         content: ImageContent?) throws {
```
macdrevx, line 479: Let's retain the default value for `content`.

Sources/MapboxMaps/Location/LocationStyleProtocol.swift
neelmistry94: I have to dig into this further, I'm not sure how involved the `LocationStyleProtocol` is.

macdrevx: I tried this locally — if you just delete this entire protocol extension, there just end up being 2 places that were depending on the default parameter for `layerPosition` assigned here where you'll need to explicitly pass nil instead. I think that's a pretty reasonable tradeoff for deleting these 20 lines of confusing code.

Sources/MapboxMaps/Location/LocationStyleProtocol.swift
```diff
@@ -8,15 +8,30 @@ internal protocol LocationStyleProtocol: AnyObject {
...
+//                           stretchY: [ImageStretches],
+//                           content: ImageContent? = nil) throws {
+//        try addImage(image,
+//                     id: id,
+//                     sdf: sdf,
+//                     stretchX: stretchX,
+//                     stretchY: stretchY,
+//                     content: content)
+//    }
+//}
```
macdrevx, line 37: Please delete instead of commenting-out

neelmistry94: whoops, my bad

macdrevx (approved):

PR #696, opened 2021-09-22
Gesture Options Cleanup

macdrevx:
## Pull request checklist:
 - [x] Update the changelog

### Summary of changes

- Updates "enabled" gesture options to have a 1-1 mapping with available gestures.
- Changes "double tap to zoom out" gesture to "double touch to zoom out" gesture for consistency with Android SDK, and splits the corresponding handler implementations into 2 classes.
- Merges the rotate gesture into the pinch gesture, which allows correct anchoring of the rotation around the midpoint of the gesture.
- Adds logic to ensure that the initial state of the gesture handlers and recognizers is consistent with the default values specified on the GestureOptions struct.

### Breaking changes

- `GestureManager.rotationGestureRecognizer` has been removed. Rotation is now handled by `.pinchGestureRecognizer` in addition to its preexisting handling of panning and zooming.
- `GestureManager.doubleTapToZoomOutGestureRecognizer` has been replaced with `.doubleTouchToZoomOutGestureRecognizer`
- `PanScrollingMode` has been renamed to `PanMode`
- `GestureOptions.zoomEnabled` has been replaced by `.doubleTapToZoomInEnabled`, `.doubleTouchToZoomOutEnabled`, and `.quickZoomEnabled`
- `GestureOptions.rotateEnabled` has been removed
- `GestureOptions.scrollEnabled` has been renamed to `.panEnabled`
- `GestureOptions.scrollingMode` has been renamed to `.panMode`
- `GestureOptions.decelerationRate` has been renamed to `.panDecelerationFactor`
- `GestureType.doubleTapToZoomOut` has been replaced with `.doubleTouchToZoomOut`
- `GestureType.rotate` has been removed
- `GestureType` cases have been reordered for consistency with `GestureOptions` and `GestureManager`

### Bug Fixes

- Fixes an issue where tapping the compass could fail to set the bearing to 0 if there was already an animation running. Tapping the compass now cancels any existing animations.
- Fixes issues with the pinch gesture when removing and re-adding one of the two required touches
- Fixes an issue where a pan gesture would fail if it interrupted the deceleration from a previous pan gesture.


Sources/MapboxMaps/Gestures/GestureOptions.swift
```diff
@@ -1,100 +1,52 @@
...
+    public var panMode: PanMode = .horizontalAndVertical
 
-    /**
-        A floating-point value that determines the rate of deceleration after the user lifts their finger.
-    */
-    public var decelerationRate: CGFloat = UIScrollView.DecelerationRate.normal.rawValue
+    /// A constant factor that determines how quickly pan deceleration animations happen.
+    /// Multiplied with the velocity vector once per millisecond during deceleration animations.
+    /// Defaults to `UIScrollView.DecelerationRate.normal.rawValue`
+    public var panDecelerationFactor: CGFloat = UIScrollView.DecelerationRate.normal.rawValue
```
neelmistry94, line 49: everywhere else in the code, this has been called decelerationFactor. Why is it that the options have a prepended pan in front of it?

Sources/MapboxMaps/Gestures/GestureType.swift
```diff
@@ -1,24 +1,19 @@
...
-    /// The rotate gesture
-    case rotate
+    /// The pitch gesture
+    case pitch
+
+    /// The double tap to zoom in gesture
+    case doubleTapToZoomIn
+
+    /// The double touch to zoom out gesture
+    case doubleTouchToZoomOut
```
neelmistry94, line 15: nit: Let's order these alphabetically

neelmistry94 (approved):

Sources/MapboxMaps/Gestures/GestureHandlers/PanGestureHandler.swift
```diff
@@ -42,31 +56,46 @@ internal final class PanGestureHandler: GestureHandler {
...
+                    // it immediately after starting the animation
+                    self.handleChange(
+                        withTouchLocation: touchLocation,
+                        initialTouchLocation: initialTouchLocation,
+                        initialCameraState: initialCameraState)
+                },
                 completion: {
-                    self.initialTouchLocation = nil
-                    self.initialCameraState = nil
-                    self.lastChangedDate = nil
```
nishant-karajgikar, line 68: Should we make this completion optional? Seems like it should be nullable instead of sending an empty closure here.

nishant-karajgikar: let's address this in a future pr.

Sources/MapboxMaps/Gestures/GestureOptions.swift
macdrevx: We want to indicate that this setting only applies to pan. In the other locations, the meaning is clear from the context. Later, we might add other separate deceleration factor options for other gestures, so the prefix sets us up to be able to do that in an additive way.

Sources/MapboxMaps/Gestures/GestureType.swift
macdrevx: Can we debate this in a separate ticket?

PR #700, opened 2021-09-22
Use explicit self to allow Xcode versions <12.5 to compile

nishant-karajgikar:
Xcode 12.5 fixes an issue where the scope of some variables weren't being figured out correctly by the compiler. Bug: https://bugs.swift.org/browse/SR-8669


nishant-karajgikar:
Moved this commit into #690

PR #703, opened 2021-09-22
v10.0.0-rc.9 docs into prod

neelmistry94:
Moving v10.0.0-rc.9 docs into prod


danswick (approved):
tested locally and all looks great!

Issue #500, opened 2021-06-27
Point Annotation


666bsw:
## Environment

- Xcode version: 12.5
- iOS version: latest
- Devices affected: all
- Maps SDK Version:V10. RC-2

## Observed behavior and steps to reproduce

1. Point Annotation halo is below the coordinates of the annotation.
2. Couldn't find a way to make the annotation layer above the line layer. The line closes the annotation.

https://ibb.co/Y3VfjfK

ZiZasaurus:
@666bsw please note that the order of your annotations/markers matters. Can you add your annotations after the line has been added to the map and confirm that it appears as expected? You can use the following methods to do so: 

- Using makePointAnnotationManager()

- Creating PolylineAnnotationManager first and then PointAnnotationManager second. 

I’m not sure what you meant by “Point Annotation halo is below the coordinates of the annotation” or “the line closes the annotation.” Would you be able to clarify these two points?

666bsw:
Hello!
with one problem solved, thanks.

Second problem
The annotation tip is not at the actual coordinates.
The center of the annotation is located on the actual coordinates.
This can be seen in the attached screenshot, the tip of the annotation is below the line

666bsw:
Hello!
with one problem solved, thanks.

Second problem
The annotation tip is not at the actual coordinates.
The center of the annotation is located on the actual coordinates.
This can be seen in the attached screenshot, the tip of the annotation is below the line

The attachment
1.how should the annotation be positioned
2. this is how the mapbox annotation is now positioned

macdrevx:
@666bsw thanks for clarifying. I think that the problem is that setting the iconImage to `.default` doesn't automatically adjust the `iconAnchor` or `iconOffset` to make it so that the point of the marker image is over the correct coordinate. I agree that this behavior is surprising, and we will discuss how we can improve this API.

In the meantime, you'll need to tune the position of the image using `iconAnchor` and `iconOffset`.

666bsw:
Hello!

thanks for the answer

ZiZasaurus:
@666bsw we have removed this feature and will re-evaluate bringing it back in the future. Please feel free to reopen if you have additional questions.

666bsw:
Hello!
thanks for the answer )

PR #707, opened 2021-09-24
Remove default marker image

macdrevx:
## Pull request checklist:
 - [x] Update the changelog

### Summary of changes

As pointed out by https://github.com/mapbox/mapbox-maps-ios/issues/500, assigning `.default` to `PointAnnotation.image` failed to adjust any of the necessary properties to ensure that the point of the marker image was placed directly over the annotation's coordinate.

After some consideration, we've decided to remove the default marker image. This approach keeps the implementation simple (setting `.image` won't magically manipulate the values of other properties) while leaving the door open for us to bring back a default annotation image feature in a non-breaking way in the future. Notably, the Android SDK also does not have a default marker feature, so this approach will allow us to reconsider how to add this feature in a way that is consistent between the two platforms.

#### Breaking Changes

- Converts `PointAnnotation.Image` from an `enum` to a `struct`.
- Removes `PointAnnotation.Image.default`
- Replaces `PointAnnotation.Image.custom` with `PointAnnotation.Image.init(image:name:)`

#### Other Changes

- Removes "PointAnnotationExample" in favor of "CustomPointAnnotationExample"
- Resolves Xcode 13 upgrade check warnings


jmkiley (approved):
Examples updates LGTM.

Sources/MapboxMaps/Annotations/PointAnnotation+ImageHandling.swift
```diff
@@ -22,114 +18,16 @@ extension PointAnnotationManager {
...
-                    case .custom(image: let image, name: let name): // Add this custom image
-                        try style.addImage(image,
-                                           id: name,
-                                           stretchX: [],
-                                           stretchY: [])
-                    }
+            // If the image is not found, add it to the style
+            if style.image(withId: pointAnnotationImage.name) == nil {
+                do {
+                    try style.addImage(pointAnnotationImage.image, id: pointAnnotationImage.name, stretchX: [], stretchY: [])
```
nishant-karajgikar, line 24: Do we need handling for cap insets here?

Sources/MapboxMaps/Annotations/PointAnnotation+ImageHandling.swift
```diff
@@ -2,17 +2,13 @@ import UIKit
 @_implementationOnly import MapboxCommon_Private
 
 extension PointAnnotation {
-    public enum Image: Hashable {
-        case `default`
-        case custom(image: UIImage, name: String)
+    public struct Image: Hashable {
+        var image: UIImage
+        var name: String
```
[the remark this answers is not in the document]
macdrevx: It's a more flexible API to let them be `var`s. There's no real reason to prevent folks from mutating them.

Sources/MapboxMaps/Annotations/PointAnnotation+ImageHandling.swift
macdrevx: Seems like that would be a new feature. Maybe we would want to do it, but I'd have to give it more thought. Regardless, if folks need this, they could bypass `PointAnnotation.image` and use `iconImage` + style image APIs directly.

nishant-karajgikar (approved):

PR #709, opened 2021-09-24
Redesign single tap event handling for annotations

nishant-karajgikar:

## Pull request checklist:
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-ios` changelog: `<changelog></changelog>`.

### Summary of changes

This PR redesigns the way single tap gestures work in the MapView and each *AnnotationManager. The changes are motivated by a desire to make the `GestureManager` the home of all gesture related functionality. Accordingly, the GestureManager now owns a `SingleTapGestureHandler` (and  exposes its associated gesture recognizer). 

The `singleTapGestureRecognizer` is injected into the `AnnotationsOrchestrator` and then into any constructed `*AnnotationManager`s. 

From a user's perspective, setting the `AnnotationInteractionDelegate` on an annotation manger results in a target-action being added to the singleTapGestureRecognizer.

### DIFF
--- a/Sources/MapboxMaps/Annotations/AnnotationOrchestrator.swift
+++ b/Sources/MapboxMaps/Annotations/AnnotationOrchestrator.swift
@@ -39,7 +39,7 @@ public protocol AnnotationInteractionDelegate: AnyObject {
 
 public class AnnotationOrchestrator {
 
-    private weak var view: UIView?
+    private weak var singleTapGestureRecognizer: UIGestureRecognizer?
 
     private let style: Style
 
@@ -47,8 +47,11 @@ public class AnnotationOrchestrator {
 
     private weak var displayLinkCoordinator: DisplayLinkCoordinator?
 
-    internal init(view: UIView, mapFeatureQueryable: MapFeatureQueryable, style: Style, displayLinkCoordinator: DisplayLinkCoordinator) {
-        self.view = view
+    internal init(singleTapGestureRecognizer: UIGestureRecognizer,
+                  mapFeatureQueryable: MapFeatureQueryable,
+                  style: Style,
+                  displayLinkCoordinator: DisplayLinkCoordinator) {
+        self.singleTapGestureRecognizer = singleTapGestureRecognizer
         self.mapFeatureQueryable = mapFeatureQueryable
         self.style = style
         self.displayLinkCoordinator = displayLinkCoordinator
@@ -62,14 +65,14 @@ public class AnnotationOrchestrator {
     public func makePointAnnotationManager(id: String = String(UUID().uuidString.prefix(5)),
                                            layerPosition: LayerPosition? = nil) -> PointAnnotationManager {
 
-        guard let view = view,
+        guard let singleTapGestureRecognizer = singleTapGestureRecognizer,
               let displayLinkCoordinator = displayLinkCoordinator else {
-            fatalError("View and displayLinkCoordinator must be present when creating an annotation manager")
+            fatalError("SingleTapGestureRecognizer and displayLinkCoordinator must be present when creating an annotation manager")
         }
 
         return PointAnnotationManager(id: id,
                                       style: style,
-                                      view: view,
+                                      singleTapGestureRecognizer: singleTapGestureRecognizer,
                                       mapFeatureQueryable: mapFeatureQueryable,
                                       shouldPersist: true,
                                       layerPosition: layerPosition,
@@ -84,14 +87,14 @@ public class AnnotationOrchestrator {
     public func makePolygonAnnotationManager(id: String = String(UUID().uuidString.prefix(5)),
                                              layerPosition: LayerPosition? = nil) -> PolygonAnnotationManager {
 
-        guard let view = view,
+        guard let singleTapGestureRecognizer = singleTapGestureRecognizer,
               let displayLinkCoordinator = displayLinkCoordinator else {
-            fatalError("View and displayLinkCoordinator must be present when creating an annotation manager")
+            fatalError("SingleTapGestureRecognizer and displayLinkCoordinator must be present when creating an annotation manager")
         }
 
         return PolygonAnnotationManager(id: id,
                                         style: style,
-                                        view: view,
+                                        singleTapGestureRecognizer: singleTapGestureRecognizer,
                                         mapFeatureQueryable: mapFeatureQueryable,
                                         shouldPersist: true,
                                         layerPosition: layerPosition,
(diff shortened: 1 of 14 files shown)


nishant-karajgikar:
Adding more tests..

Sources/MapboxMaps/Annotations/AnnotationOrchestrator.swift
```diff
@@ -39,16 +39,19 @@ public protocol AnnotationInteractionDelegate: AnyObject {
 
 public class AnnotationOrchestrator {
 
-    private weak var view: UIView?
+    private weak var singleTapGestureRecognizer: UIGestureRecognizer?
```
macdrevx, line 42: I don't think this needs to be `weak` anymore. `UIGestureRecognizer` only keeps a weak ref to its view, so there's no risk of a cycle.

Sources/MapboxMaps/Annotations/Generated/CircleAnnotationManager.swift
```diff
@@ -33,7 +33,7 @@ public class CircleAnnotationManager: AnnotationManager {
     private let mapFeatureQueryable: MapFeatureQueryable
 
     /// Dependency required to add gesture recognizer to the MapView
-    private weak var view: UIView?
+    private weak var singleTapGestureRecognizer: UIGestureRecognizer?
```
macdrevx, line 36: Same here - no need for weak

Sources/MapboxMaps/Annotations/Generated/CircleAnnotationManager.swift
```diff
@@ -206,36 +206,23 @@ public class CircleAnnotationManager: AnnotationManager {
...
-            if delegate != nil {
-                setupTapRecognizer()
-            } else {
-                guard let view = view, let recognizer = tapGestureRecognizer else { return }
-                view.removeGestureRecognizer(recognizer)
-                tapGestureRecognizer = nil
+            if delegate != nil && oldValue == nil {
+                singleTapGestureRecognizer?.addTarget(self, action: #selector(handleTap(_:)))
+            } else if delegate == nil && oldValue != nil {
+                singleTapGestureRecognizer?.removeTarget(self, action: #selector(handleTap(_:)))
```
macdrevx, line 217: It'd be simpler and less error-prone to just set up the target/action once in init and check for a non-nil delegate in the handler.

macdrevx: If you do this, you won't even need to store a ref to the gesture recognizer

Sources/MapboxMaps/Gestures/GestureHandlers/SingleTapGestureHandler.swift
```diff
@@ -0,0 +1,27 @@
...
+    @objc private func handleGesture(_ gestureRecognizer: UITapGestureRecognizer) {
+        switch gestureRecognizer.state {
+        case .recognized:
+            cameraAnimationsManager.cancelAnimations()
+            delegate?.gestureBegan(for: .singleTap)
+        default:
+            break
+        }
+    }
+}
```
macdrevx, line 27: I think we should remove this. The `cancelAnimations` will soon be handled centrally, and limiting the delegate to only deal with gestures that manipulate the camera seems like a reasonable design as well.

macdrevx: From discussion: we'll keep this handler for consistency, and will update it to handle the new delegate methods after landing https://github.com/mapbox/mapbox-maps-ios/pull/697

Sources/MapboxMaps/Annotations/Generated/CircleAnnotationManager.swift
```diff
@@ -206,36 +205,18 @@ public class CircleAnnotationManager: AnnotationManager {
...
+    @objc internal func handleTap(_ tap: UITapGestureRecognizer) {
 
-    internal func setupTapRecognizer() {
-        let tapRecognizer = UITapGestureRecognizer(target: self, action: #selector(handleTap(_:)))
-        tapRecognizer.numberOfTapsRequired = 1
-        tapRecognizer.numberOfTouchesRequired = 1
-        view?.addGestureRecognizer(tapRecognizer)
-        tapGestureRecognizer = tapRecognizer
-    }
+        guard let delegate = delegate else { return }
```
macdrevx, line 215: Since qrf is async, I wouldn't capture a strong ref to the delegate since that could result in messages to the delegate after it's been set to nil. The check here should just be an optimization to skip qrf if we know the delegate is nil.

Sources/MapboxMaps/Gestures/GestureManager.swift
```diff
@@ -64,6 +64,11 @@ public final class GestureManager: GestureHandlerDelegate {
         return quickZoomGestureHandler.gestureRecognizer
     }
 
+    /// The gesture recognizer for the single tap gesture
```
macdrevx, line 67: Would like to see a description of how this relates to annotation interaction delegate

Sources/MapboxMaps/Gestures/GestureHandlers/SingleTapGestureHandler.swift
```diff
@@ -0,0 +1,27 @@
+import UIKit
+
+/// `SingleTapGestureHandler` manages a gesture recognizer looking for single tap touch events
+internal final class SingleTapGestureHandler: GestureHandler {
+
+    internal init(gestureRecognizer: UITapGestureRecognizer,
+                  mapboxMap: MapboxMapProtocol,
```
macdrevx, line 7: I'm feeling less good about the existence of this handler again. It's fundamentally different than the other gestures since it doesn't manipulate the map, but still depends on mapboxMap since that's a requirement of the base class.

Tests/MapboxMapsTests/Gestures/GestureManagerTests.swift
```diff
@@ -11,6 +11,7 @@ final class GestureManagerTests: XCTestCase {
     var doubleTapToZoomInGestureHandler: GestureHandler!
     var doubleTouchToZoomOutGestureHandler: GestureHandler!
     var quickZoomGestureHandler: GestureHandler!
+    var singleTapGestureHandler: GestureHandler!
```
macdrevx, line 14: needs a "delegate assignment" test too. see examples below.

Sources/MapboxMaps/Gestures/GestureHandlers/SingleTapGestureHandler.swift
nishant-karajgikar: Yeah this part bothers me as well. I'd love to move away from GestureHandler being a super class to be honest. It feels like it'd be better suited to be a protocol.

Sources/MapboxMaps/Annotations/Generated/CircleAnnotationManager.swift
nishant-karajgikar: ! , thank you.

Sources/MapboxMaps/Gestures/GestureManager.swift
nishant-karajgikar: Hmm not sure if that context should live in the gesture manager. Seems like some information that should be housed in the annotation manager.

macdrevx: why not both?

I'm thinking about the scenario we discussed yesterday about how it might be unclear to developers what the effect of disabling this gesture recognizer might be.

Sources/MapboxMaps/Gestures/GestureHandlers/SingleTapGestureHandler.swift
macdrevx: okay. let's leave it for now. we can refactor it later

macdrevx (approved):

PR #714, opened 2021-09-28
Update offline manager examples to avoid internal initializer

1ec5:
Fixed the compiler error introduced in https://github.com/mapbox/mapbox-maps-ios/pull/683#discussion_r717427570.

/cc @jmkiley @neelmistry94

### DIFF
--- a/Apps/Examples/Examples/All Examples/OfflineManagerExample.swift
+++ b/Apps/Examples/Examples/All Examples/OfflineManagerExample.swift
@@ -124,7 +124,7 @@ final class OfflineManagerExample: UIViewController, ExampleProtocol {
 
         // Load the tile region
         let tileRegionLoadOptions = TileRegionLoadOptions(
-            geometry: MapboxCommon.Geometry(coordinate: tokyoCoord),
+            geometry: MapboxCommon.Geometry(point: tokyoCoord as NSValue),
             descriptors: [outdoorsDescriptor],
             metadata: ["tag": "my-outdoors-tile-region"],
             acceptExpired: true)!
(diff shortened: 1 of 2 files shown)


1ec5:
This PR fixes the builds in anticipation of #711, which replaces this usage with something more type-safe.

Apps/Examples/Examples/All Examples/OfflineManagerExample.swift
```diff
@@ -124,7 +124,7 @@ final class OfflineManagerExample: UIViewController, ExampleProtocol {
 
         // Load the tile region
         let tileRegionLoadOptions = TileRegionLoadOptions(
-            geometry: MapboxCommon.Geometry(coordinate: tokyoCoord),
+            geometry: MapboxCommon.Geometry(point: tokyoCoord as NSValue),
```
macdrevx, line 127: we typically convert the CLLocationCoordinate to a CGPoint and then to NSValue. I think that will yield different results than this since the lat/lng are reversed.

Regardless, the need for this workaround will be short-lived since we're in the process of refining this interface to work with Turf.

neelmistry94:
@1ec5 @macdrevx I believe this issue will be fixed in https://github.com/mapbox/mapbox-maps-ios/pull/711

Apps/Examples/Examples/All Examples/OfflineManagerExample.swift
macdrevx: Nevermind, I looked at this some more, and I think it works as expected. Our implementation doesn't actually flip lat/lng in this case.

macdrevx (approved):

PR #725, opened 2021-09-30
Hold on to annotation managers once constructed

nishant-karajgikar:
## Pull request checklist:
 - [x] Update the changelog

### Summary of changes

Annotation managers are now kept alive by the `AnnotationOrchestrator`. This means that any annotations added to an annotation manager will continue to be displayed until one of the following occurs:

* the map view is destroyed
* the annotation manager is removed explicitly via a call to `removeAnnotationManager(withId:)`
* the annotation manager is removed implicitly by creating another annotation manager with the same ID

#### Public API Changes

* The `AnnotationManager` protocol now conforms to `AnyObject`
* `AnnotationOrchestrator.annotationManagersById` has been added. This dictionary contains all annotation managers that have not been removed.
* `AnnotationOrchestrator.removeAnnotationManager(withId:)` has been added. Use this method to explicitly remove an annotation manager.

#### Other Changes

* Cleaned up DebugViewController
* Updated examples for consistency with this behavior change
* Updated access control in modified examples to be appropriate for an app (no need for `public`, implicit `internal`)
* Added an internal protocol `AnnotationManagerInternal` which conforms to `AnnotationManager` and to which all annotation managers now conform. This protocol exposes the `destroy()` method which the `AnnotationOrchestrator` uses when removing an annotation manager.
* Annotation managers no longer remove their source and layer on deinit — `destroy()` must be called prior to deinit.
* When annotation managers are destroyed, they now also remove themselves as targets of the tap gesture recognizer and remove themselves from the list of display link participants.
* Reorganized the contents of MBXGeometry.swift into Geometry.swift for consistency with Feature.swift
* Updated the `mapView` and `style` properties of `MapViewIntegrationTestCase` to be implicitly unwrapped.


Sources/MapboxMaps/Annotations/AnnotationOrchestrator.swift
```diff
@@ -57,6 +57,10 @@ public class AnnotationOrchestrator {
         self.displayLinkCoordinator = displayLinkCoordinator
     }
 
+    /// Dictionary of annotation managers keyed by their identifiers.
+    /// - NOTE: Removing an annotation manager from this collection can result in the corresponding annotations to be removed from the map view.
+    public var annotationManagersMap: [String: AnnotationManager] = [:]
```
macdrevx, line 62: Let's name this annotationManagersById

macdrevx: And can we make the setter private?

macdrevx: And add a "remove annotation manager" or something like that

macdrevx: https://github.com/mapbox/mapbox-maps-android/blob/main/plugin-annotation/src/main/java/com/mapbox/maps/plugin/annotation/AnnotationPluginImpl.kt#L63-L68

Something like that

nishant-karajgikar: Agreed, addressed.

Sources/MapboxMaps/Annotations/AnnotationOrchestrator.swift
```diff
@@ -132,12 +147,23 @@ public class AnnotationOrchestrator {
...
+        annotationManagersById[id] = annotationManager
+
+        return annotationManager
+    }
+
+    /// Removes an annotation manager, this will remove the underlying layer and source from the style.
+    /// A removed annotation manager will not be able to reuse anymore, you will need to create new annotation manger to add annotations.
+    /// - Parameter id: Identifer of annotation manager to remove
+    public func removeAnnotationManager(withId id: String) {
+        annotationManagersById[id] = nil
```
neelmistry94, line 167: Rather than set it to nil, can we just remove it all together? using the dictionary [remove function](https://developer.apple.com/documentation/swift/dictionary/2831050-remove)

jmkiley (approved):

PR #728, opened 2021-10-01
Update titles/descriptions of style/expression examples

colleenmcginnis:
This PR suggests some ways to make the descriptions more specific and align with similar [Android examples](https://github.com/mapbox/mapbox-maps-android/pull/683). I think including a generic title and a specific description (mentioning specific classes/methods/etc) will help with discoverability when we publish examples to docs.mapbox.com pages (I'm pulling these titles and descriptions programmatically into the metadata for each example page).

The thinking is that examples will show up more prominently in search results if a developer is using a query like "switchCase" if they know they want to use that expression, but want to see an example in the larger context of a view.

@sbma44 @mapbox/maps-ios I did my best to summarize the classes/methods used for the primary purpose of each example, but feel free to make direct edits if any of my interpretations of the example code are incorrect!


neelmistry94 (approved):

Apps/Examples/Examples/Models/Examples.swift
```diff
@@ -199,22 +199,22 @@ public struct Examples {
                 description: "Add a custom rendered Metal layer.",
                 type: CustomLayerExample.self),
         Example(title: "Add a line with a color gradient",
-                description: "Add a line with a rainbow color gradient.",
+                description: "Load a polyline to a style using GeoJsonSource, display it on a map using LineLayer, and style it with a rainbow color gradient.",
```
1ec5, line 202: GeoJSONSource is capitalized according to [Swift naming conventions](https://swift.org/documentation/api-design-guidelines/#follow-case-conventions) rather than Java/JavaScript naming conventions.

```suggestion
                description: "Load a polyline to a style using GeoJSONSource, display it on a map using LineLayer, and style it with a rainbow color gradient.",
```

macdrevx (approved):

PR #730, opened 2021-10-01
GeoJSON objects as expression arguments

### LINKED ISSUE #530: Improve GeoJSON object support for `distance` and `within` expressions
## New Feature

We should improve support for the [`distance`](https://docs.mapbox.com/mapbox-gl-js/style-spec/expressions/#distance) and [`within`](https://docs.mapbox.com/mapbox-gl-js/style-spec/expressions/#within) expressions. These expressions each take a GeoJSON object. Expressions arguments need to conform to `ExpressionArgumentBuilder`. Turf features currently do not. 

One option may be to add a `geoJSONObject` [argument](https://github.com/mapbox/mapbox-maps-ios/blob/cde53c67406bba7cd2d18d0ff5d290b835fb80a0/Sources/MapboxMaps/Style/Types/Expression.swift#L128-L136).

## Why

Because `Turf.GeoJSONObject` and Turf geometries do not conform to `ExpressionArgumentBuilder`, users currently need to add conformity in their own implementations. This means that users need to extend the feature or GeoJSON object class for their use case so that it conforms to the `ExpressionArgumentBuilder` protocol. They then need to create and pass their geometry object into their `distance` or `within` expression.

cc @nishant-karajgikar

1ec5:
Added a case to ExpressionArgument that takes a `GeoJSONObject` from Turf.

For example, to filter a symbol layer to only show symbols in Colorado:

```swift
let coloradoCorners: [CLLocationCoordinate2D] = [
    .init(latitude: 37, longitude: -109-2/60-48/60/60),
    .init(latitude: 37, longitude: -102-2/60-48/60/60),
    .init(latitude: 41, longitude: -102-2/60-48/60/60),
    .init(latitude: 41, longitude: -109-2/60-48/60/60),
    .init(latitude: 37, longitude: -109-2/60-48/60/60),
]
var colorado = Geometry.polygon(Polygon([coloradoCorners]))

symbolLayer.filter = Exp(.within) {
    GeoJSONObject.geometry(colorado)
}
```

This might be a backwards-compatible change, since `Expression.Argument` isn’t open to arbitrary `ExpressionArgumentConvertible`-conforming types. However, making this change after v2.0.0 would break backwards compatibility, because application code might switch on the cases of `Expression.Argument`.

Fixes #530.

/cc @jmkiley @macdrevx

### DIFF
--- a/Sources/MapboxMaps/Style/Types/Expression.swift
+++ b/Sources/MapboxMaps/Style/Types/Expression.swift
@@ -137,6 +137,7 @@ public struct Expression: Codable, CustomStringConvertible, Equatable {
         case numberArray([Double])
         case stringArray([String])
         case option(Option)
+        case geoJSONObject(GeoJSONObject)
         case null
         case expression(Expression)
 
@@ -154,33 +155,12 @@ public struct Expression: Codable, CustomStringConvertible, Equatable {
                 return "\(exp)"
             case .option(let option):
                 return "\(option)"
+            case .geoJSONObject(let object):
+                return "\(object)"
             case .numberArray(let array):
                 return "\(array)"
             case .stringArray(let stringArray):
                 return "\(stringArray)"
-            }
-        }
-
-        public static func == (lhs: Expression.Argument, rhs: Expression.Argument) -> Bool {
-            switch (lhs, rhs) {
-            case (.number(let lhsNumber), .number(let rhsNumber)):
-                return lhsNumber == rhsNumber
-            case (.string(let lhsString), .string(let rhsString)):
-                return lhsString == rhsString
-            case (.boolean(let lhsBool), .boolean(let rhsBool)):
-                return lhsBool == rhsBool
-            case (.option(let lhsOption), .option(let rhsOption)):
-                return lhsOption == rhsOption
-            case (.null, .null):
-                return true
-            case (.expression(let lhsExpression), .expression(let rhsExpression)):
-                return lhsExpression == rhsExpression
-            case (.numberArray(let lhsArray), .numberArray(let rhsArray)):
-                return lhsArray == rhsArray
-            case (.stringArray(let lhsArray), .stringArray(let rhsArray)):
-                return lhsArray == rhsArray
-            default:
-                return false
             }
         }
 
@@ -198,6 +178,8 @@ public struct Expression: Codable, CustomStringConvertible, Equatable {
                 try container.encode(boolean)
             case .option(let option):
                 try container.encode(option)
+            case .geoJSONObject(let object):
+                try container.encode(object)
             case .null:
                 try container.encodeNil()
             case .numberArray(let array):
@@ -217,6 +199,8 @@ public struct Expression: Codable, CustomStringConvertible, Equatable {
                 self = .number(validNumber)
             } else if let validBoolean = try? container.decode(Bool.self) {
                 self = .boolean(validBoolean)
+            } else if let object = try? container.decode(GeoJSONObject.self) {
+                self = .geoJSONObject(object)
             } else if let validExpression = try? container.decode(Expression.self) {
                 self = .expression(validExpression)
             } else if let validOption = try? container.decode(Option.self) {
--- a/Sources/MapboxMaps/Style/Types/ExpressionArgumentBuilder.swift
+++ b/Sources/MapboxMaps/Style/Types/ExpressionArgumentBuilder.swift
@@ -90,3 +90,9 @@ extension UIColor: ExpressionArgumentConvertible {
         return [.string(StyleColor(self).rgbaString)]
     }
 }
+
+extension GeoJSONObject: ExpressionArgumentConvertible {
+    public var expressionArguments: [Expression.Argument] {
+        return [.geoJSONObject(self)]
+    }
+}
--- a/Tests/MapboxMapsTests/Style/ExpressionTests/ExpressionTests.swift
+++ b/Tests/MapboxMapsTests/Style/ExpressionTests/ExpressionTests.swift
@@ -1,5 +1,6 @@
 import XCTest
 @testable import MapboxMaps
+import CoreLocation
 
 final class ExpressionTests: XCTestCase {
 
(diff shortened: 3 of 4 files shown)


Sources/MapboxMaps/Style/Types/ExpressionArgumentBuilder.swift
```diff
@@ -90,3 +91,16 @@ extension UIColor: ExpressionArgumentConvertible {
...
+
+extension GeoJSONObject: ExpressionArgumentConvertible {
+    public var expressionArguments: [Expression.Argument] {
+        do {
+            let data = try JSONEncoder().encode(self)
+            let expression = try JSONDecoder().decode(Expression.Argument.self, from: data)
+            return [expression]
+        } catch {
+            // Throwable getters are new to Swift 5.5, so we’ll have to live with a fatal error for now.
+            fatalError("Unable to convert GeoJSON object to expression argument: \(error)")
```
1ec5, line 103: Alternatively, we could turn `ExpressionArgumentConvertible.expressionArguments` into a method until we can require Swift 5.5 to get throwable getters. Changing the property to a method would be a breaking change.

Sources/MapboxMaps/Style/Types/ExpressionArgumentBuilder.swift
```diff
@@ -1,6 +1,7 @@
 import Foundation
 import UIKit
 @_implementationOnly import MapboxCommon_Private
+import Turf
```
neelmistry94, line 4: nit: Reorder and switch lines 3/4

Sources/MapboxMaps/Style/Types/ExpressionArgumentBuilder.swift
1ec5: If Turf fails to round-trip a GeoJSONObject for some reason, that would be an internal error instead of programmer error, so throwing an assertion would probably be more appropriate than throwing an error anyways.

Sources/MapboxMaps/Style/Types/Expression.swift
```diff
@@ -1,4 +1,5 @@
 import Foundation
+import Turf
```
macdrevx, line 2: We reexport Turf elsewhere which makes it unnecessary to `import Turf` in each file.

Sources/MapboxMaps/Style/Types/Expression.swift
```diff
@@ -171,6 +175,8 @@ public struct Expression: Codable, CustomStringConvertible, Equatable {
                 return lhsBool == rhsBool
             case (.option(let lhsOption), .option(let rhsOption)):
                 return lhsOption == rhsOption
+            case (.geoJSONObject(let lhsObject), .geoJSONObject(let rhsObject)):
+                return lhsObject == rhsObject
```
macdrevx, line 179: This entire == implementation could be synthesized. Feel free to delete it.

Sources/MapboxMaps/Style/Types/ExpressionArgumentBuilder.swift
```diff
@@ -90,3 +91,16 @@ extension UIColor: ExpressionArgumentConvertible {
...
+extension GeoJSONObject: ExpressionArgumentConvertible {
+    public var expressionArguments: [Expression.Argument] {
+        do {
+            let data = try JSONEncoder().encode(self)
+            let expression = try JSONDecoder().decode(Expression.Argument.self, from: data)
+            return [expression]
+        } catch {
+            // Throwable getters are new to Swift 5.5, so we’ll have to live with a fatal error for now.
+            fatalError("Unable to convert GeoJSON object to expression argument: \(error)")
+        }
```
macdrevx, line 104: I think this can just be:
```suggestion
        return [.geoJSONObject(self)]
```

Tests/MapboxMapsTests/Style/ExpressionTests/ExpressionTests.swift
```diff
@@ -97,4 +99,45 @@ final class ExpressionTests: XCTestCase {
...
+        colorado.identifier = "CO"
+        colorado.properties = [
+            "population": 5_773_714,
+        ]
+
+        let withinExpression = Exp(.within) {
+            GeoJSONObject.feature(colorado)
+        }
+        XCTAssertNotNil(withinExpression.arguments.first)
+        guard let withinObject = withinExpression.arguments.first else { return }
```
macdrevx, line 121:
```suggestion
        let withinObject = try XCTUnwrap(withinExpression.arguments.first)
```

Sources/MapboxMaps/Style/Types/Expression.swift
```diff
@@ -198,6 +204,8 @@ public struct Expression: Codable, CustomStringConvertible, Equatable {
                 try container.encode(boolean)
             case .option(let option):
                 try container.encode(option)
+            case .geoJSONObject(let object):
```
macdrevx, line 207: What if we just named it `geoJSON`? The word "object" always makes me think I'm dealing with an instance of a class.

macdrevx: this is feedback for Turf too

Tests/MapboxMapsTests/Style/ExpressionTests/ExpressionTests.swift
macdrevx: Also, I think I'd rename the local variable to `firstArgument`. The fact that we're dealing with a "within" expression is clear enough from the context.

Tests/MapboxMapsTests/Style/ExpressionTests/ExpressionTests.swift
```diff
@@ -97,4 +99,45 @@ final class ExpressionTests: XCTestCase {
...
+        XCTAssertEqual(withinObject, .geoJSONObject(.feature(colorado)))
+
+        var withinData: Data?
+        XCTAssertNoThrow(withinData = try JSONEncoder().encode(withinExpression))
```
macdrevx, line 125: It'd simplify the test to remove `XCTAssertNoThrow` and let the rest of the test get skipped if there's a failure here.

macdrevx (changes requested):
This is looking good. I'd like to review again after any discussion & updates.

macdrevx (approved):

PR #715, opened 2021-09-28
Migrate Turf GeoJSON usage to GeoJSONObject and JSONValue

1ec5:
This PR integrates the GeoJSON API changes in mapbox/turf-swift#154 and [Turf v2.0.0-rc.2](https://github.com/mapbox/turf-swift/releases/tag/v2.0.0-rc.2). Most of the changes are in generated code and in tests, though some examples have also changed in minor ways. There is one breaking change on top of the Turf changes: the generic parameter on the `Style.updateGeoJSONSource<T: GeoJSONObject>(withId:geoJSON:)` method has been removed. Instead of passing in the expected GeoJSON object type, you perform pattern matching on the return value using `case let`. (#715)

Depends on mapbox/turf-swift#154.

/cc @mapbox/maps-ios

PR #741, opened 2021-10-06
Turf v2.0.0

1ec5:
Updated SPM and CocoaPods manifests for MapboxMaps to require Turf v2._x_. [Turf v2.0.0](https://github.com/mapbox/turf-swift/releases/tag/v2.0.0) comes with a number of backwards-incompatible changes compared to either v1._x_ or v2.0.0-rc.1 related to GeoJSON types, as seen in #715. However, there have been no backwards-incompatible changes since v2.0.0-rc.2, so this PR is solely about updating the dependency files.

/cc @mapbox/maps-ios

### DIFF
--- a/MapboxMaps.podspec
+++ b/MapboxMaps.podspec
@@ -24,6 +24,6 @@ Pod::Spec.new do |m|
   m.dependency 'MapboxCoreMaps', '10.0.0'
   m.dependency 'MapboxCommon', '20.0.0'
   m.dependency 'MapboxMobileEvents', '1.0.5'
-  m.dependency 'Turf', '2.0.0-rc.2'
+  m.dependency 'Turf', '~> 2.0'
 
 end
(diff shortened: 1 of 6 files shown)


MapboxMaps.podspec
```diff
@@ -24,6 +24,6 @@ Pod::Spec.new do |m|
   m.dependency 'MapboxCoreMaps', '10.0.0'
   m.dependency 'MapboxCommon', '20.0.0'
   m.dependency 'MapboxMobileEvents', '1.0.5'
-  m.dependency 'Turf', '2.0.0-rc.2'
+  m.dependency 'Turf', '~> 2.0'
```
1ec5, line 27: The patch version is omitted for consistency with SPM’s `Package.Dependency.package(name:url:from:)`.

macdrevx:
Need to update `scripts/release/packager/versions.json`

macdrevx (approved):

PR #756, opened 2021-10-13
Export only required MapboxCoreMaps members

endanke:

Fixes: https://github.com/mapbox/mapbox-maps-internal/issues/1144

## Pull request checklist:

### Summary of changes

Exporting all members of MapboxCoreMaps lead to some name collision with some default Swift types. (For example MapboxCoreMaps.Task is resolved by default instead of Swift.Task)
Only exporting the required members reduces the chance of unexpected collisions, but doesn't fully solve it in all cases. (Eg.: Event, Map, and Size still needs to be exported, which could collide with default class names.)

### DIFF
--- a/Tests/MapboxMapsTests/Foundation/Camera/BasicCameraAnimatorTests.swift
+++ b/Tests/MapboxMapsTests/Foundation/Camera/BasicCameraAnimatorTests.swift
@@ -1,4 +1,5 @@
 import XCTest
+import CoreLocation
 @testable import MapboxMaps
 
 let cameraOptionsTestValue = CameraOptions(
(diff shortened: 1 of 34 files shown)


endanke:
@macdrevx @alexshalamov 
There seems to be another issue, a warning related to the new imports when archiving MapboxMaps: 
`scoped imports are not yet supported in module interfaces`
And because the warnings are treated as errors in release mode, the CI can't build the binary.

As far as I understand this warning means that the users of the SDK wouldn't be able to inspect the definitions in the `MapboxCoreMaps` and `MapboxCommon` members, although it would be still possible to build and use it.

endanke:
Closing as it will be fixed by: https://github.com/mapbox/mapbox-gl-native-internal/pull/2367
One-by-one exporting is not possible due to the warning mentioned above: https://github.com/mapbox/mapbox-maps-ios/pull/756#issuecomment-943228474

PR #776, opened 2021-10-20
View Annotations

endanke:
This PR adds the view annotations API to the SDK.
View annotations enables customers to add arbitrary UIView subclasses to the map, which are automatically positioned at a given coordinate or at a specified feature's location on a symbol layer.

Simulator Screen Shot - iPhone 11 Pro - 2021-11-16 at 16 49 02

## Pull request checklist:
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-ios` changelog: `<changelog>Add view annotations feature, which enables the usage of custom UIView subclasses as annotations</changelog>`.


Apps/DebugApp/DebugApp/DebugViewController.swift
```diff
@@ -7,14 +7,168 @@ import MapboxMaps
...
+    let sanfrancisco = CLLocationCoordinate2D(latitude: 37.7749, longitude: -122.4194)
+    let boston = CLLocationCoordinate2D(latitude: 42.3601, longitude: -71.0589)
+    let nullIsland = CLLocationCoordinate2D(latitude: 0, longitude: 0)
+    let calloutViewHeight = 40
+    let calloutViewWidth = 60
+
+    func setupAnnotations() {
+
+        let frame = CGRect(origin: .zero, size: .init(width: calloutViewWidth, height: calloutViewHeight))
+        mapView.viewAnnotations.addViewAnnotation(
```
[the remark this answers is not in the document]
endanke: Yeah this DebugViewController is actually not used, I'll remove the changes from it. But `viewAnnotationManager` is the actual name of the variable, you can find it here:
https://github.com/mapbox/mapbox-maps-ios/pull/776/files#diff-5be8ad5aa3482beb6c6ec989ea84a88e38d60e6a294963b0051c9b9f3eddedbeR295

Sources/MapboxMaps/Annotations/ViewAnnotation.swift
```diff
@@ -0,0 +1,117 @@
+import UIKit
+@_implementationOnly import MapboxCommon_Private
+@_implementationOnly import MapboxCoreMaps_Private
+
+public protocol ViewAnnotationManager: AnyObject {
+    // TODO: Add documentation
+    func addViewAnnotation(_ annotatonView: UIView, _ options: ViewAnnotationOptions)
```
kiryldz, line 7: in order not to forget from our discussion today: rename param here and below to `view` for consistency with Android and in general it seems to fit better.

Sources/MapboxMaps/Annotations/ViewAnnotation.swift
```diff
@@ -0,0 +1,206 @@
+import UIKit
+@_implementationOnly import MapboxCommon_Private
+@_implementationOnly import MapboxCoreMaps_Private
+
+public protocol ViewAnnotationManager: AnyObject {
```
macdrevx, line 5: This needs to be internal so that we can modify it freely. On iOS, we only make protocols public when a feature requires developers to implement the protocol.

Sources/MapboxMaps/Foundation/MapboxMap.swift
```diff
@@ -835,6 +841,59 @@ extension MapboxMap {
...
+        // TODO: error handling
+        let result = __map.getViewAnnotationOptions(forIdentifier: id)
+        guard !result.isError() else {
+            fatalError()
+        }
+        guard let options = result.value as? MapboxCoreMaps.ViewAnnotationOptions else {
+            fatalError()
+        }
+        return ViewAnnotationOptions(options)
+    }
```
macdrevx, line 895: Are we 100% sure we want to make all of these public?

Sources/MapboxMaps/Foundation/MapboxMap.swift
```diff
@@ -835,6 +841,59 @@ extension MapboxMap {
...
     }
+
+}
+
+// MARK: - View Annotations
+
+extension MapboxMap {
+
+    // TODO: Add documentation
+    public func setViewAnnotationPositionsUpdateListenerFor(listener: ViewAnnotationPositionsListener) {
```
macdrevx, line 852:
```suggestion
    public func setViewAnnotationPositionsUpdateListener(_ listener: ViewAnnotationPositionsListener) {
```

Sources/MapboxMaps/Foundation/MapboxMap.swift
```diff
@@ -835,6 +841,59 @@ extension MapboxMap {
...
+        __map.setViewAnnotationPositionsUpdateListenerFor(listener)
```
macdrevx, line 853: What's the resulting ownership graph look like? Does this create a cycle if `listener` owns a strong reference to `MapboxMap`? If so, how do we break that cycle in `ViewAnnotationManagerImpl`?

Sources/MapboxMaps/Foundation/MapboxMap.swift
macdrevx, line 852: The name suggests that there can only be 1 listener at a time. Is that true?

Sources/MapboxMaps/Foundation/MapboxMap.swift
```diff
@@ -835,6 +841,59 @@ extension MapboxMap {
...
+    }
+
+    /**
+     * Add view annotation.
+     *
+     * @return position for all views that need to be updated on the screen or null if views' placement remained the same.
+     */
+    public func addViewAnnotation(forId id: String, options: ViewAnnotationOptions) {
```
macdrevx, line 861: What happens if a view annotation with the same id has already been added?

Sources/MapboxMaps/Foundation/MapboxMap.swift
```diff
@@ -835,6 +841,59 @@ extension MapboxMap {
...
+        __map.addViewAnnotation(forIdentifier: id, options: MapboxCoreMaps.ViewAnnotationOptions(options))
+    }
+
+    /**
+     * Update view annotation if it exists.
+     *
+     * @return position for all views that need to be updated on the screen or null if views' placement remained the same.
+     */
+    public func updateViewAnnotation(forId id: String, options: ViewAnnotationOptions) {
```
macdrevx, line 870: What happens if no annotation exists with the specified id?

Sources/MapboxMaps/Foundation/MapboxMap.swift
```diff
@@ -835,6 +841,59 @@ extension MapboxMap {
...
+     *
+     * @return position for all views that need to be updated on the screen or null if views' placement remained the same.
+     */
+    public func removeViewAnnotation(forId id: String) {
+        __map.removeViewAnnotation(forIdentifier: id)
+    }
+
+    // TODO: Add documentation
+    // Make this throw, return ViewAnnotationOptions directly
+    public func getViewAnnotationOptions(forId id: String) throws -> ViewAnnotationOptions {
```
macdrevx, line 885:
```suggestion
    public func options(forViewAnnotationWithId id: String) throws -> ViewAnnotationOptions {
```

Sources/MapboxMaps/Foundation/MapboxMap.swift
```diff
@@ -835,6 +841,59 @@ extension MapboxMap {
...
+    public func updateViewAnnotation(forId id: String, options: ViewAnnotationOptions) {
+        __map.updateViewAnnotation(forIdentifier: id, options: MapboxCoreMaps.ViewAnnotationOptions(options))
+    }
+
+    /**
+     * Remove view annotation if it exists.
+     *
+     * @return position for all views that need to be updated on the screen or null if views' placement remained the same.
+     */
+    public func removeViewAnnotation(forId id: String) {
```
macdrevx, line 879:
```suggestion
    public func removeViewAnnotation(withId id: String) {
```

Sources/MapboxMaps/Foundation/MapboxMap.swift
macdrevx, line 870:
```suggestion
    public func updateViewAnnotation(withId id: String, options: ViewAnnotationOptions) {
```

Sources/MapboxMaps/Foundation/MapboxMap.swift
macdrevx, line 861:
```suggestion
    public func addViewAnnotation(withId id: String, options: ViewAnnotationOptions) {
```

Sources/MapboxMaps/Foundation/MapView.swift
```diff
@@ -36,6 +36,9 @@ open class MapView: UIView {
     /// Controls the addition/removal of annotations to the map.
     public internal(set) var annotations: AnnotationOrchestrator!
 
+    // TODO: the naming might be slightly confusing, as the annotations object already covers most cases. Need to check if they should be merged or not. Currently their implementation is pretty different.
+    public internal(set) var viewAnnotationManager: ViewAnnotationManager!
```
macdrevx, line 40:
```suggestion
    public internal(set) var viewAnnotations: ViewAnnotationManager!
```

Sources/MapboxMaps/Foundation/MapView.swift
```diff
@@ -36,6 +36,9 @@ open class MapView: UIView {
     /// Controls the addition/removal of annotations to the map.
     public internal(set) var annotations: AnnotationOrchestrator!
 
+    // TODO: the naming might be slightly confusing, as the annotations object already covers most cases. Need to check if they should be merged or not. Currently their implementation is pretty different.
```
macdrevx, line 39: I would love to see the interface be merged. Implementations could still be separated by using a wrapper.

Sources/MapboxMaps/Foundation/MapView.swift
```diff
@@ -286,11 +289,10 @@ open class MapView: UIView {
         location = LocationManager(style: mapboxMap.style)
 
         // Initialize/Configure annotations orchestrator
-        annotations = AnnotationOrchestrator(
-            gestureRecognizer: gestures.singleTapGestureRecognizer,
-            mapFeatureQueryable: mapboxMap,
-            style: mapboxMap.style,
-            displayLinkCoordinator: self)
+        annotations = AnnotationOrchestrator(gestureRecognizer: gestures.singleTapGestureRecognizer, mapFeatureQueryable: mapboxMap, style: mapboxMap.style, displayLinkCoordinator: self)
```
macdrevx, line 292: Why the unwrapping?

Sources/MapboxMaps/Annotations/ViewAnnotationOptions.swift
```diff
@@ -0,0 +1,89 @@
...
+    public var associatedFeatureId: String?
+    public var allowOverlap: Bool?
+    public var visible: Bool?
+    public var anchor: ViewAnnotationAnchor?
+    public var offsetX: CGFloat?
+    public var offsetY: CGFloat?
+    public var selected: Bool?
+
+    // Used for hashing and constructing objc value
+    fileprivate var mbxGeometry: MapboxCommon.Geometry?
```
macdrevx, line 21: I don't think we should do this. `MapboxCommon.Geometry` doesn't implement a useful `-isEqual:` method, so we'd also need to write out `==` explicitly to exclude this stored property. Let's just omit `geometry` from the hash.

Sources/MapboxMaps/Annotations/ViewAnnotationOptions.swift
```diff
@@ -0,0 +1,89 @@
...
+        self.anchor = anchor
+        self.offsetX = offsetX
+        self.offsetY = offsetY
+        self.selected = selected
+        self.mbxGeometry = geometry.map({ MapboxCommon.Geometry.init($0) }) ?? nil
+    }
+
+    internal init(_ objcValue: MapboxCoreMaps.ViewAnnotationOptions) {
+        self.init(
+            geometry: objcValue.__geometry.map({ Geometry($0) }) ?? nil,
```
macdrevx, line 49:
```suggestion
            geometry: objcValue.__geometry.map(Geometry.init),
```

Sources/MapboxMaps/Annotations/ViewAnnotationOptions.swift
```diff
@@ -0,0 +1,89 @@
...
+            width: objcValue.__width?.CGFloat,
+            height: objcValue.__height?.CGFloat,
+            associatedFeatureId: objcValue.__associatedFeatureId,
+            allowOverlap: objcValue.__allowOverlap?.boolValue,
+            visible: objcValue.__visible?.boolValue,
+            anchor: objcValue.__anchor.map({ ViewAnnotationAnchor(rawValue: $0.intValue) }) ?? nil,
```
macdrevx, line 55:
```suggestion
            anchor: objcValue.__anchor.map { ViewAnnotationAnchor(rawValue: $0.intValue) },
```

Sources/MapboxMaps/Annotations/ViewAnnotation.swift
```diff
@@ -0,0 +1,206 @@
...
+
+}
+
+extension ViewAnnotationManagerImpl: ViewAnnotationPositionsListener {
+
+    public func onViewAnnotationPositionsUpdate(forPositions positions: [ViewAnnotationPositionDescriptor]) {
+        placeAnnotations(positions: positions)
+    }
+
+}
```
macdrevx, line 149: When we make `ViewAnnotationManagerImpl` public, let's avoid making this method public by following the delegation pattern used in a few other places like `DelegatingMapClient`

Sources/MapboxMaps/Annotations/ViewAnnotation.swift
```diff
@@ -0,0 +1,206 @@
...
+            viewAnnotationWrapper.rightAnchor.constraint(equalTo: view.rightAnchor)
+        ])
+
+        self.mapboxMap = mapboxMap
+        mapboxMap.setViewAnnotationPositionsUpdateListenerFor(listener: self)
+    }
+
+    // MARK: - ViewAnnotationManagerInterface protocol functions
+
+    public func addViewAnnotation(_ view: UIView, _ options: ViewAnnotationOptions) throws -> AnnotationView {
```
macdrevx, line 43: Could we flip the terminology to `annotationView`? I think it would make all these methods read more fluidly.

macdrevx:
```suggestion
    public func addAnnotationView(withContentView view: UIView, _ options: ViewAnnotationOptions) throws -> AnnotationView {
```

macdrevx: Maybe something like that

Sources/MapboxMaps/Annotations/ViewAnnotation.swift
```diff
@@ -0,0 +1,206 @@
...
+        }
+        let annotatonView = AnnotationView(view: view, annotationManager: self)
+        annotationViewsById[annotatonView.id] = annotatonView
+        annotationIdByViews[annotatonView] = annotatonView.id
+        viewAnnotationWrapper.addSubview(annotatonView)
+        mapboxMap.addViewAnnotation(forId: annotatonView.id, options: options)
+        return annotatonView
+    }
+
+    public func removeViewAnnotation(_ annotatonView: AnnotationView) -> Bool {
```
macdrevx, line 55:
```suggestion
    public func remove(_ annotatonView: AnnotationView) -> Bool {
```

Based on the Swift API naming guidelines, I think we could shorten this ^

Sources/MapboxMaps/Annotations/ViewAnnotation.swift
```diff
@@ -0,0 +1,206 @@
...
+            return false
+        }
+        mapboxMap.removeViewAnnotation(forId: id)
+        annotatonView.removeFromSuperview()
+        annotationViewsById.removeValue(forKey: id)
+        annotationIdByViews.removeValue(forKey: annotatonView)
+        return true
+    }
+
+    public func updateViewAnnotation(_ annotatonView: AnnotationView, _ options: ViewAnnotationOptions) -> Bool {
```
macdrevx, line 66:
```suggestion
    public func update(_ annotatonView: AnnotationView, options: ViewAnnotationOptions) -> Bool {
```

Sources/MapboxMaps/Annotations/ViewAnnotation.swift
macdrevx, line 55: These methods return Bool to indicate success/failure. Should the return value be discardable? `@discardableResult`

Sources/MapboxMaps/Annotations/ViewAnnotation.swift
macdrevx, line 43: There should be a label for the second argument. `options` would be fine.

macdrevx (changes requested):
Pausing my review here. Let's find some time to talk through the feedback & come up with some next steps.

Sources/MapboxMaps/Annotations/ViewAnnotation.swift
endanke: Will change it, originally I based it on the Android implementation but I agree it's more suitable that way for iOS.

Sources/MapboxMaps/Foundation/MapboxMap.swift
endanke: This and the other functions here are named like this to match the underlying implementation. Should we rename those too? I also changed their visibility now to internal, so I wonder if we could just ignore it.

Sources/MapboxMaps/Foundation/MapView.swift
endanke: No reason, just haven't cleaned up some earlier changes yet.

Sources/MapboxMaps/Annotations/ViewAnnotationOptions.swift
endanke: That's a good point, but the reason why I was thinking to include it is because currently the geometry field is required when adding a view annotation, but the rest can be omitted. So I expect that most users will just use a `ViewAnnotationOptions` with only a geometry value, maybe it would be useful to consider it in the hash.
(But I'll remove it from now just for simplicity.)

Sources/MapboxMaps/Foundation/MapboxMap.swift
endanke: Only the first one will be added. (Currently there's no error handling on GL-Native side, but it simply ignores the call if the ID is already used.)

Sources/MapboxMaps/Foundation/MapboxMap.swift
endanke: The call is simply ignored at the moment.

Sources/MapboxMaps/Foundation/MapboxMap.swift
endanke: Yes, and it's one more reason to make this function internal. 👌

Sources/MapboxMaps/Foundation/MapboxMap.swift
endanke: No, I've just forgot to change them to internal.

Sources/MapboxMaps/Annotations/ViewAnnotationOptions.swift
endanke: This doesn't compile for me without the `?? nil` fallback. Or did you mean something else?

Sources/MapboxMaps/Annotations/ViewAnnotationOptions.swift
endanke: Same as below: https://github.com/mapbox/mapbox-maps-ios/pull/776#discussion_r745529814

Sources/MapboxMaps/Foundation/MapboxMap.swift
endanke: (Same question applies to all other naming change suggestions related to these API functions.)

Sources/MapboxMaps/Annotations/ViewAnnotation.swift
endanke: I wonder if we need a protocol at all?

Sources/MapboxMaps/Annotations/ViewAnnotation.swift
endanke: Thanks for the tip, I've added an implementation based on that. Although the name of the class might be a bit weird: https://github.com/mapbox/mapbox-maps-ios/pull/776/files#diff-6be4c15b3e0af24102f7ee6ab1a8187765ac65a9c83b8f314070d6eec3be7791R8

Sources/MapboxMaps/Foundation/MapView.swift
endanke: So the access would be something like `annotations.viewAnnotations` ? 
One other reason I left it separated, is because on Android the rest of the annotations are managed through a plugin, so there the view annotations will be accessed separately from them. Although I don't mind placing them under the orchestrator in our case.

Sources/MapboxMaps/Annotations/ViewAnnotation.swift
```diff
@@ -0,0 +1,206 @@
...
+
+internal class ViewAnnotationManagerImpl: ViewAnnotationManager {
+
+    private let mapboxMap: MapboxMapProtocol
+    private let viewAnnotationWrapper = UIView(frame: .zero)
+    private var annotationViewsById: [String: AnnotationView] = [:]
+    private var annotationIdByViews: [AnnotationView: String] = [:]
+
+    internal init(view: UIView, mapboxMap: MapboxMapProtocol) {
+        view.insertSubview(viewAnnotationWrapper, at: 1)
```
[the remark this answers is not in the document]
endanke: Good idea, I've added a class named as `SubviewInteractionOnlyView` and I think it can be used for the `AnnotationView` wrapper class too.

Sources/MapboxMaps/Annotations/ViewAnnotation.swift
macdrevx: If the protocol is unused, please remove it. The main use case for protocols like this is to facilitate mocking in the unit tests.

Sources/MapboxMaps/Foundation/MapboxMap.swift
macdrevx: The "for" in the MapboxCoreMaps method names are produced by bindgen, which isn't yet sophisticated enough to make good choices when it comes to English grammar. Don't worry about renaming those, but let's try to make our wrappers read better.

> I also changed their visibility now to internal, so I wonder if we could just ignore it.

If they're internal, it's definitely not as important to get right the first time since we can always change it later. New public APIs get a higher level of scrutiny.

Sources/MapboxMaps/Foundation/MapView.swift
macdrevx: If we do it at all, it'd be nice to achieve some kind of consistency with the existing annotations APIs. So maybe something like `annotations.viewAnnotationManager` to mimic the existing factory methods for the other annotation managers.

That said, it's more important to be consistent with Android, so we would only do this on iOS if we also do it there.

Sources/MapboxMaps/Annotations/ViewAnnotationOptions.swift
macdrevx: I agree with this decision. I'll make a note about adding geometry into the hash in the future.

Sources/MapboxMaps/Annotations/ViewAnnotationOptions.swift
macdrevx: Let's look at this together. It's not easy for me to test this out locally in a time-efficient manner since the GL Native changes are still in flight.

Sources/MapboxMaps/Annotations/ViewAnnotation.swift
macdrevx: Name seems good to me

Sources/MapboxMaps/Foundation/MapboxMap.swift
endanke: Good question, I think this could have been a retain cycle, but after the review change we only pass `DelegatingViewAnnotationPositionsListener` to it which only has a weak reference to the annotation manager. I tested it and it seems it's not leaking.
Also I've added an extra call to the destructor of the manager the clear the listener after we don't need it anymore.

Sources/MapboxMaps/Annotations/ViewAnnotationOptions.swift
```diff
@@ -0,0 +1,80 @@
+import Foundation
+import UIKit
+
+public struct ViewAnnotationOptions: Hashable {
```
endanke, line 4: @macdrevx do we have to add documentation to this class or can we just have a link pointing to `MapboxCoreMaps.ViewAnnotationOptions`?

I see that in case of the `CameraOptions` only the constructor is documented, but even there I think it would be cleaner to point to the description of the parameters without duplicating them in the constructor's documentation.

macdrevx: We should fully document these public APIs. Links to MapboxCoreMaps documentation doesn't work well currently, so we should assume folks will only read these docs. The docs for each public interface should largely be able to stand on their own to minimize how much jumping around folks need to do when using the documentation.

endanke: Alright, I'll add them. Thanks for the explanation!

Sources/MapboxMaps/Annotations/ViewAnnotation.swift
```diff
@@ -0,0 +1,232 @@
+import UIKit
+@_implementationOnly import MapboxCommon_Private
+@_implementationOnly import MapboxCoreMaps_Private
+
+public enum ViewAnnotationManagerError: Error {
+    case insertionFailure(reason: String)
+    case updateFailure(reason: String)
+    case removalFailure(reason: String)
+    case optionsQueryFailure(reason: String)
```
endanke, line 9: Not sure if we need to have an error case for the options queries.
Currently the APIs which retrieve options are returning optionals, which might be enough indication for the status. This is also thrown if we fail to unwrap the value that's coming from gl-native, but that shouldn't happen if the response was not an error.

Sources/MapboxMaps/Annotations/ViewAnnotation.swift
```diff
@@ -0,0 +1,237 @@
...
+        containerView.addSubview(annotatonView)
+        return annotatonView
+    }
+
+    /// Remove given `AnnotationView` from the map if it was present.
+    ///
+    /// - Throws:
+    ///   - `ViewAnnotationManagerError.annotationNotFound`: the supplioed view was not found
+    ///   - `ViewAnnotationManagerError.removalFailure`: errors during the removal of the view
+    public func remove(_ annotatonView: AnnotationView) throws {
```
endanke, line 83: I've removed the boolean return value from this function and from `update` because we have to throw errors from the GL-Native side and I thought it might be more convenient for the user to handle the "not found" case the same way as other issues. (I can of course put back the return value if needed, no strong opinions on that.)

Sources/MapboxMaps/Foundation/MapView.swift
```diff
@@ -67,6 +70,9 @@ open class MapView: UIView {
 
     private let cameraViewContainerView = UIView()
 
+    /// Holds ViewAnnotation views
+    private let viewAnnotationContainerView = SubviewInteractionOnlyView()
```
endanke, line 74: I was considering to only add this view to the map when the `ViewAnnotationManager` is used. However it might not be a big performance improvement and would probably just complicate the setup, so I think it's easier to just leave it as it is.

Sources/MapboxMaps/Annotations/ViewAnnotation.swift
```diff
@@ -0,0 +1,240 @@
...
+    internal func validateAnnotation(byAnnotationId id: String) {
+        guard let annotationView = annotationViewsById[id] else { return }
+        // If the user explicitly removed the ViewAnnotation or it's wrapped view
+        // we need to remove it from our layout calculation
+        if annotationView.subviews.isEmpty || annotationView.superview == nil {
+            try? remove(annotationView)
+        }
+        // View is still considered for layout calculation, users should not modify the visibility of the wrapped view
+        if let wrappedView = annotationView.subviews.first, wrappedView.isHidden {
+            Log.warning(forMessage: "Visibility changed for wrapped view", category: "Annotations")
```
endanke, line 173: This warning message might need a bit more details. 

Also I was thinking to add a flag to disable this warning, which would allow the users to hide the wrapped views while they're still considered in the layout calculation. (Although the same could be achieved by setting the opacity to zero, so it might not be relevant.)

kiryldz (approved):

Sources/MapboxMaps/Foundation/MapView.swift
```diff
@@ -249,6 +255,15 @@ open class MapView: UIView {
             mapboxMap.setCamera(to: cameraOptions)
         }
 
+        insertSubview(viewAnnotationContainerView, at: 1)
```
macdrevx, line 258: Would be less brittle to insert the subview relative to some other view (e.g. above the metalView)

Sources/MapboxMaps/Foundation/MapboxMap.swift
```diff
@@ -835,6 +841,49 @@ extension MapboxMap {
...
+    }
+
+    internal func options(forViewAnnotationWithId id: String) throws -> ViewAnnotationOptions {
+        let expected = __map.getViewAnnotationOptions(forIdentifier: id)
+        if expected.isError(), let reason = expected.error as? String {
+            throw ViewAnnotationManagerError.optionsQueryFailure(reason: reason)
+        }
+        guard let options = expected.value as? MapboxCoreMaps.ViewAnnotationOptions else {
+            throw ViewAnnotationManagerError.optionsQueryFailure(reason: "Failed to unwrap ViewAnnotationOptions")
+        }
```
macdrevx, line 883: If the value isn't of the expected type, we should just crash. This should never happen.

Sources/MapboxMaps/Annotations/ViewAnnotationOptions.swift
```diff
@@ -0,0 +1,117 @@
...
+        hasher.combine(visible)
+        hasher.combine(anchor)
+        hasher.combine(offsetX)
+        hasher.combine(offsetY)
+        hasher.combine(selected)
+    }
+}
+
+extension MapboxCoreMaps.ViewAnnotationOptions {
+    public convenience init(_ swiftValue: ViewAnnotationOptions) {
```
macdrevx, line 105: Should be internal

Sources/MapboxMaps/Foundation/SubviewInteractionOnlyView.swift
```diff
@@ -0,0 +1,11 @@
+import UIKit
+
+/// This class is a wrapper view which forwards all touch events to it's subviews
+public class SubviewInteractionOnlyView: UIView {
```
macdrevx, line 4: This is public so that AnnotationView can inherit from it. I'd rather have AnnotationView inherit directly from UIView, keep this class internal, and duplicate the hitTest logic into AnnotationView.

Exposing this class as public leaks too much about the specific implementation we're going with right now and will make it more difficult if we want to take a different path in the future.

Sources/MapboxMaps/Annotations/ViewAnnotation.swift
```diff
@@ -0,0 +1,240 @@
...
+
+extension ViewAnnotationManager: DelegatingViewAnnotationPositionsUpdateListenerDelegate {
+
+    internal func onViewAnnotationPositionsUpdate(forPositions positions: [ViewAnnotationPositionDescriptor]) {
+        placeAnnotations(positions: positions)
+    }
+
+}
+
+public final class AnnotationView: SubviewInteractionOnlyView {
```
macdrevx, line 187: Should be moved to a separate file.

Sources/MapboxMaps/Annotations/ViewAnnotation.swift
```diff
@@ -0,0 +1,240 @@
...
+
+    private static var currentId = 0
+    internal let id: String = {
+        let id = String(currentId)
+        currentId += 1
+        return id
+    }()
```
macdrevx, line 194: Let's try to avoid using static storage. A simple solution would be for ViewAnnotationManager to have an instance property to keep track of `currentId` and pass that into AnnotationView's init.

Sources/MapboxMaps/Annotations/ViewAnnotation.swift
```diff
@@ -0,0 +1,240 @@
...
+    ///
+    /// - Returns: `ViewAnnotationOptions` if view was found and `nil` otherwise.
+    public func options(byFeatureId identifier: String) -> ViewAnnotationOptions? {
+        return viewAnnotation(byFeatureId: identifier).flatMap { options(byAnnotationView: $0) }
+    }
+
+    /// Get current `ViewAnnotationOptions` for given `AnnotationView`.
+    ///
+    /// - Returns: `ViewAnnotationOptions` if view was found and `nil` otherwise.
+    public func options(byAnnotationView view: AnnotationView) -> ViewAnnotationOptions? {
```
macdrevx, line 129:
```suggestion
    public func options(for view: AnnotationView) -> ViewAnnotationOptions? {
```

Sources/MapboxMaps/Annotations/ViewAnnotation.swift
```diff
@@ -0,0 +1,240 @@
...
+            return nil
+        }
+        return annotationViewsById[id]
+    }
+
+    /// Find `ViewAnnotationOptions` of view annotation by feature id if it was specified as part of `ViewAnnotationOptions.associatedFeatureId`.
+    ///
+    /// - Returns: `ViewAnnotationOptions` if view was found and `nil` otherwise.
+    public func options(byFeatureId identifier: String) -> ViewAnnotationOptions? {
+        return viewAnnotation(byFeatureId: identifier).flatMap { options(byAnnotationView: $0) }
```
macdrevx, line 123:
```suggestion
    public func options(forFeatureId featureId: String) -> ViewAnnotationOptions? {
        return viewAnnotation(forFeatureId: featureId).flatMap(options(for:))
```

macdrevx:
Public API naming still needs work.

macdrevx:
Need to document parameters of public methods

Sources/MapboxMaps/Annotations/ViewAnnotation.swift
```diff
@@ -0,0 +1,240 @@
...
+///
+/// View annotations are not explicitly bound to any sources however `ViewAnnotationOptions.associatedFeatureId` could be
+/// used to bind given view annotation with some `Feature` by `Feature.identifier` meaning visibility of view annotation will be driven
+/// by visibility of given feature.
+public final class ViewAnnotationManager {
+
+    private let containerView: SubviewInteractionOnlyView
+    private let mapboxMap: MapboxMapProtocol
+    private var annotationViewsById: [String: AnnotationView] = [:]
+    private var annotationIdByViews: [AnnotationView: String] = [:]
```
macdrevx, line 29:
```suggestion
    private var annotationIdsByView: [AnnotationView: String] = [:]
```

Sources/MapboxMaps/Annotations/ViewAnnotation.swift
```diff
@@ -0,0 +1,240 @@
...
+    }
+
+    /// Remove given `AnnotationView` from the map if it was present.
+    ///
+    /// - Throws:
+    ///   - `ViewAnnotationManagerError.annotationNotFound`: the supplioed view was not found
+    ///   - `ViewAnnotationManagerError.removalFailure`: errors during the removal of the view
+    public func remove(_ annotatonView: AnnotationView) throws {
+        guard let id = annotationIdByViews[annotatonView], let annotatonView = annotationViewsById[id] else {
+            throw ViewAnnotationManagerError.annotationNotFound
```
macdrevx, line 85: Is it necessary to throw here? Why not just do nothing if the view isn't present?

Sources/MapboxMaps/Annotations/ViewAnnotation.swift
macdrevx: Agreed. Returning a Bool would be redundant

Sources/MapboxMaps/Annotations/ViewAnnotation.swift
```diff
@@ -0,0 +1,240 @@
...
+    /// Important thing to keep in mind that only properties present in `options` will be updated,
+    /// all other will remain the same as specified before.
+    ///
+    /// - Throws:
+    ///   - `ViewAnnotationManagerError.annotationNotFound`: the supplioed view was not found
+    ///   - `ViewAnnotationManagerError.updateFailure`: errors during the up date of the view (eg. incorrect parameters)
+    public func update(_ annotatonView: AnnotationView, options: ViewAnnotationOptions) throws {
+        guard let id = annotationIdByViews[annotatonView] else {
+            throw ViewAnnotationManagerError.annotationNotFound
+        }
```
macdrevx, line 103: Is this error handling redundant? Could we just let gl-native throw if the id isn't found?

Sources/MapboxMaps/Annotations/ViewAnnotation.swift
```diff
@@ -0,0 +1,240 @@
...
+    /// Find `AnnotationView` by feature id if it was specified as part of `ViewAnnotationOptions.associatedFeatureId`.
+    ///
+    /// - Returns: `AnnotationView` if view was found and `nil` otherwise.
+    public func viewAnnotation(byFeatureId identifier: String) -> AnnotationView? {
+        guard let id = annotationViewsById.keys.first(where: { id in
+            (try? mapboxMap.options(forViewAnnotationWithId: id).associatedFeatureId == identifier) ?? false
+        }) else {
+            return nil
+        }
+        return annotationViewsById[id]
```
macdrevx, line 116: Can simplify this implementation using flatMap

Sources/MapboxMaps/Annotations/ViewAnnotation.swift
endanke: Good point I'll change it. But also we have to make sure we don't use the same ID twice, because gl-native uses them to uniquely identify the managed view annotations. Since we will only have one ViewAnnotationManager per map, I think it's fine to just reset it now. But I'll add a comment about it.

macdrevx: Actually, looking at this more, it seems unnecessary to store the id on the AnnotationView at all

Sources/MapboxMaps/Annotations/ViewAnnotation.swift
```diff
@@ -0,0 +1,240 @@
...
+        annotationViewsById.removeValue(forKey: id)
+        annotationIdByViews.removeValue(forKey: annotatonView)
+    }
+
+    /// Update given `AnnotationView` with `ViewAnnotationOptions`.
+    /// Important thing to keep in mind that only properties present in `options` will be updated,
+    /// all other will remain the same as specified before.
+    ///
+    /// - Throws:
+    ///   - `ViewAnnotationManagerError.annotationNotFound`: the supplioed view was not found
```
macdrevx, line 98:
```suggestion
    ///   - `ViewAnnotationManagerError.annotationNotFound`: the supplied view was not found
```

Sources/MapboxMaps/Annotations/ViewAnnotation.swift
```diff
@@ -0,0 +1,240 @@
...
+    ///   - `ViewAnnotationManagerError.updateFailure`: errors during the up date of the view (eg. incorrect parameters)
```
macdrevx, line 99:
```suggestion
    ///   - `ViewAnnotationManagerError.updateFailure`: errors during the update of the view (eg. incorrect parameters)
```

Sources/MapboxMaps/Annotations/ViewAnnotation.swift
```diff
@@ -0,0 +1,240 @@
...
+        didSet {
+            guard !ignoreUserEvents else { return }
+            guard let manager = annotationManager else { return }
+            let options = manager.options(byAnnotationView: self)
+            let visible = !isHidden
+            if visible != options?.visible {
+                try? manager.update(self, options: ViewAnnotationOptions(visible: visible))
+            }
+        }
+    }
```
macdrevx, line 209: This seems like the only behavior that warrants exposing `AnnotationView` as a public API. I am concerned that making `AnnotationView` public is too limiting to future iteration that we might want to do on this implementation.

Could we do some more exploration to see if there's another reasonable way to work around this issue? (and even confirm that it's worth working around given the complexity & maintenance commitment)?

endanke: One alternative I was considering is just to show a warning, like it's already implemented for the subview's visibility state in `validateAnnotation`.
The only issue with that is the user won't see it until they move the map around, and check the logs for warnings.
And we can also highlight in the docs that hiding the view won't remove them from the layout calculation.

Sources/MapboxMaps/Annotations/ViewAnnotation.swift
```diff
@@ -0,0 +1,240 @@
...
+
+        for position in positions {
+            validateAnnotation(byAnnotationId: position.identifier)
+            guard let annotationView = annotationViewsById[position.identifier] else {
+                continue
+            }
+            annotationView.translatesAutoresizingMaskIntoConstraints = true
+            annotationView.frame = CGRect(
+                origin: position.leftTopCoordinate.point,
+                size: CGSize(width: CGFloat(position.width), height: CGFloat(position.height))
```
macdrevx, line 149: We have a convenience constructor that converts a Size to a CGSize

macdrevx: See Size.swift

Sources/MapboxMaps/Annotations/ViewAnnotation.swift
```diff
@@ -0,0 +1,240 @@
...
+            annotationView.setInternalVisibility(isHidden: false)
+            visibleAnnotationIds.insert(position.identifier)
+        }
+
+        let annotationsToHide = Set<String>(annotationViewsById.keys).subtracting(visibleAnnotationIds)
+        for id in annotationsToHide {
+            validateAnnotation(byAnnotationId: id)
+            if let annotationView = annotationViewsById[id] {
+                annotationView.setInternalVisibility(isHidden: true)
+            }
```
macdrevx, line 160: Can use optional chaining here

macdrevx (approved):
I think this is good enough to ship in the beta. Let's merge it and address the feedback in a follow-up PR.

Tests/MapboxMapsTests/Foundation/Camera/MockMapboxMap.swift
```diff
@@ -46,8 +47,8 @@ final class MockMapboxMap: MapboxMapProtocol {
         var from: CGPoint
         var to: CGPoint
     }
-    let dragCameraOptionsStub = Stub<DragCameraOptionsParams, CameraOptions>(defaultReturnValue: CameraOptions())
-    func dragCameraOptions(from: CGPoint, to: CGPoint) -> CameraOptions {
+    let dragCameraOptionsStub = Stub<DragCameraOptionsParams, MapboxMaps.CameraOptions>(defaultReturnValue: CameraOptions())
+    func dragCameraOptions(from: CGPoint, to: CGPoint) -> MapboxMaps.CameraOptions {
```
macdrevx, line 51: Curious why these prefixes needed to be added

endanke: After I imported `MapboxCoreMaps_Private` it had some clashes.

macdrevx:
Would love to see more comprehensive unit tests. Let's assess what's needed and follow up in a separate PR.

Sources/MapboxMaps/Annotations/ViewAnnotation.swift
endanke: You mean for the `position`? It's not a `Size` instance, but a custom type that we get for these layout updates.

macdrevx (approved):

Sources/MapboxMaps/Annotations/ViewAnnotation.swift
macdrevx: ah, I see

PR #787, opened 2021-10-25
Use default values for wrapped MapProjection types

endanke:
This small PR simplifies the MapProjection usage further by using the default values in the MapProjection enum for the wrapped types.
It also makes it more similar to the Android implementation, where the underlying values are not explicitly initialized at the moment.

cc @macdrevx

### DIFF
--- a/Apps/Examples/Examples/All Examples/GlobeViewExample.swift
+++ b/Apps/Examples/Examples/All Examples/GlobeViewExample.swift
@@ -13,7 +13,7 @@ public class GlobeViewExample: UIViewController, ExampleProtocol {
     }
 
     internal var mapView: MapView!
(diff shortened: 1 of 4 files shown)


macdrevx:
A quick audit of our CI configs show that we're already using 12.5 when building XCFrameworks for distribution. With that in mind, I'd be fine with moving forward with this change after restoring the experimental designation to the structs.

endanke:
> A quick audit of our CI configs show that we're already using 12.5 when building XCFrameworks for distribution. With that in mind, I'd be fine with moving forward with this change after restoring the experimental designation to the structs.

Sounds good, I'll do it. At the same time I have to update the bots in the `mobile-metrics` repo because they're using 12.0 and originally their bots break this PR.

macdrevx (approved):

PR #808, opened 2021-11-03
Update CI for long-lived release branches

macdrevx:
## Pull request checklist:

### Summary of changes

Combines release-related CI workflows into a single workflow that runs when a release is tagged.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -312,6 +312,7 @@ jobs:
     <<: *base-job
     steps:
       - checkout
+      - add-mapbox-submodules-key
       - install-mbx-ci
       - configure-environment
       # Building and testing are split into 2, with the aim that we'll be able to reuse
@@ -343,6 +344,7 @@ jobs:
     <<: *base-job
     steps:
       - checkout
+      - add-mapbox-submodules-key
       - install-mbx-ci
       - configure-environment
       - ensure-text-exists-in-commit:
@@ -415,6 +417,7 @@ jobs:
     <<: *base-job
     steps:
       - checkout
+      - add-mapbox-submodules-key
       - install-mbx-ci
       - ensure-text-exists-in-commit:
           commit-text: "[run app device tests]"
@@ -486,6 +489,7 @@ jobs:
     <<: *base-job
     steps:
       - checkout
+      - add-mapbox-submodules-key
       - install-mbx-ci
       - configure-environment
       - ensure-text-exists-in-commit:
@@ -521,6 +525,7 @@ jobs:
           message: '<$CIRCLE_BUILD_URL|Release tag for \`v$VERSION\` started.>'
           include_visit_job_action: true
       - checkout
+      - add-mapbox-submodules-key
       - install-mbx-ci
       - configure-environment
       - install-dependencies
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -138,15 +138,10 @@ workflows:
...
-          filters:
-            branches:
-              only: /Release\/.*/
-      - ios-release-tag:
-          name: "iOS Release Tag & Docs"
+      - release:
+          name: "Release"
           filters:
             tags:
               only: /.*/
```
OdNairy, line 147: Should we run release job for v-tags only?

.circleci/config.yml
```diff
@@ -501,8 +496,8 @@ jobs:
     steps:
       - checkout
       - add_ssh_keys:
-           fingerprints:
-             - "3b:cd:47:bf:57:9c:e5:36:b0:4d:5f:12:5e:d3:b3:3e"
+          fingerprints:
+            - "3b:cd:47:bf:57:9c:e5:36:b0:4d:5f:12:5e:d3:b3:3e"
```
OdNairy, line 500: There are 5 times we reuse key fingerprint. Probably it would make sense to make a custom command with that.

.circleci/config.yml
```diff
@@ -521,29 +516,31 @@ jobs:
...
-      - ensure-text-exists-in-commit:
-          commit-text: "[release]"
-          always-run: << parameters.device-tests-always-run >>
       - run:
-          name: Extract Version From Commit
+          name: Verify that tag $CIRCLE_TAG adheres to the Semantic Versioning format
           command: |
-            GIT_COMMIT_MSG=`git log --format=oneline -n 1`
-            export VERSION=$(echo $GIT_COMMIT_MSG| cut -d' ' -f3)
-            echo "export VERSION=$VERSION" >> $BASH_ENV
```
OdNairy, line 543: Can we discuss the drop of the `$VERSION` env variable? It looks like a good idea to share actual `VERSION` with all nested processes. We just have to parse `CIRCLE_TAG` for a semver instead of git commit message. 

We had the same practice in Search SDK and was cool:
https://github.com/mapbox/mapbox-search-ios/blob/32028ae77fee0fbe0dadb5fdde9d1d3a4a2db512/.circleci/config.yml#L347-L356

OdNairy: Another reason I would prefer VERSION over CIRCLE_TAG is variable support in custom scripts. It's much more intuitive to pass `VERSION=10.1.0-rc2 ./run-me.sh` than passing a `CIRCLE_TAG` in the local environment (in case we have to call versioned scripts locally someday).

.circleci/config.yml
macdrevx: Yes. I think there's a step in this workflow that checks for that (and skips the job if the tag doesn't match), but maybe we can refactor it to just do that check here instead.

.circleci/config.yml
macdrevx: I'm open to this. I'll follow up with you directly.

.circleci/config.yml
OdNairy: My point is that's it better to not run a workflow if it wouldn't work. The only change we have to make is:
```suggestion
              only: /v.*/
```

macdrevx: Updated — I just transferred the existing regex we were using within the job

.circleci/config.yml
macdrevx: I just moved this into the install-mbx-ci command since that's what it's associated with (see https://github.com/mapbox/mapbox-maps-ios/pull/424)

OdNairy (approved):

Issue #704, opened 2021-09-22
Ambiguous methods when importing `Style` SPIs


devioustree:
## Environment

- Xcode version: 12.5.1
- iOS version:
- Devices affected:
- Maps SDK Version: 10.0.0-rc.9

## Observed behavior and steps to reproduce

`Style.swift` contains two definitions for `layerProperty(for:property:)`:

```swift
public func layerProperty(for layerId: String, property: String) -> Any
@_spi(Experimental) public func layerProperty(for layerId: String, property: String) -> StylePropertyValue
```

I wish to use the non-SPI method here. But if I need to use an unrelated SPI, both methods are exposed. Since the experimental SPI is more specific, the Swift compiler wants that to take precedence. This means the following code is ambiguous:

```swift
@_spi(Experimental) import MapboxMaps

let style: Style = ...
let type: String = style.layerProperty(for: "foo", property: "type") as! String
```

<img width="353" alt="Screenshot 2021-09-22 at 16 45 52" src="https://user-images.githubusercontent.com/38645/134436644-58d1a1e3-cff6-4fe2-aa6d-b8385deec180.png">

macdrevx:
Thanks for reporting this.

neelmistry94:
@devioustree This issue has been fixed in https://github.com/mapbox/mapbox-maps-ios/pull/708. Please let us know if there are any issues still.

devioustree:
Hah, `moveLayer` was the experimental method I needed and that's been marked stable!

neelmistry94:
@devioustree Hopefully that recent PR fixed a lot of issues and prevents further issues. Thanks for opening an issue and being a user of mapbox.

PR #832, opened 2021-11-11
Access @_spi protected functions from pre-compiled packages

e-001:
Generate private module interface in order to access `@spi_` protected functions from pre-compiled packages.

`@_spi(Experimental) import MapBox` fails to import protected functions if the package is not local to the project. [Certain examples](https://github.com/mapbox/mapbox-maps-ios/blob/main/Apps/Examples/Examples/All%20Examples/GlobeViewExample.swift) dependent on protected functions are unusable if the package is pre-compiled. 

Resolves warning on import:
```
'@_spi' import of 'MapboxMaps' will not include any SPI symbols; 'MapboxMaps' was built from the public interface at [redacted]/Build/Products/Debug-iphoneos/MapboxMaps.framework/Modules/MapboxMaps.swiftmodule/arm64-apple-ios.swiftinterface
```
## Pull request checklist:
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-ios` changelog: `<changelog>Generate private module interface in order to access protected @spi_(Experimental) functions from pre-compiled packages.</changelog>`.

### Summary of changes

Adds the `SWIFT_EMIT_PRIVATE_MODULE_INTERFACE` flag to the base config, which includes `@_spi` protected methods in the MapboxMaps swiftinterface.

Proactively included the `-verify-emitted-module-interface` Swift compiler flag, assuming this project would be affected by [these](https://bugs.swift.org/browse/SR-14195) [two](https://bugs.swift.org/browse/SR-898) Swift compiler bugs.

### Testing

Because the repo tests and example app import the local package, the only way to test this is by including it in an entirely separate project, which I have done by including this source branch as an SPM dependency. Fails on upstream, succeeds on head repo.

### User impact

See above. Should BOLO for bugs like #704.

### DIFF
--- a/Configurations/base.xcconfig
+++ b/Configurations/base.xcconfig
@@ -61,6 +61,7 @@ SWIFT_ACTIVE_COMPILATION_CONDITIONS[config=Release] = RELEASE
 
 SWIFT_OPTIMIZATION_LEVEL[config=Debug] = -Onone
 SWIFT_OPTIMIZATION_LEVEL[config=Release] = -O
+SWIFT_EMIT_PRIVATE_MODULE_INTERFACE = YES
 
 // -----------------------------------------------------------------------------
 // Clang


macdrevx:
I tested this out by…

```
cd scripts/release/packager
./package-mapbox-maps.sh dynamic
```

but I'm getting validation errors. @e-001 if you have time to debug, let me know what you find. Otherwise, please open an issue corresponding to this ticket so that we can track and prioritize internally.

macdrevx:
Also, it would probably be better to add these settings to scripts/release/packager/project.yml instead of to the base config so that it's targeted just to when we build the XCFrameworks. That yml file is a config file for [xcodegen](https://github.com/yonaskolb/XcodeGen).

e-001:
> Also, it would probably be better to add these settings to scripts/release/packager/project.yml instead of to the base config so that it's targeted just to when we build the XCFrameworks. That yml file is a config file for [xcodegen](https://github.com/yonaskolb/XcodeGen).

@macdrevx Hmm, this wouldn't solve my primary case (installed via package manager). Unless I'm mistaken, the only way to get an XCF of mapbox-ios is to build locally, which burns in your own MB API keys.

Considering the frequency and heft of the other closed XCFrameworks (mb-{common/core}), is there anything holding the whole repo from switching to precompiled distribution by default? Otherwise, I think it makes sense to open the API protection level to package-managed installs by default. I got here by scratching my head trying to get the globe examples running.

I'll look into the packaging script errors now.

macdrevx:
@e-001 let me take a step back here to make sure I understand what you're trying to do. I'd assumed you were using the direct download artifacts linked at the bottom of our [release pages](https://github.com/mapbox/mapbox-maps-ios/releases/tag/v10.1.0) like:

https://api.mapbox.com/downloads/v2/mobile-maps-ios/releases/ios/10.1.0/MapboxMaps.zip?access_token=<access-token>

The guidance I provided was aimed at updating the scripts we use when building the MapboxMaps.xcframework that's bundled with that download.

But it sounds like you have a different integration workflow. Could you provide a few more details on how it works?

e-001:
> Could you provide a few more details on how it works?

@macdrevx Of course.

Current workflow:
1. Add dependency via SPM or cocoapod.
2. Utilize `@_spi` protected functions, namely globe projections, via the fetched package.
3. 😪
4. Patch a fix for SPM & pods (testable with my fork).

Ideally, upstream repo includes mappings for protected functions and we don't need to maintain a 2-line diff fork.

Even if the precompiled releases had the private modulemap, I'd prefer not to commit such large blobs (esp. on such a quick release schedule), and always try to avoid manual update workflows for my team. All things considered, a typical workflow.

macdrevx:
> Utilize @_spi protected functions, namely globe projections, via the fetched package

Could you share a code snippet that shows specifically how you're doing this to help us reproduce it?

macdrevx:
We'd need to see the import statement and the api usage

e-001:
> > Utilize @_spi protected functions, namely globe projections, via the fetched package
> 
> Could you share a code snippet that shows specifically how you're doing this to help us reproduce it?

@macdrevx Relevent experimental APIs below. [Truncated MVC here](https://gist.github.com/e-001/1ce77dceb12be822bec0c326e3b84069).

```swift
 private func observeProjectionToggle() {
        self.viewModel.$projection.sink { [unowned self] projection in
            do {
                try self.mapboxMap.setMapProjection(projection)
            } catch {
                assertionFailure("Failed to set mapbox to globe view projection")
            }
            guard let userLocation = userLocationHandler.estimatedLocation else {
                return
            }

            switch projection {
            case .mercator:
                mapboxMap.setCamera(to: .init(center: userLocation.coordinate, zoom: 8.5, bearing: .zero, pitch: 0.0))
            case .globe:
                self.flyToGlobeCenteredAt(userLocation)
            }

        }.store(in: &self.cancelBag)
    }
```

macdrevx:
I've tried several times to reproduce this, but it's working consistently for me.

If you're using SPM, base.xcconfig doesn't have any influence on how the library is built.

There must be something else going on. Could you see if you can reproduce the issue using this sample project?

[Issue832.zip](https://github.com/mapbox/mapbox-maps-ios/files/7570953/Issue832.zip)

e-001:
Can you confirm your Xcode version? I recall reading the handling was changed in Xcode 13.

macdrevx:
I've been testing on Xcode 13.1. Which version are you using?

e-001:
Same. Will make repro later

e-001:
On 13.1. The OG warning:

> ```
> '@_spi' import of 'MapboxMaps' will not include any SPI symbols; 'MapboxMaps' was built from the public interface at [redacted]/Build/Products/Debug-iphoneos/MapboxMaps.framework/Modules/MapboxMaps.swiftmodule/arm64-apple-ios.swiftinterface
> ```

Will test your repro and put together my own if things play out differently. Thanks @macdrevx!

macdrevx:
The warning includes `MapboxMaps.framework`. When I integrate via SPM, as in the sample project I posted above, there is no such framework in the build folder. There's only a MapboxMaps.o since it gets linked with the target app statically.

e-001:
> The warning includes `MapboxMaps.framework`. When I integrate via SPM, as in the sample project I posted above, there is no such framework in the build folder. There's only a MapboxMaps.o which gets linked with the target app statically.

**Resolved** in both your reproduction and our codebase. I am _absolutely_ certain this was a "thing" the other day; I have a gut feeling (considering the newness of `@_spi`) that now that we have both compiled locally, embedded locally, etc. Xcode is somehow finding and using an old module map (`@_spi` is only enforced at compile time AFAIK).

Drowning my derived data, failing that will try and repro on a new machine.

macdrevx:
Please do open a new issue if you are able to reproduce again.

macdrevx:
And thanks for investigating with me!

e-001:
She's back! Moved from SPM to the precompiled framework hosted at the link above, any SPI-protected symbols are undefined. All embeded and signed.

Screen Shot 2021-11-20 at 7 40 11 PM

[Reproduced using your example as a base](https://github.com/e-001/mapbox-issue-832). If that repo builds for you, I'd try in clean env (started failing again when I switched machines). Tricky little bug; I'm all but certain it stems from the shared module cache.

[Build log](https://github.com/mapbox/mapbox-maps-ios/files/7575576/Build.Issue832-alt.iOS._2021-11-20T19-15-38.txt).

macdrevx:
Thanks! I can reproduce it now.

macdrevx:
I opened a new PR #854 to move the config to a different location that is more specific to when we build the XCFramework. I also verified that with this setting enabled, the `@_spi` APIs are included in the private module interface. Thanks again for your help!

e-001:
Great! Could you file a radar with Apple? I'm assuming it's an Xcode not compiler issue. Will get more visibility from your team.

macdrevx:
> Could you file a radar with Apple?

As far as I can tell, Xcode is working correctly. Our build was just misconfigured.

e-001:
@macdrevx Can you please notify here when this is fixed? Still broken in latest beta.

macdrevx:
The fix will first appear in v10.2.0-rc.1. To subscribe to release notifications, please watch this repo (if you only want release notifications, you can use the "custom" watch settings.

PR #836, opened 2021-11-15
Update MapboxCoreMaps and MapboxCommon

macdrevx:
## Pull request checklist:

#### Features ✨ and improvements 🏁

* Add heatmap layer support to globe view
* Add circle layer support to globe
* Improve performance for querying large number of features
* Cache layer layout key inside layer, so that it is not re-evaluated at every parse of the every tile
* Model layer now supports loading of external gltf files
* Add cancelable Query Rendered Features API to the Map interface
* Core renderer prints its version on initialization
* Introduce experimental Map.setMemoryBudget method for setting memory budget for a Map and runtime "resource-budget" property for data sources
* Improve performance by avoiding re-layout of invisible fading tiles
* Reduce memory consumption of a model layer by avoiding duplicate texture uploads
* Introduce ViewAnnotation API and corresponding support in gl-native
* Update mapbox-common to v21.0.0-rc.1

#### Bug fixes 🐞

* Release all unused resources when `Map.reduceMemoryUse` is invoked
* Fix crash for the case when an empty fill extrusion bucket is tried to be rendered
* Fix transparency issues with value < 0.5 for models rendered by the model layer 
* Fix regression where setting the same geojson source URL did not refresh the data
* Fix symbol layers with variable anchor placement not being placed correctly on globe view
* Fix crash in symbol reprojection code caused by division by zero
* Fix issue with bounds constraining behavior when terrain is enabled

### DIFF
--- a/Apps/Apps.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/Apps/Apps.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -42,8 +42,8 @@
         "repositoryURL": "https://github.com/mapbox/turf-swift.git",
         "state": {
           "branch": null,
-          "revision": "dccd79aee4c21298d1beb81cc0e86eccbe911ae0",
-          "version": "2.0.0"
+          "revision": "132aaf7f90fb3e334acd685ae7040a2b70b80d26",
+          "version": "2.1.0"
         }
       }
     ]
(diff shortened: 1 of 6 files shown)


endanke (approved):
Seems like Turf might need an update too: https://app.circleci.com/pipelines/github/mapbox/mapbox-maps-ios/4785/workflows/b14bbabe-bf21-4fb7-8653-ca9edc4ab29b/jobs/43407
But approving it anyway, thanks!

PR #846, opened 2021-11-19
View Annotations follow-up

endanke:
This PR fixes some of the open review items for the View Annotations feature.
Original PR: https://github.com/mapbox/mapbox-maps-ios/pull/776
 
Items to cover:
- [x] Remove `AnnotationView` and come up with an alternative way to handle manual `isHidden` changes by the user.
- [x] Move view ID counter to `ViewAnnotationManager`.
- [x] Add simplified example. Android PR: https://github.com/mapbox/mapbox-maps-android/pull/873
- [x] Add more unit tests.
- [x] Further requests (Add below)
    - [x] Audit public API surface

## Pull request checklist:
~~- [ ] Include before/after visuals or gifs if this PR includes visual changes.~~
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-ios` changelog: `<changelog>Remove AnnotationView class and enable usage of UIView in ViewAnnotationManager</changelog>`.
~~- [ ] Update the migration guide, API Docs, Markdown files - Readme, Developing, etc~~


Sources/MapboxMaps/Annotations/ViewAnnotation.swift
```diff
@@ -22,8 +22,15 @@ public final class ViewAnnotationManager {
...
-    private var annotationViewsById: [String: AnnotationView] = [:]
-    private var annotationIdByView: [AnnotationView: String] = [:]
+    private var currentViewId = 0
+    internal private(set) var viewsById: [String: UIView] = [:]
+    internal private(set) var idsByView: [UIView: String] = [:]
+    internal private(set) var expectedHiddenForId: [String: Bool] = [:]
+
+    /// If the `UIView.isHidden` property of a custom view annotation is changed manually by the users
+    /// the SDK prints a warning, as the view is still considered for layout calculation.
+    /// The default value is false, and setting this value to true will disabled the warning.
```
macdrevx, line 32:
```suggestion
    /// The default value is false, and setting this value to true will disable the warning.
```

Sources/MapboxMaps/Annotations/ViewAnnotation.swift
```diff
@@ -48,14 +55,17 @@ public final class ViewAnnotationManager {
     /// Width and height could be specified explicitly but better idea will be not specifying them
     /// as they will be calculated automatically based on view layout.
     ///
+    /// Note: Use `update(view: UIView, options: ViewAnnotationOptions)` for changing the visibilty of the view, instead
+    /// of `UIView.isHidden` to remove it from the layout calculation of the annotations.
```
macdrevx, line 59:
```suggestion
    /// of `UIView.isHidden` so that it is removed from the layout calculation.
```

Sources/MapboxMaps/Annotations/ViewAnnotation.swift
```diff
@@ -66,60 +76,68 @@ public final class ViewAnnotationManager {
...
+    public func remove(_ view: UIView) throws {
+        guard let id = idsByView[view], let annotatonView = viewsById[id] else {
             return
         }
         try mapboxMap.removeViewAnnotation(withId: id)
         annotatonView.removeFromSuperview()
-        annotationViewsById.removeValue(forKey: id)
-        annotationIdByView.removeValue(forKey: annotatonView)
+        viewsById.removeValue(forKey: id)
+        idsByView.removeValue(forKey: annotatonView)
```
macdrevx, line 106: Under what circumstances would `try mapboxMap.removeViewAnnotation(withId: id)` fail? Would we still want to perform the rest of the method in those situations?

Sources/MapboxMaps/Annotations/ViewAnnotation.swift
```diff
@@ -66,60 +76,68 @@ public final class ViewAnnotationManager {
...
-        return annotationViewsById.keys.first(where: { id in
+    /// - Returns: `UIView` if view was found and `nil` otherwise.
+    public func view(forFeatureId identifier: String) -> UIView? {
+        return viewsById.keys.first(where: { id in
             (try? mapboxMap.options(forViewAnnotationWithId: id).associatedFeatureId == identifier) ?? false
         }).flatMap({
-            annotationViewsById[$0]
+            viewsById[$0]
         })
     }
```
macdrevx, line 142: We should profile this. Seems like it could be pretty slow.

Sources/MapboxMaps/Annotations/ViewAnnotation.swift
```diff
@@ -130,17 +148,17 @@ public final class ViewAnnotationManager {
     ///
     /// - Returns: `ViewAnnotationOptions` if view was found and `nil` otherwise.
     public func options(forFeatureId identifier: String) -> ViewAnnotationOptions? {
-        return viewAnnotation(forFeatureId: identifier).flatMap(options(for:))
+        return view(forFeatureId: identifier).flatMap(options(for:))
```
macdrevx, line 151: I think we could come up with a more efficient implementation. This one results in an extra call to `options(forViewAnnotationWithId:)` that could be avoided.

Sources/MapboxMaps/Annotations/ViewAnnotation.swift
```diff
@@ -153,35 +171,36 @@ public final class ViewAnnotationManager {
...
+        guard let view = viewsById[id] else { return }
+        // If the user removed the view from it's superview, then we can remove it from our layout calculation
+        if view.superview == nil {
+            try? remove(view)
         }
-        // View is still considered for layout calculation, users should not modify the visibility of the wrapped view
-        if let wrappedView = annotationView.subviews.first, wrappedView.isHidden {
-            Log.warning(forMessage: "Visibility changed for wrapped view", category: "Annotations")
+        // View is still considered for layout calculation, users should not modify the visibility of view directly
+        if !suppressVisibilityWarnings && view.isHidden != expectedHiddenForId[id] {
```
macdrevx, line 202: Do you want to check the view's alpha here too?

Sources/MapboxMaps/Annotations/ViewAnnotation.swift
```diff
@@ -153,35 +171,36 @@ public final class ViewAnnotationManager {
...
         }
     }
 
     internal func validateAnnotation(byAnnotationId id: String) {
-        guard let annotationView = annotationViewsById[id] else { return }
-        // If the user explicitly removed the ViewAnnotation or it's wrapped view
-        // we need to remove it from our layout calculation
-        if annotationView.subviews.isEmpty || annotationView.superview == nil {
-            try? remove(annotationView)
+        guard let view = viewsById[id] else { return }
```
macdrevx, line 196: Is it possible for this condition to fail?

Sources/MapboxMaps/Annotations/ViewAnnotation.swift
```diff
@@ -153,35 +171,36 @@ public final class ViewAnnotationManager {
...
+        // If the user removed the view from it's superview, then we can remove it from our layout calculation
+        if view.superview == nil {
+            try? remove(view)
         }
```
macdrevx, line 200: Another option would be for us to log a warning and re-add the view for symmetry with how we handle isHidden.

macdrevx (commented):
I like this direction better.

Sources/MapboxMaps/Annotations/ViewAnnotation.swift
endanke: True, and I was thinking to cache the `associatedFeatureId` of the annotations. (Maybe as another key-value pair.)
Internally there's another loop, so if we have a lot of annotations it could get slow: https://github.com/mapbox/mapbox-gl-native-internal/blob/e0253c010a42b7e57dc17331cb47234e41a428aa/internal/src/mapbox/maps/view_annotation_manager.cpp#L307

However it was a decided in the Android implementation to use the underlying functions to query the options, because that way there's a single source of truth for the options.
Personally I think we could make an exception for this to speed up these functions.

Sources/MapboxMaps/Annotations/ViewAnnotation.swift
endanke: Currently it would only fail if it was not found: https://github.com/mapbox/mapbox-gl-native-internal/blob/e0253c010a42b7e57dc17331cb47234e41a428aa/internal/src/mapbox/maps/view_annotation_manager.cpp#L299

Currently it will only happen if the user already removed it. I think I could make the rest of the calls deferred, since we would like to remove it anyway. Unless the implementation changes and the removal could fail for another reason.

Sources/MapboxMaps/Annotations/ViewAnnotation.swift
endanke: Not sure. I think if we set the alpha to zero we still expect it to be part of the layout. Also it was decided on Android to ignore the zero alpha case, so probably it's easier if we don't handle it either.

Sources/MapboxMaps/Annotations/ViewAnnotation.swift
endanke: Only in case we get an invalid ID in `placeAnnotations(positions)` from gl-native. (Maybe it could happen if the view is in the middle of a removal.)
I think I'll just change this function to have the view as the parameter, since I unwrap them in the placement function anyway.

Sources/MapboxMaps/Annotations/ViewAnnotation.swift
endanke: Good idea! Changed it to work that way.

Sources/MapboxMaps/Annotations/ViewAnnotation.swift
macdrevx: Or maybe add a method to GL native that allows getting just the feature id to reduce the amount of serialization/deserialization that has to happen

Sources/MapboxMaps/Annotations/ViewAnnotation.swift
endanke: Actually I've changed it to only use a view, so now it's a bit more cleaner.

Sources/MapboxMaps/Annotations/ViewAnnotation.swift
macdrevx: What about making the method fail silently? Remove `throws` and use `try?`.

Sources/MapboxMaps/Annotations/ViewAnnotation.swift
macdrevx: I also like your idea of caching the feature IDs. Reducing round trips through GL Native is probably going to be the fastest.

Sources/MapboxMaps/Annotations/ViewAnnotation.swift
endanke: That's fine, I think it's better if we throw less. 👌

Sources/MapboxMaps/Annotations/ViewAnnotation.swift
endanke: I think I'll go with the caching, just because it keeps the underlying API more simple. But we can change it later if needed.

macdrevx: And would be great to optimize the GL Native impl too to avoid the looping there. If the expected usage pattern is key-based, wouldn't it make sense to use a key-based lookup structure instead of a list?

endanke: Yes we had a discussion about that. If I remember correctly the problem was that `std::map` is unordered, but we need to keep them ordered for the collision detection. Miao said she will explore some other options to optimize it later on.

Sources/MapboxMaps/Annotations/ViewAnnotation.swift
```diff
@@ -66,6 +67,9 @@ public final class ViewAnnotationManager {
     ///   -  `ViewAnnotationManagerError.geometryFieldMissing` if options did not include geometry
     ///   - `MapError`: errors during insertion
     public func add(_ view: UIView, options: ViewAnnotationOptions) throws {
+        guard idsByView[view] == nil else {
+            return
```
macdrevx, line 71: Should this be an error? What if the options passed this time are different from before?

Sources/MapboxMaps/Annotations/ViewAnnotation.swift
```diff
@@ -86,24 +90,25 @@ public final class ViewAnnotationManager {
         viewsById[id] = view
         idsByView[view] = id
         expectedHiddenByView[view] = !(creationOptions.visible ?? true)
+        creationOptions.associatedFeatureId.flatMap { viewsByFeatureIds[$0] = view }
```
macdrevx, line 93: I don't think I'd use flatMap just for its side-effects. `if let` is more appropriate in this situation.

Sources/MapboxMaps/Annotations/ViewAnnotation.swift
endanke: I was thinking about that too, but thought it might want it to behave similarly to `UIView.addSubview()` which also doesn't fail if we do it twice. But then again if we signal an error it might be more clear for the user in case they try different options.

macdrevx:
Two notes on overall public API:

* We should reconsider the naming of `ViewAnnotationManager.disableViewValidation` in light of this note from the [Swift API Design Guidelines](https://www.swift.org/documentation/api-design-guidelines/):

Screen Shot 2021-11-22 at 11 31 53 AM

`validatesViews` would satisfy that guideline.

* `SubviewInteractionOnlyView.hitTest(_:with:)` is still marked `public` but can now be `internal` since we changed the containing class to be `internal`.

Sources/MapboxMaps/Annotations/ViewAnnotation.swift
endanke: I'll add an error, since it's already a throwing function.

macdrevx:
## Missing tests

- DelegatingViewAnnotationPositionsUpdateListener
    - `onViewAnnotationPositionsUpdate(forPositions:)`
- ViewAnnotationOptions
    - Memberwise init
    - Conversions to/from MapboxCoreMaps objects
- SubviewInteractionOnlyView
    - `hitTest(_:with:)`
- ViewAnnotationManager
    - Test the logic for setting (if necessary) the width and height on the options in `add(_:options:)`
    - Test removing a view that was never added (shouldn't call mapboxMap methods)
    - Test id management (each add should use a new ID)
    - Test update when changing associatedFeatureId; old associatedFeatureId should be disassociated

## APIs

- The following APIs on ViewAnnoationManager are internal just for testing purposes. Can we find a way to make them private?
    - Dicrionary getters
    - `placeAnnotations(positions:)`
    - `validate(_:)`
- Does ViewAnnotationManager depend on containerView being a SubviewInteractionOnlyView, or could it just be a UIView?
    - Would be nice to just inject a UIView when testing for better isolation.

## Behavior questions

- What should happen if you try to add more than one annotation with the same associatedFeatureId?
- When allowOverlap is true, what determines the z ordering of the annotations?

## Style

- In ViewAnnotationTests, rename `mockMapboxMap` to `mapboxMap`
- Rename `ViewAnnotationTests` to `ViewAnnotationManagerTests`
- Refine `ViewAnnotationPositionDescriptor` init method in unit tests to clean up implementation

## Improvements to existing tests

- ViewAnnotationTests
    - testAddView
        - instead of just checking the invocation count of mapboxMap.addViewAnnotationStub, actually compare the parameters to what is expected.
    - testRemove
        - verify that the expected parameters are passed to mapboxMap methods

endanke:
Thanks @macdrevx! I've added checkmarks to the list, and here are also some notes:

- Setting the dictionary getters private
    - I think leaving them internal simplifies the testing a lot. (Same setup on Android.) Do you have any suggestions how else could we check their state without exposing them? I could maybe set them to protected and subclass the manager for testing.
- What should happen if you try to add more than one annotation with the same associatedFeatureId?
   - Discussed on slack, we'll add a check on both platforms to prevent using the same associatedFeatureId twice.
- When allowOverlap is true, what determines the z ordering of the annotations?
   - Based on the gl-native implementation the z-ordering depends on the order of the additions. So the last added view will be on top. I'll add a documentation comment about this.

Tests/MapboxMapsTests/Foundation/DelegatingViewAnnotationPositionsUpdateListenerTests.swift
```diff
@@ -0,0 +1,33 @@
...
+        let receiver = Receiver()
+        let delegatingPositionsListener = DelegatingViewAnnotationPositionsUpdateListener()
+        delegatingPositionsListener.delegate = receiver
+
+        let descriptor = ViewAnnotationPositionDescriptor(
+            __identifier: "test",
+            width: UInt32(100),
+            height: UInt32(50),
+            leftTopCoordinate: ScreenCoordinate(x: 100.0, y: 100.0)
+        )
```
macdrevx, line 26: Can use the refined constructor

Tests/MapboxMapsTests/Foundation/DelegatingViewAnnotationPositionsUpdateListenerTests.swift
```diff
@@ -0,0 +1,33 @@
...
+
+    private class Receiver: DelegatingViewAnnotationPositionsUpdateListenerDelegate {
+
+        var positions: [ViewAnnotationPositionDescriptor] = []
+
+        internal func onViewAnnotationPositionsUpdate(forPositions positions: [ViewAnnotationPositionDescriptor]) {
+            self.positions = positions
+        }
+
+    }
```
macdrevx, line 14: We typically put these in separate files and name them according to the pattern `Mock<ProtocolName>`

We'd also use `Stub` to capture the arguments instead of a simple property.

macdrevx: Mock and test source files are organized like this:

For mocks, if the protocol `C` is located at `/Sources/A/B/C.swift`, then the corresponding mock would be placed at `/Tests/A/B/Mocks/MockC.swift`

For tests, if the entity `Z` is located at `/Sources/X/Y/Z.swift`, then the corresponding the test would be place at `/Tests/X/Y/ZTests.swift`

macdrevx: cc: @OdNairy FYI

macdrevx:
Not seeing the test for "Test update when changing associatedFeatureId; old associatedFeatureId should be disassociated"

Tests/MapboxMapsTests/Foundation/ViewAnnotationManagerTests.swift
```diff
@@ -0,0 +1,261 @@
...
+    }
+
+    func testAddView() {
+        let testView = UIView()
+        let geometry = Geometry.point(Point(CLLocationCoordinate2D(latitude: 0.0, longitude: 0.0)))
+        let options = ViewAnnotationOptions(geometry: geometry, width: 0.0, height: 0.0)
+        try? manager.add(testView, options: options)
+        XCTAssertEqual(mapboxMap.addViewAnnotationStub.invocations.count, 1)
+        XCTAssertEqual(mapboxMap.addViewAnnotationStub.invocations.last?.parameters.id, "0")
+        XCTAssertEqual(mapboxMap.addViewAnnotationStub.invocations.last?.parameters.options, options)
```
macdrevx, line 32: If you make the params struct Equatable, you can write this succinctly as

```
XCTAssertEqual(mapboxMap.addViewAnnotationStub.parameters, [.init(id: "0", options: options)])
```

Tests/MapboxMapsTests/Foundation/ViewAnnotationManagerTests.swift
```diff
@@ -0,0 +1,261 @@
...
+
+fileprivate extension MapboxCoreMaps.ViewAnnotationPositionDescriptor {
+    convenience init(identifier: String, width: Int, height: Int, leftTopCoordinate: CGPoint) {
+        self.init(__identifier: identifier,
+                  width: UInt32(width),
+                  height: UInt32(height),
+                  leftTopCoordinate: ScreenCoordinate(x: leftTopCoordinate.x, y: leftTopCoordinate.y)
+        )
+    }
+}
```
macdrevx, line 261: Let's move this to a separate file so it can be reused. We can put it in `/Tests/MapboxMapsTests/Helpers/Extensions/ViewAnnotationPositionDescriptor.swift`

macdrevx:
Looking great!

endanke:
> Not seeing the test for "Test update when changing associatedFeatureId; old associatedFeatureId should be disassociated"

Yes I forgot about that, only added a test for checking the dissociation when the view is removed. But great that you mentioned it because there was a bug that this test found! :D

macdrevx (approved):

Issue #855, opened 2021-11-22
Localisation replaces all instances of fallback languages of a layer


sosthoff:
## Environment

- Xcode version: 13.1
- iOS version: 15.0.2
- Devices affected: all
- Maps SDK Version: 10.1.0

## Observed behavior and steps to reproduce

Setting 
```
let deLocale = Locale(identifier: "de")
            do {
                try style.localizeLabels(into: deLocale)
```
results in all name_* of every layer to be overridden with name_de

## Expected behavior

Only the primary localized language should be overridden.

## Notes / preliminary analysis

Style+Localization::updateExpression applies to all occurrences in range

## Fix 
Add
```
    /// Updates string once using the first occurrence of a regex
    /// - Parameters:
    ///   - replacement: New string to replace the matched pattern
    ///   - regex: The regex pattern that will be matched for replacement
    internal mutating func updateOnceExpression(replacement: String, regex: NSRegularExpression) {
        var range = NSRange(location: 0, length: self.count)
        range = regex.rangeOfFirstMatch(in: self, options: [], range: range)
        if (range.lowerBound == NSNotFound) {
            return
        }
        self = regex.stringByReplacingMatches(in: self,
                                              options: [],
                                              range: range,
                                              withTemplate: replacement)
    }
```
and replace 
`stringExpression.updateExpression(replacement: replacement, regex: expressionCoalesce)`
with
`stringExpression.updateOnceExpression(replacement: replacement, regex: expressionCoalesce)`

sosthoff:
[Commit](https://github.com/mapbox/mapbox-maps-ios/commit/d916cfade73084f3dbe4d1c46a43e32e3e3ea166) for now

Issue #653, opened 2021-09-07
Unable to unlocalize a layer’s labels


1ec5:
`Style.localizeLabels(into:forLayerIds:)` should have a way to explicitly unlocalize a layer’s labels – that is, make it show names in the local language instead of a specific passed-in language. This was supported in Mapbox Maps SDK v6._x_ but has regressed as of #480.

## Rationale

It should be possible to easily undo what `Style.localizeLabels(into:forLayerIds:)` does when passing in a specific locale.

It should also be possible to ensure that a layer matches the local language, in the event that the original style hard-coded a particular language. For example, the Mapbox Navigation Day/Night v1 styles hard-code English in its road labels, falling back to the local language. This is suboptimal for the Mapbox Navigation SDK, which is marketed around the world, so it would be beneficial to force the local language name instead of English.

## Design considerations

In v6._x_ and the proof-of-concept port in mapbox/mapbox-navigation-ios#2933, `-[NSExpression mgl_expressionLocalizedIntoLocale:]` [supports two special Locale values](https://github.com/mapbox/mapbox-gl-native-ios/blob/b34748257b7507b12684eb1fe3930ca177ac44cd/platform/darwin/src/NSExpression%2BMGLAdditions.mm#L1562-L1568), both documented:

* `nil` automatically selects the first language listed in the system Preferred Content Languages list (`+[NSLocale preferredLanguages]`) that is also supported by the Streets source.
* A Locale with the identifier `mul` (the standard ISO&nbsp;639-2 code for multilingual text) selects the `name` field, which is [typically in the local language](https://wiki.openstreetmap.org/wiki/Key:name).

Unfortunately, neither of these options are available in v10. The `locale` argument is non-optional, though it isn’t a big deal to pass in the first value in `NSLocale.preferredLanguages` manually. More problematic is that passing in `mul` currently crashes, because it isn’t one of the hard-coded locale codes supported by Streets:

https://github.com/mapbox/mapbox-maps-ios/blob/34ea7ada3cea926ea308d00dd003178a16d3164e/Sources/MapboxMaps/Style/Style%2BLocalization.swift#L10-L12

/cc @mapbox/maps-ios @mapbox/navigation-ios

1ec5:
> `nil` automatically selects the first language listed in the system Preferred Content Languages list (`+[NSLocale preferredLanguages]`) that is also supported by the Streets source.

[...]

PR #856, opened 2021-11-23
Fix to only localize the primary localization and not the fall-throug…

sosthoff:
…h localizations

Fixes: https://github.com/mapbox/mapbox-maps-ios/issues/855

## Pull request checklist:
 ~~- [ ] Include before/after visuals or gifs if this PR includes visual changes.~~
 ~~- [ ] Document any changes to public APIs.~~

### Summary of changes

Reduces the programatically set localization to the primary localization, instead of overriding all fallbacks with the primary.
Eg. layer defines: name_en, name_de, name - localizeLabales("ko") no longer results in: name_ko, name_ko, name - but in name_ko, name_de, name.

This should help until we can have: https://github.com/mapbox/mapbox-maps-ios/issues/653
to hopefully also programatically set all fallback languages.

No tests added, since existing cover the previously expected behavior.

### User impact (optional)

a customer with be able ship their v10 migration

### DIFF
--- a/Sources/MapboxMaps/Style/Style+Localization.swift
+++ b/Sources/MapboxMaps/Style/Style+Localization.swift
@@ -113,7 +113,7 @@ extension String {
                                               range: range,
                                               withTemplate: replacement)
     }
-    
+
     /// Updates string once using the first occurrence of a regex
     /// - Parameters:
     ///   - replacement: New string to replace the matched pattern
@@ -121,9 +121,8 @@ extension String {
     internal mutating func updateOnceExpression(replacement: String, regex: NSRegularExpression) {
         var range = NSRange(location: 0, length: self.count)
         range = regex.rangeOfFirstMatch(in: self, options: [], range: range)
-        if (range.lowerBound == NSNotFound) {
+        if (range.lowerBound == NSNotFound)
             return
-        }
         self = regex.stringByReplacingMatches(in: self,
                                               options: [],
                                               range: range,
--- a/Sources/MapboxMaps/Style/Style+Localization.swift
+++ b/Sources/MapboxMaps/Style/Style+Localization.swift
@@ -119,7 +119,8 @@ extension String {
     ///   - replacement: New string to replace the matched pattern
     ///   - regex: The regex pattern that will be matched for replacement
     internal mutating func updateOnceExpression(replacement: String, regex: NSRegularExpression) {
-        var range = regex.rangeOfFirstMatch(in: self, options: [], range: range)
+        var range = NSRange(location: 0, length: NSString(string: self).length)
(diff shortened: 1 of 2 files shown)


sosthoff:
My first PR here 😀
I assume migration guide, API Docs, Markdown files can be skipped , trying to identify where to add the label and changelog entry 😅

Sources/MapboxMaps/Style/Style+Localization.swift
```diff
@@ -113,4 +113,19 @@ extension String {
                                               range: range,
                                               withTemplate: replacement)
     }
+
+    /// Updates string once using the first occurrence of a regex
+    /// - Parameters:
+    ///   - replacement: New string to replace the matched pattern
+    ///   - regex: The regex pattern that will be matched for replacement
+    internal mutating func updateOnceExpression(replacement: String, regex: NSRegularExpression) {
+        var range = NSRange(location: 0, length: self.count)
```
macdrevx, line 122: we should be using NSString(self).length instead of self.count. They can produce different results under some circumstances.

I would actually be in favor of converting this entire extension to be an extension on NSString instead of String.

Sources/MapboxMaps/Style/Style+Localization.swift
```diff
@@ -113,4 +113,19 @@ extension String {
...
+        range = regex.rangeOfFirstMatch(in: self, options: [], range: range)
+        if (range.lowerBound == NSNotFound)
+            return
```
macdrevx, line 125: Swift requires { }

Sources/MapboxMaps/Style/Style+Localization.swift
sosthoff: we could also remove the line `NSRange(location: 0, length: self.count)` entirely, I wanted to use it as an initializer, but rangeOf is guaranteed to return a range either way, so it should be fine

Sources/MapboxMaps/Style/Style+Localization.swift
```diff
@@ -113,4 +113,19 @@ extension String {
                                               range: range,
                                               withTemplate: replacement)
     }
+
+    /// Updates string once using the first occurrence of a regex
+    /// - Parameters:
+    ///   - replacement: New string to replace the matched pattern
+    ///   - regex: The regex pattern that will be matched for replacement
+    internal mutating func updateOnceExpression(replacement: String, regex: NSRegularExpression) {
+        var range = regex.rangeOfFirstMatch(in: self, options: [], range: range)
```
macdrevx, line 122: I don't think this will compile. `range` is needed as the 3rd argument to `rangeOfFirstMatch`

macdrevx (approved):

PR #862, opened 2021-11-24
Delay adding puck images until adding layer

### LINKED ISSUE #860: Puck not showing during initial launch with 10.2.0-beta.1
## Environment

- Xcode version: 13.1
- iOS version: 15.0.2
- Devices affected: tested iPad & Simulator
- Maps SDK Version: 10.2.0-beta.1

When running the sample application https://docs.mapbox.com/ios/maps/examples/tracking-mode/ for the first time, the puck is not displayed.
Steps: 
1. Run application for the first time on a device
2. Grant Location permission while in use
3. Observe camera move to location & puck missing

Future launches of the application display the puck

This issue does not exist in 10.1.0

## Expected behavior

Steps: 
1. Run application for the first time on a device
2. Grant Location permission while in use
3. Observe camera move to location & puck

It seems to be a regression as I cannot reproduce in 10.1.0 and probably related to the fact that `mapView.location.options.puckType = .puck2D()`is called before the location access permission is granted during the first launch.

macdrevx:
Fixes #860

## Pull request checklist:
 - [x] Update the changelog

### Summary of changes

Fixes an issue that caused the location puck to fail to render if it was configured prior to the style loading and the first location update didn't happen until after the style loaded.

Previously, the style images were added to the style as soon as the puck was activated. However, in the situation described above, the persistent puck layer would not be added until after the style loaded. MapboxCoreMaps clears style images during reload if there are no persistent layers. That meant that the puck images were missing when the puck layer was finally added.

This fix updates the logic to delay adding the images to the style until it's time to add the persistent layer.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ Mapbox welcomes participation and contributions from everyone.
 * Removed experimental designation from persistent layer APIs. ([#849](https://github.com/mapbox/mapbox-maps-ios/pull/849))
 * Removed `AnnotationView` wrapper views from `ViewAnnotationManager` API. ([#846](https://github.com/mapbox/mapbox-maps-ios/pull/846))
 * Reduce geometry wrapping using GeometryConvertible. ([#861](https://github.com/mapbox/mapbox-maps-ios/pull/861))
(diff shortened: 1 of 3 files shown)


sosthoff:
functional ✅

OdNairy (approved):

PR #888, opened 2021-12-01
Improve panning behavior on pitched maps

macdrevx:
## Pull request checklist:
 - [x] Update the changelog

### Summary of changes

Improves panning behavior on pitched maps:

* Panning does not begin until a touch is below the horizon
* Pan deceleration is disabled if the final touch location is above the horizon
* Pan deceleration is simulated as a series of small pan gestures in the bottom portion of the map to avoid allowing the simulated touches to go near the more sensitive horizon area.

### DIFF
--- a/Sources/MapboxMaps/Gestures/GestureHandlers/PanGestureHandler.swift
+++ b/Sources/MapboxMaps/Gestures/GestureHandlers/PanGestureHandler.swift
@@ -84,9 +84,7 @@ internal final class PanGestureHandler: GestureHandler, PanGestureHandlerProtoco
                 return
             }
             lastChangedDate = dateProvider.now
-            let clampedTouchLocation = clampTouchLocation(
-                touchLocation,
-                previousTouchLocation: previousTouchLocation)
+            let clampedTouchLocation = touchLocation.clamped(to: previousTouchLocation, panMode: panMode)
             pan(from: previousTouchLocation, to: clampedTouchLocation)
             self.previousTouchLocation = clampedTouchLocation
         case .ended:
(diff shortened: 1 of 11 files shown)


astojilj:
I find the fling inertia on iPad (Pro 11) on tilted map very nice:
https://user-images.githubusercontent.com/549216/144435099-d8a703a2-405b-482e-9acc-24355b055ccd.mov

Sources/MapboxMaps/Gestures/GestureHandlers/PanGestureHandler.swift
```diff
@@ -73,46 +96,55 @@ internal final class PanGestureHandler: GestureHandler, PanGestureHandlerProtoco
...
+            // to avoid simulated interaction with the more sensitvie area near the horizon.
+            let height = mapboxMap.size.height
+            let initialDecelerationLocation = CGPoint(
+                x: touchLocation.x,
+                y: max(touchLocation.y, 3 * height / 4))
             cameraAnimationsManager.decelerate(
-                location: touchLocation,
-                velocity: gestureRecognizer.velocity(in: view),
+                location: initialDecelerationLocation,
+                velocity: clampTouchLocation(velocity, previousTouchLocation: .zero),
```
astojilj, line 114: Just would like to understand, why is .zero here not causing issues when pan direction is clamped to be vertical?

astojilj (approved):
Mostly rubber stamp on tests code, reviewed gesture implementation. Code change looks good and it performs nice on iOS.

Sources/MapboxMaps/Gestures/GestureHandlers/PanGestureHandler.swift
macdrevx: Previously, to achieve horizontal or vertical clamping during fling, we'd clamp the simulated touchLocation. While working on this, I realized that we could achieve the same effect more elegantly by setting the velocity to 0 for the dimension where we don't want any movement.

I'm just reusing the same method that was originally named with the idea that it'd be used to clamp positions. Now that it's used to clamp positions during drag and velocities during fling, I can update the name to be more generic.

macdrevx: Updated in 86f19cb to hopefully be a bit more clear

Issue #864, opened 2021-11-25
Panning while zooming still feels a bit strange


georgbachmann:
## Environment

- Xcode version: 13.1
- iOS version: 15.1
- Devices affected: All my devices
- Maps SDK Version: 10.1 but also 10.2-beta

## Observed behavior and steps to reproduce

When you zoom in to your map using a two finger pinch gesture and also while doing so move your two fingers up and down (i do that very often when zooming in on a specific part of the map) the horizontal-movement of the map is off. It seems as if the lat/lng translation of the map (the one from the panning) still has the original-zoom level as a translation-factor. I hope you understand what I mean.
So... when you first zoom IN using two finger pinch... keep your fingers on the display... now pan -> the map will move much faster than your fingers.
On the other hand -> if you first zoom OUT and then pan -> the map will move much slower than your fingers.

I guess (didn't look at code as I don't know if that's even for me to see?!?) it's because the factor for the panning is zoom-releated, right? And maybe that does not update while the zooming is in progress?!? Just a thought...

Hope you understand what I mean... I could try to make a video somehow if not :)

## Expected behavior

The map should feel sticky under the fingers. Like Apple or Google Maps does.

macdrevx:
Verified. Thanks for reporting this.

macdrevx:
This was introduced in 7c2c073838a338504af81246c31e107a95c4e7ba which was first included in 10.2.0-beta.1.

I'm actually not able to repro in 10.1.x unless I zoom out so far that the map isn't allowed to pan north-south enough to follow the touch locations, but that's expected behavior.

macdrevx:
In https://github.com/mapbox/mapbox-maps-ios/pull/890, we're restoring the default behavior to match 10.1. This should resolve this issue. Please reopen if you find that is not the case.

PR #837, opened 2021-11-15
Implement iterative pan gesture handling

macdrevx:
## Pull request checklist:
 - [x] Briefly describe the changes in this PR.
 - [x] Write tests for all new functionality. If tests were not written, please explain why.
 - [x] Apply changelog label ('breaking change', 'bug :beetle:', 'build', 'docs', 'feature :green_apple:', 'performance :zap:', 'testing :100:') or use the label 'skip changelog'

### Summary of changes

Previously, each frame of a gesture update was calculated as a function of the state when the gesture started + the total gesture change. However, the Drag API was designed to calculated updates incrementally, so we're switching the implementation.

This change anticipates in-progress Drag API improvements, but also provides some immediate benefits by avoiding camera change notifications when resetting the camera to the initial state during pan and pinch gestures:

Fixes #775

Note that each frame of a pinch gesture will still result in 2 camera updates since the bearing and zoom changes have to be set with an anchor and the center changes must not be set with anchor. However, these updates should be logically progressing toward the new camera state rather than jumping back to the initial state at each frame.

Additionally, we've added some defensive logic to try to help avoid and diagnose some crashes we were seeing with PinchGestureHandler:

Fixes #833

PR #890, opened 2021-12-02
Add pinch gesture tradeoff configuration option

### LINKED ISSUE #775: Camera state update on camera changing issue
## Environment

- Xcode version: 13
- iOS version: 15
- Devices affected: iPad
- Maps SDK Version: v10.0.1

Change map camera by pinch and pan gestures.
The `.cameraChanged` event is reported.
Print camera state on every reported event. Printed state will switch between new correct state and old state at the moment of gesture start.

```
mapView.mapboxMap.onEvery(.cameraChanged) { _ in
            print(self.mapView.mapboxMap.cameraState)
        }
```

The print out will be similar to this. See that first state is equal to third state when second state is correct and it will go like that: correct value, then old value, then again correct and so on.
```
MapboxMaps.CameraState(center: __C.CLLocationCoordinate2D(latitude: 48.67914111569259, longitude: 34.77753011541918), padding: __C.UIEdgeInsets(top: 100.0, left: 100.0, bottom: 100.0, right: 100.0), zoom: 5.762091022431934, bearing: -0.0, pitch: 0.0)
MapboxMaps.CameraState(center: __C.CLLocationCoordinate2D(latitude: 46.83731384346994, longitude: 45.52068366519899), padding: __C.UIEdgeInsets(top: 100.0, left: 100.0, bottom: 100.0, right: 100.0), zoom: 5.762091022431934, bearing: -0.0, pitch: 0.0)
MapboxMaps.CameraState(center: __C.CLLocationCoordinate2D(latitude: 48.67914111569259, longitude: 34.77753011541918), padding: __C.UIEdgeInsets(top: 100.0, left: 100.0, bottom: 100.0, right: 100.0), zoom: 5.762091022431934, bearing: -0.0, pitch: 0.0)
```

## Expected behavior

self.mapView.mapboxMap.cameraState will return a new correct (currently visible map) state on every event reported.

macdrevx:
## Pull request checklist:
 - [x] Update the changelog

### Summary of changes

In #837, we fixed #775, but the solution introduced a new issue, #864. At the moment, we don't have a way to solve both issues simultaneously, so this PR introduces a configuration option to allow developers to make the tradeoff between the two issues that is best for their use case.

The new configuration option is marked as experimental because we hope to eliminate the need for developers to make this tradeoff in a future release. Note that experimental APIs may be modified or removed without a major version bump.

The default configuration is set to match the v10.1 behavior.

To use the new option, annotate your import statement with `@_spi(Experimental)` and then set `mapView.gestures.options.pinchBehavior = .doesNotResetCameraAtEachFrame`.

This option may be changed dynamically at any time and the change will take effect when the next pinch gesture begins.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Mapbox welcomes participation and contributions from everyone.
 * Fixed an issue that could prevent the location puck from appearing. ([#862](https://github.com/mapbox/mapbox-maps-ios/pull/862))
 * Add support for exponentials to `StyleColor`. ([#873](https://github.com/mapbox/mapbox-maps-ios/pull/873))
 * Improve panning behavior on pitched maps. ([#888](https://github.com/mapbox/mapbox-maps-ios/pull/888))
+* Add pinch gesture tradeoff configuration option. ([#890](https://github.com/mapbox/mapbox-maps-ios/pull/890))
 
 ## 10.2.0-beta.1 - November 19, 2021
 
--- a/Sources/MapboxMaps/Gestures/GestureHandlers/PinchGestureHandler.swift
+++ b/Sources/MapboxMaps/Gestures/GestureHandlers/PinchGestureHandler.swift
@@ -4,171 +4,83 @@ internal protocol PinchGestureHandlerProtocol: GestureHandler {
     var rotateEnabled: Bool { get set }
 }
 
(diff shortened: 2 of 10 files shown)


OdNairy (approved):

pengdev (approved):
The experimental API looks good to me

PR #902, opened 2021-12-06
Add refinement to access TransitonOptions

ZiZasaurus:
Created to address issue where users are unable to control the animation of symbol layers when they are added to the map view.


Sources/MapboxMaps/Foundation/Extensions/Core/TransitionOptions.swift
```diff
@@ -0,0 +1,9 @@
+import Foundation
+extension MapboxCoreMaps.TransitionOptions {
+    public convenience init(duration: NSNumber,
+                            delay: NSNumber,
+                            enablePlacementTransitions: NSNumber.BooleanLiteralType? = nil){
```
macdrevx, line 5:
```suggestion
    public convenience init(duration: TimeInterval?,
                            delay: TimeInterval?,
                            enablePlacementTransitions: Bool?){
```

macdrevx: And maybe they should all default to `nil`

macdrevx (approved):

PR #903, opened 2021-12-06
Ensure Metal view has a non-zero area

macdrevx:
Fixes a crash that could occur otherwise.

cc: @pjleonard37


pjleonard37 (approved):
Looks good to me 👍

Issue #895, opened 2021-12-03
Globe Example incorrect display on iPhone6s plus


lishuo0710:
## Environment

- Xcode version:12
- iOS version:13.6.1
- Devices affected:iPhone6s plus
- Maps SDK Version: 10.2.0. rc1

Globe Example incorrect display as below
![20211203162359](https://user-images.githubusercontent.com/25873946/144569790-abf349c4-58ca-4e4c-9af2-c86a8a9089bd.jpg)
.

macdrevx:
cc: @astojilj @mpulkki-mapbox

macdrevx:
also @endanke in case this is an issue with our example

endanke:
@macdrevx I have a strong feeling that this is caused by the pixel ratio of the device. (I saw a similar issue previously when moved the surface to an external display with different density.) 
I'll investigate this further after the long weekend and try to prepare a fix!

endanke:
I did some further investigation and indeed the pixel ratio is the issue: currently we render the atmosphere behind the globe using the size and pixel ratio of the screen. (It's planned to be refactored later on to make it independent of those.)
In case the device is using a different screen scaling than the native one, then the atmosphere will have this alignment issue. I think if we use `UIScreen.mainScreen.nativeScale` instead of `UIScreen.mainScreen.scale` as a default pixel ratio in `MapboxCoreMaps.MapOptions` then it should solve the issue. 

I tested this using the Display Zoom feature on an iPhone 11 Pro, which also changes the scaling:

Using `UIScreen.mainScreen.scale` in options:

![IMG_8837](https://user-images.githubusercontent.com/2576246/145038425-1170b1aa-c972-4195-adec-33568401b3dd.PNG)

Using `UIScreen.mainScreen.nativeScale` in options:

![IMG_8838](https://user-images.githubusercontent.com/2576246/145038418-4e12ceb2-5db1-48f8-8194-13b516a37e78.PNG)

endanke:
Created a PR to change the default pixel ratio: https://github.com/mapbox/mapbox-maps-ios/pull/906
Also opened an issue to make the atmosphere shader independent from this value, which would prevent other issues in the future (in case a customer wants to explicitly use a different pixel ratio): https://github.com/mapbox/mapbox-3d-team/issues/71

PR #906, opened 2021-12-07
Use UIScreen.mainScreen.nativeScale as default pixel ratio in MapOptions

endanke:
Fixes: https://github.com/mapbox/mapbox-maps-ios/issues/895

Before:
IMG_8837

After:
IMG_8838

## Pull request checklist:
~~- [ ] Write tests for all new functionality. If tests were not written, please explain why.~~
~~- [ ] Add example if relevant.~~
 - [x] Add an entry inside this element for inclusion in the `mapbox-maps-ios` changelog: `<changelog>Set default pixel ratio to use UIScreen.main.nativeScale</changelog>`.
~~- [ ] Update the migration guide, API Docs, Markdown files - Readme, Developing, etc~~

### Summary of changes

See: https://github.com/mapbox/mapbox-maps-ios/issues/895#issuecomment-987930619

### DIFF
--- a/Sources/MapboxMaps/Foundation/Extensions/Core/MapOptions.swift
+++ b/Sources/MapboxMaps/Foundation/Extensions/Core/MapOptions.swift
@@ -11,7 +11,7 @@ extension MapboxCoreMaps.MapOptions {
     ///   - orientation: The view orientation; default is `.upwards`.
     ///   - crossSourceCollisions: Whether cross-source symbol collision detection should be enabled; default is `true`
     ///   - size: Size of the map, if nil (the default), a minimal default size will be used.
-    ///   - pixelRatio: Pixel scale of the map view; default is thee main screen's scale.
+    ///   - pixelRatio: Pixel scale of the map view; default is the main screen's native scale.
     ///   - glyphsRasterizationOptions: A `GlyphsRasterizationOptions` object.
     public convenience init(constrainMode: ConstrainMode = .heightOnly,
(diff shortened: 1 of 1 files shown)


macdrevx:
Seems like the setting we're using in GL Native should always match the setting used by the metal view. This change fixes the default behavior since MTKView is [configured based on nativeScale by default](https://developer.apple.com/library/archive/documentation/3DDrawing/Conceptual/MTLBestPracticesGuide/NativeScreenScale.html). But what happens if someone tries to use a custom pixel ratio?

endanke:
@macdrevx 
Normally our shaders should be independent from the screen's size and pixel ratio, but in this case the globe's background was implemented in a way that it only supports the native scale. I've opened a ticket to change this: https://github.com/mapbox/mapbox-3d-team/issues/71

So I think we could still consider it as an open bug, but changing the default value could be useful regardless.

macdrevx:
I agree. We should change the default value regardless. Could you please open a ticket to track keeping the metal view's config in sync with the specified pixel ratio?

macdrevx (approved):

endanke:
I think we set the metal view's scale based the supplied pixel ratio during the initialization: https://github.com/mapbox/mapbox-maps-ios/blob/cf8d26314ed0f7df7842a80c88fb8579a0cdaeae/Sources/MapboxMaps/Foundation/MapView.swift#L524

Although now I wonder why did we still have this issue then. I'll open a ticket and we can check it later.

PR #908, opened 2021-12-07
Send a turnstile event using Snapshotter

jmkiley:
## Pull request checklist:
 - [x] Update the changelog

### Summary of changes

`Snapshotter` will now send a turnstile event upon initialization. This mirrors https://github.com/mapbox/mapbox-maps-android/pull/920

`EventsManager` has been simplified and converted to a singleton to address [the comment below](https://github.com/mapbox/mapbox-maps-ios/pull/908#discussion_r769074668).


Sources/MapboxMaps/Snapshot/Snapshotter.swift
```diff
@@ -96,6 +96,8 @@ public class Snapshotter {
                 return
             }
 
+            sendTurnstileEvent()
```
macdrevx, line 99: Can you explain why you chose to send the turnstile here instead or say at the beginning of the method or all the way at the end prior to invoking the completion block with success?

Sources/MapboxMaps/Snapshot/Snapshotter.swift
```diff
@@ -403,3 +405,11 @@ extension Snapshotter {
         attributionView.layer.render(in: context)
     }
 }
+
+// MARK: - Telemetry
+internal func sendTurnstileEvent() {
```
macdrevx, line 410: Could/should this be an instance method instead of a global function?

Sources/MapboxMaps/Snapshot/Snapshotter.swift
jmkiley: This probably should be on line 107, actually. I intended to put it after the guard statements that resulted in failures.

Sources/MapboxMaps/Snapshot/Snapshotter.swift
```diff
@@ -251,6 +257,14 @@ public class Snapshotter {
                                 bearing: bearing?.NSNumber,
                                 pitch: pitch?.NSNumber))
     }
+
+    // MARK: - Telemetry
+    internal func sendTurnstileEvent() {
+        let accessToken = ResourceOptionsManager.default.defaultAccessToken()
```
macdrevx, line 263: Snapshotter is initialized with `MapSnapshotOptions` which have `ResourceOptions`. We should use the access token from there instead.

Sources/MapboxMaps/Snapshot/Snapshotter.swift
```diff
@@ -251,6 +257,14 @@ public class Snapshotter {
...
+        let eventsManager = EventsManager(accessToken: accessToken)
```
macdrevx, line 264: We create a single EventsManager per MapView instance, but this will create a separate EventsManager per snapshot (not per Snapshotter). We could avoid paying the initialization cost repeatedly if we just created this once at Snapshotter initialization.

Sources/MapboxMaps/Snapshot/Snapshotter.swift
```diff
@@ -251,6 +257,14 @@ public class Snapshotter {
...
+
+        eventsManager.telemetry.turnstile()
```
macdrevx, line 266: Let's avoid accessing the `telemetry` property of `EventsManager`. Instead, we should create some internal API on `EventsManager` that provides an abstraction. That will hopefully limit the scope of changes when we adopt core telemetry.

macdrevx, line 266: Also, should we send a turnstile with each snapshot or just once per Snapshotter (e.g. after the first successful snapshot for a given Snapshotter)?

Sources/MapboxMaps/Snapshot/Snapshotter.swift
```diff
@@ -32,6 +32,7 @@ public class Snapshotter {
         self.options = options
         mapSnapshotter = MapSnapshotter(options: MapboxCoreMaps.MapSnapshotOptions(options))
         style = Style(with: mapSnapshotter)
+        EventsManager(accessToken: options.resourceOptions.accessToken).sendTurnstile()
```
macdrevx, line 35: This is the important change. This approach matches [Android](https://github.com/mapbox/mapbox-maps-android/pull/920), but there are a few questions we need to think about:

* EventsManager is a thin wrapper around `MMEEventsManager.shared()`. Each time an `EventsManager` is initialized, it will call `MMEEventsManager.shared().initialize(…)`. Is it okay for that to happen more than once (will happen with each snapshotter and map view that is created). What if you are using different access tokens for different purposes in the same app?
    * Based on https://github.com/mapbox/mapbox-events-ios/blob/main/Sources/MapboxMobileEvents/MMEEventsManager.m#L92-L101, it seems like subsequent calls to initialize will just update the access token, but won't repeat any other setup.
* One problem with having more than one instance of `EventsManager` is that each instance will observe changes to the `"MGLMapboxMetricsEnabled"` user defaults key and will call `flush` upon receiving a memory warning. These behaviors only need to happen once per `MMEEventsManager` instance.
    * This MME design means we need to convert EventsManager to a singleton as well to avoid doing more work than necessary. @vesh93 I'd love to see us move away from using a singleton pattern. It's a bad match for our initialization APIs which would lead you to believe there's nothing stopping you from using multiple access tokens simultaneously in the same process.

cc: @vesh93

macdrevx: Updated to convert EventsManager to a singleton.

Chaoba (approved):
LGTM

vesh93 (approved):

PR #917, opened 2021-12-09
Release automation: open documentation pull requests

OdNairy:

## Pull request checklist:


.circleci/config.yml
```diff
@@ -632,12 +633,32 @@ commands:
...
+        type: string
+        default: /usr/local/bin
     steps:
       - run:
           name: Install latest mbx-ci
           command: |
-            curl -Ls https://mapbox-release-engineering.s3.amazonaws.com/mbx-ci/latest/mbx-ci-darwin-amd64 > mbx-ci && chmod 755 ./mbx-ci
-            ./mbx-ci aws setup
+            curl -Ls https://mapbox-release-engineering.s3.amazonaws.com/mbx-ci/<<parameters.version>>/mbx-ci-<<parameters.platform>>-<<parameters.arch>> > <<parameters.install_path>>/mbx-ci
+            chmod 755 <<parameters.install_path>>/mbx-ci
```
macdrevx, line 660: Not convinced that the added complexity is worth it if we always use the defaults at the moment. Do you foresee a future in which we'll use this feature?

.circleci/config.yml
```diff
@@ -709,17 +730,23 @@ commands:
...
     steps:
       - run: pip3 install awscli gitpython requests python-dateutil
-      - run: brew install jq xcodegen
+      - run: brew install jq xcodegen gh
 
   make-docs:
     description: "Generate Documentation"
     steps:
       - run: git submodule update --init -- scripts/doc-generation/jazzy-theme
-      - run: cd scripts/doc-generation && ./generate-docs.sh
```
macdrevx, line 718: Are we ready to delete the old script files too?

.circleci/config.yml
```diff
@@ -709,17 +730,23 @@ commands:
...
+      - run: scripts/doc-generation/generate-maps-docs.sh
       - run: zip -r api-docs.zip api-docs
       - store_artifacts:
           path: api-docs.zip
 
+  open-docs-pull-requests:
+    steps:
+      - run: 
+          name: Create documentation Pull requests
```
macdrevx, line 747:
```suggestion
          name: Create documentation pull requests
```

scripts/release/create-api-downloads-pr.sh
```diff
@@ -14,18 +14,18 @@ VERSION=$2
 BRANCH_NAME="${PROJECT_ROOT}/${VERSION}"
 GITHUB_TOKEN=$(./scripts/release/get_token.js)
 
-TMPDIR=`mktemp -d`
+TMPDIR=$(mktemp -d)
 
-git clone https://x-access-token:$GITHUB_TOKEN@github.com/mapbox/api-downloads.git ${TMPDIR}
-cd ${TMPDIR}
+git clone https://x-access-token:$GITHUB_TOKEN@github.com/mapbox/api-downloads.git "${TMPDIR}"
+cd "${TMPDIR}" || exit 1
```
macdrevx, line 20:
```suggestion
cd "$TMPDIR" || exit 1
```

scripts/release/create-api-downloads-pr.sh
```diff
@@ -14,18 +14,18 @@ VERSION=$2
 BRANCH_NAME="${PROJECT_ROOT}/${VERSION}"
 GITHUB_TOKEN=$(./scripts/release/get_token.js)
 
-TMPDIR=`mktemp -d`
+TMPDIR=$(mktemp -d)
 
-git clone https://x-access-token:$GITHUB_TOKEN@github.com/mapbox/api-downloads.git ${TMPDIR}
-cd ${TMPDIR}
+git clone https://x-access-token:$GITHUB_TOKEN@github.com/mapbox/api-downloads.git "${TMPDIR}"
```
macdrevx, line 19:
```suggestion
git clone https://x-access-token:$GITHUB_TOKEN@github.com/mapbox/api-downloads.git "$TMPDIR"
```

scripts/release/create-api-downloads-pr.sh
```diff
@@ -14,18 +14,18 @@ VERSION=$2
...
+cd "${TMPDIR}" || exit 1
 echo "Checking out to ${TMPDIR}"
-git checkout -b ${BRANCH_NAME}
+git checkout -b "${BRANCH_NAME}"
```
macdrevx, line 22:
```suggestion
git checkout -b "$BRANCH_NAME"
```

macdrevx: there are more in this file. not going to call them all out

scripts/release/create-docs-prs.sh
```diff
@@ -0,0 +1,160 @@
...
+    step "Open mapbox/ios-sdk PR"
+    ios_sdk_open_pr
+
+    exit 0
+}
+
+maps_ios_upload_docs() {
+    info "Checkout publisher-staging worktree"
+    local staging_docs_path="$TMP_ROOT/docs-staging"
+    git worktree add "$staging_docs_path" "publisher-staging" --quiet
```
macdrevx, line 45: the fact that we're always using publisher-staging and force pushing it means we can never run more than 1 release at a time. that would have been a problem today when I was patching 10.0 and 10.1 simultaneously. Is there a way for us to make this work better when running releases in parallel?

scripts/release/create-docs-prs.sh
```diff
@@ -0,0 +1,160 @@
...
+    git_configure_release_user
+
+    git checkout -b "${IOS_SDK_BRANCH_NAME}" --quiet
+
+    # Apply only for release versions
+    if [[ "$VERSION" =~ ^v?[0-9]*\.[0-9]*\.[0-9]*$ ]]; then
+        info "Update release version"
+        # shellcheck disable=SC2005 # This solution is intentional to modify json in one line with jq
+        echo "$(jq ".VERSION_IOS_MAPS_SDK = \"$VERSION\"" src/constants.json)" > src/constants.json
+    fi
```
macdrevx, line 104: Should also never replace an existing version that's newer. This would come in to play with patch releases of older minor versions.

scripts/release/create-docs-prs.sh
```diff
@@ -0,0 +1,160 @@
...
+
+    info "Add version to the list"
+    # shellcheck disable=SC2005
+    echo "$(jq "[\"${VERSION}\"] + ." src/data/ios-maps-sdk-versions.json)" > src/data/ios-maps-sdk-versions.json
```
macdrevx, line 108: Does it matter if we keep the array sorted?

.circleci/config.yml
OdNairy: It's a template for future reuse orbs and so on. Let's keep it. Defaults are fine for Maps-iOS usage now.

scripts/release/create-docs-prs.sh
OdNairy: Ah, I see the problem. Do we expect it to be a problem in the future? Should we support parallel releases?
As a workaround I can make a version-based PRs but that means we would lost an option to preview docs in the staging env.

scripts/release/create-docs-prs.sh
OdNairy: Oh, nice catch! Would fix it

scripts/release/create-docs-prs.sh
OdNairy: But we do not enforce sorted array here, right?

scripts/release/create-docs-prs.sh
macdrevx: What if we do a hybrid — make version-based-branches for PRs, but also update publisher-staging to match those branches. That way, in the normal case, we still get previews, and in the parallel-releases case, latter PRs don't clobber previous ones.

scripts/release/create-docs-prs.sh
macdrevx: The array in that file is sorted at the moment. I believe it's the source for the list at https://docs.mapbox.com/ios/maps/api-reference/. If we scramble it, would the content on that page be scrambled too, or is there some sorting that happens when that array is consumed? @colleenmcginnis could you advise?

scripts/release/create-api-downloads-pr.sh
OdNairy: Done

scripts/release/create-api-downloads-pr.sh
OdNairy: Done

scripts/release/create-api-downloads-pr.sh
OdNairy: Done

.circleci/config.yml
OdNairy: Done

scripts/release/create-docs-prs.sh
OdNairy: Done

scripts/release/create-docs-prs.sh
OdNairy: I've added sorting logic (implemented by `sort -rV`) but I'm not sure is it okay.
Here is the changeset sample:
https://github.com/mapbox/ios-sdk/commit/f5f0f1609445000660e64ba1e7111e88419088fa?diff=split

colleenmcginnis: We have logic in the component that renders the versions on the page that will sort versions automatically so it doesn't matter what order versions are in in the array.

macdrevx: Thanks @colleenmcginnis. @OdNairy sounds like we don't need to worry about it!

macdrevx (approved):

PR #942, opened 2021-12-13
Update MapboxCommon to v20.0.3 and Prepare v10.0.3 Release

macdrevx:
MapboxCommon v20.0.3 fixes a billing issue when upgrading Mapbox Maps SDK from v6 to v10. This patch replaces the previous patch that was released (and subsequently recalled) as v10.0.2.


OdNairy (approved):

parrots:
@macdrevx can we get some documentation on what the billing issue was, and when customers should inspect their invoice to see if they were affected?

sbma44:
@parrots we identified an issue that could potentially manifest as incorrect mobile MAU counts during the first month in which a developer deployed one of the affected versions. After pulling the release we immediately reviewed every account that downloaded an affected version of the SDK to confirm that there was no impact on their invoices. You shouldn't have to check your invoice--I have just reconfirmed this--but I'd be happy to discuss your account specifics in more detail if you'd like (tlee at mapbox).

PR #961, opened 2021-12-16
One-touch automation for Examples App

ZiZasaurus:


fastlane/report.xml
```diff
@@ -0,0 +1,63 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
OdNairy, line 1: Do we need this final under version control?

fastlane/Appfile
```diff
@@ -0,0 +1,11 @@
+app_identifier "com.mapbox.examples" # The bundle identifier of your app
+apple_id "zizi.miles@mapbox.com"  # Your Apple email address
```
OdNairy, line 2: Is it possible to use `applemachine` account here?

fastlane/Fastfile
```diff
@@ -0,0 +1,83 @@
...
+# All lines starting with a # are ignored when running `fastlane`
+# If you want to automatically update fastlane if a new version is available:
+# update_fastlane
+# This is the minimum version number required.
+# Update this, if you use features of a newer version
+fastlane_version "2.120.0"
+default_platform :ios
+# Set default target for get_version_number()
+ENV["FL_VERSION_NUMBER_TARGET"] = "Examples"
+SPACESHIP_SKIP_2FA_UPGRADE=1
```
OdNairy, line 15: Should it be environment variable or fastlane is smart enough to process this expression?

fastlane/Fastfile
```diff
@@ -0,0 +1,83 @@
...
+platform :ios do
+  before_all do
+    # ENV["SLACK_URL"] = "https://hooks.slack.com/services/..."
+  end
```
OdNairy, line 19: Nit: remove this block

fastlane/Fastfile
```diff
@@ -0,0 +1,83 @@
...
+    )
+    deploy_to_testflight
+  end
+  desc "Submit a new Beta Build to Apple TestFlight"
+  desc "This will also make sure that the signing certificate and provisioning profiles are up to date."
+  lane :build_and_submit do
+    spm(
+      command: "update",
+      configuration: "release"
+    )
```
OdNairy, line 33: It looks like we don't need `swift package update` call here as there is no Package.swift manifest.

fastlane/Fastfile
```diff
@@ -0,0 +1,83 @@
...
+    match(type: "appstore", readonly: true)
```
OdNairy, line 34: I did help @Chaoba with firebase test lab setup during which we made a fastlane configuration. It was important that on CircleCI we have to call `setup_circle_ci` to fix the Match behavior. 
It also would enable readonly mode by default.

fastlane/Fastfile
```diff
@@ -0,0 +1,83 @@
...
+    gym(
+      scheme: "Examples",
+      workspace: "Apps/Apps.xcworkspace",
+      # project: "Apps/Examples/Examples.xcodeproj",
```
OdNairy, line 38: Nit: remove commented line

fastlane/Fastfile
```diff
@@ -0,0 +1,83 @@
...
+  desc "Build and submit a new beta build to Apple TestFlight."
+  lane :deploy_to_testflight do
+    build_and_submit
+  end
+  after_all do |lane|
+    # This block is called, only if the executed lane was successful
+    # slack(
+    #   message: "Successfully deployed new App Update."
+    # )
+  end
```
OdNairy, line 71: Should we remove that block?

fastlane/Appfile
ZiZasaurus: Absolutely will change and test.

fastlane/Fastfile
ZiZasaurus: From my testing, fastlane was smart enough to process the expression as is.

fastlane/Fastfile
ZiZasaurus: Yep! Will remove.

fastlane/Appfile
ZiZasaurus: Hmm. When I attempt with the applemachine account, I receive an unauthorized access error. Perhaps we can set it to grab the account of whoever is releasing.

fastlane/Fastfile
ZiZasaurus: Correction: I changed it to an environment variable.

fastlane/report.xml
ZiZasaurus: Nope, will remove

fastlane/Fastfile
```diff
@@ -0,0 +1,73 @@
...
+      xcodeproj: Dir.glob("#{ENV['PWD']}/Apps/Examples/Examples.xcodeproj").first
+    )
+    deploy_to_testflight
+  end
+  desc "Submit a new Beta Build to Apple TestFlight"
+  desc "This will also make sure that the signing certificate and provisioning profiles are up to date."
+  lane :build_and_submit do
+    spm(
+      configuration: "release"
+    )
```
OdNairy, line 29: I think we can totally remove that `spm` action.

fastlane/Fastfile
```diff
@@ -0,0 +1,73 @@
...
+    setup_circle_ci
+    match(readonly: true)
```
OdNairy, line 31:
```suggestion
    match
```

PR #964, opened 2021-12-16
Expose triggerRepaint API

ZiZasaurus:
Exposes the `triggerRepaint` API to allow manual map repainting. Please note that calling this method is typically unnecessary but may be needed if using a custom layer that needs to be redrawn independently of other map changes.

### DIFF
--- a/Sources/MapboxMaps/Foundation/MapboxMap.swift
+++ b/Sources/MapboxMaps/Foundation/MapboxMap.swift
@@ -59,6 +59,11 @@ public final class MapboxMap: MapboxMapProtocol {
 
     // MARK: - Style loading
 
+    /// Triggers a repaint of the map.
+    public func triggerRepaint() {
+        __map.triggerRepaint()
+    }
+
     private func observeStyleLoad(_ completion: @escaping (Result<Style, Error>) -> Void) {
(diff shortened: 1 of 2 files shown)


Sources/MapboxMaps/Foundation/MapboxMap.swift
```diff
@@ -59,6 +59,11 @@ public final class MapboxMap: MapboxMapProtocol {
 
     // MARK: - Style loading
 
+    /// Triggers a repaint of the map.
+    public func triggerRepaint() {
+        __map.triggerRepaint()
+    }
+
```
macdrevx, line 66:
```suggestion
    // MARK: - Render loop

    /// Triggers a repaint of the map. Calling this method is typically unnecessary but
    /// may be needed if using a custom layer that needs to be redrawn independently
    /// of other map changes.
    public func triggerRepaint() {
        __map.triggerRepaint()
    }

    // MARK: - Style loading

```

macdrevx:
@ZiZasaurus please update the PR description with some information to help folks who visit this PR in the future from the changelog or release notes understand the change better and whether it pertains to them.

Also, please add an entry to the changelog.

macdrevx (approved):

macdrevx (approved):

PR #1040, opened 2022-01-19
Viewport API Refinements

macdrevx:
## Pull request checklist:
 - [x] Update the changelog

### Summary of changes

* Viewport transition registration APIs have been replaced with an optional transition argument in `Viewport.transition(to:transition:completion:)` which defaults to nil. Passing nil indicates that the value of `Viewport.defaultTransition` should be used.
* `fromState` has been removed from `ViewportStatus.transition` and `ViewportTransition.run(to:completion:)`
* `ViewportStatus.state`'s associated value is now non-optional and a new `ViewportStatus.idle` case has been added to represent idle.
* `FollowingViewportState` has been renamed to `FollowPuckViewportState` to better communicate the scope of its functionality.
* All public viewport APIs have been annotated with `@_spi(Experimental)`. While these APIs are considered ready for production, this annotation is a signal that we reserve the right to make breaking API changes outside of a major version bump.
* The viewport factory methods have been tested
* Random.swift has been split up into multiple files and some other tests and test helpers have been reorganized.
* Deletes flaky testDataClearing


evil159 (approved):
Nice changes 👍 

One note that for extra neatness we could have those `.random()` functions as properties(e.g. `.random`). But that's probably also subject for a debate, and even if not just a very minor thing.

OdNairy (approved):
👍

PR #1058, opened 2022-01-24
Improve support for custom viewport gestures

macdrevx:
## Pull request checklist:
 - [x] Update the changelog

### Summary of changes

* Introduces a 0.15 second delay before disabling animations when
  interacting with the map. This allows short discrete gestures like
  the double tap, double touch, and single tap gestures to be performed
  without canceling animations. It also makes repeated use of double tap
  or double touch gestures more smooth since the animations are not
  interrupted in between each gesture.
* Updates AnyTouchGestureRecognizer with a configurable delay.
* Updates ViewportImpl to depend on the same instance of
  AnyTouchGestureRecognizer that is used to disable animations. Also
  updates ViewportImpl to depend on the double touch and double tap
  gesture recognizers. When its transitionsToIdleUponUserInteraction
  option is true, ViewportImpl will now idle when any of those three
  gestures is recognized. This design will also allow quick single-tap
  gestures that are performed more quickly than the 0.15 s threshold to
  avoid idling the viewport.
* ViewportOptions.transitionsToIdleUponUserInteraction now controls
  whether the AnyTouchGestureRecognizer is enabled. This means that
  when it is false, animations will not be disabled automatically when
  a gesture begins. In this situation, it will be the developer's responsibility
  to manage animation cancellation as needed for the specific custom
  gestures being implemented.
* Together, these changes provide improved ability to implement custom
  gestures for viewport states. For example, by setting the
  ViewportOptions.transitionsToIdleUponUserInteraction flag to false and
  setting nil for the zoom of the FollowPuckViewportState, the zoom
  gestures can be performed without causing the viewport to idle.

### DIFF
--- a/Sources/MapboxMaps/Foundation/MapViewDependencyProvider.swift
+++ b/Sources/MapboxMaps/Foundation/MapViewDependencyProvider.swift
@@ -8,9 +8,10 @@ internal protocol MapViewDependencyProviderProtocol: AnyObject {
                             cameraAnimationsManager: CameraAnimationsManagerProtocol) -> GestureManager
     func makeLocationProducer(mayRequestWhenInUseAuthorization: Bool) -> LocationProducerProtocol
     func makeLocationManager(locationProducer: LocationProducerProtocol, style: StyleProtocol) -> LocationManager
-    func makeViewportImpl(mapboxMap: MapboxMapProtocol,
+    func makeViewportImpl(view: UIView,
+                          mapboxMap: MapboxMapProtocol,
                           cameraAnimationsManager: CameraAnimationsManagerProtocol,
-                          idleGestureRecognizer: UIGestureRecognizer) -> ViewportImplProtocol
+                          idleGestureRecognizers: [UIGestureRecognizer]) -> ViewportImplProtocol
 }
 
 internal final class MapViewDependencyProvider: MapViewDependencyProviderProtocol {
@@ -93,7 +94,9 @@ internal final class MapViewDependencyProvider: MapViewDependencyProviderProtoco
     func makeAnimationLockoutGestureHandler(view: UIView,
                                             mapboxMap: MapboxMapProtocol,
                                             cameraAnimationsManager: CameraAnimationsManagerProtocol) -> GestureHandler {
-        let gestureRecognizer = AnyTouchGestureRecognizer()
+        let gestureRecognizer = ContinuousAnyTouchGestureRecognizer(
+            minimumPressDuration: 0.15,
+            timerProvider: TimerProvider())
         view.addGestureRecognizer(gestureRecognizer)
         return AnimationLockoutGestureHandler(
             gestureRecognizer: gestureRecognizer,
@@ -162,9 +165,14 @@ internal final class MapViewDependencyProvider: MapViewDependencyProviderProtoco
             puckManager: puckManager)
     }
 
-    func makeViewportImpl(mapboxMap: MapboxMapProtocol,
+    func makeViewportImpl(view: UIView,
+                          mapboxMap: MapboxMapProtocol,
                           cameraAnimationsManager: CameraAnimationsManagerProtocol,
-                          idleGestureRecognizer: UIGestureRecognizer) -> ViewportImplProtocol {
+                          idleGestureRecognizers: [UIGestureRecognizer]) -> ViewportImplProtocol {
+        let gestureRecognizer = DiscreteAnyTouchGestureRecognizer(
+            minimumPressDuration: 0.15,
+            timerProvider: TimerProvider())
+        view.addGestureRecognizer(gestureRecognizer)
         return ViewportImpl(
             options: .init(),
             mainQueue: MainQueue(),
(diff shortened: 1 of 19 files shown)


Sources/MapboxMaps/Gestures/GestureOptions.swift
```diff
@@ -73,5 +73,13 @@ public struct GestureOptions: Equatable {
     /// Defaults to `UIScrollView.DecelerationRate.normal.rawValue`
     public var panDecelerationFactor: CGFloat = UIScrollView.DecelerationRate.normal.rawValue
 
+    /// Whether animations (existing and new) are automatically canceled while a gesture is
+    /// in progress. Cancellation begins after a short delay to allow short, discrete gestures
+    /// like double tap and double touch to be used repeatedly without interruption. Set this
+    /// flag to `false` in situations where a gestures should be allowed to run in parallel
+    /// with animations (e.g. when an animation is modifying the camera center and a
+    /// gesture should be allowed to modify zoom). Defaults to `true`.
+    public var disableAnimationsDuringGestures: Bool = true
```
macdrevx, line 82: here's the new public gesture option

evil159 (approved):

Sources/MapboxMaps/Gestures/GestureOptions.swift
pengdev: do you have an example on how to use this option together with the viewport to customise gestures?

Sources/MapboxMaps/Foundation/MapViewDependencyProvider.swift
```diff
@@ -93,7 +94,9 @@ internal final class MapViewDependencyProvider: MapViewDependencyProviderProtoco
     func makeAnimationLockoutGestureHandler(view: UIView,
                                             mapboxMap: MapboxMapProtocol,
                                             cameraAnimationsManager: CameraAnimationsManagerProtocol) -> GestureHandler {
-        let gestureRecognizer = AnyTouchGestureRecognizer()
+        let gestureRecognizer = ContinuousAnyTouchGestureRecognizer(
+            minimumPressDuration: 0.15,
```
pengdev, line 98: How is the 0.15 defined? is it the threshold that we recognise the double tap gesture?

Sources/MapboxMaps/Gestures/GestureOptions.swift
pengdev: Also do you feel this flag redundant with the `transitionToIdleUponUserInteraction` flag in the viewport? e.g `transitionToIdleUponUserInteraction = false` must go with `disableAnimationsDuringGestures = false`?

OdNairy (approved):

Sources/MapboxMaps/Foundation/MapViewDependencyProvider.swift
macdrevx: It's how long touches must remain on the screen before animations will begin to be cancelled.

Sources/MapboxMaps/Gestures/GestureHandlers/AnimationLockoutGestureHandler.swift
```diff
@@ -6,7 +6,6 @@ internal final class AnimationLockoutGestureHandler: GestureHandler {
 
     internal init(gestureRecognizer: UIGestureRecognizer,
                   cameraAnimationsManager: CameraAnimationsManagerProtocol) {
-        gestureRecognizer.cancelsTouchesInView = false
```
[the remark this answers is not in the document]
macdrevx: Moved it to the initialization of the gesture recognizer.

Sources/MapboxMaps/Gestures/GestureOptions.swift
macdrevx: after discussion: we'll remove this option and control the behavior using the viewport option

macdrevx: Updated

Sources/MapboxMaps/Foundation/MapViewDependencyProvider.swift
```diff
@@ -162,9 +165,14 @@ internal final class MapViewDependencyProvider: MapViewDependencyProviderProtoco
...
     }
 
-    func makeViewportImpl(mapboxMap: MapboxMapProtocol,
+    func makeViewportImpl(view: UIView,
+                          mapboxMap: MapboxMapProtocol,
                           cameraAnimationsManager: CameraAnimationsManagerProtocol,
-                          idleGestureRecognizer: UIGestureRecognizer) -> ViewportImplProtocol {
+                          idleGestureRecognizers: [UIGestureRecognizer]) -> ViewportImplProtocol {
+        let gestureRecognizer = DiscreteAnyTouchGestureRecognizer(
+            minimumPressDuration: 0.15,
```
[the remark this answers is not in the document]
macdrevx: Following refactoring, it's no longer repeated in multiple places. I actually prefer it inline since it lets you see the context of how it's used without searching.

I will add a comment to describe why I picked 0.15 though.

PR #1066, opened 2022-01-26
Update pull request template

macdrevx:
Trying to streamline PR creation…

* CLA info moved into PR checklist.
* Separated documentation comment checklist item from guides/markdown files since it's more commonly needed.
* Removed checklist item about visuals since it's covered in the comment under Description of changes.
* Removed the suggestion that the PR description should be brief. Brevity hasn't been a problem. If anything, I'd like a bit more thoroughness when describing changes so that we have a better record for future reference.
* Removed the changelog label and changelog entry mechanism. We're not using it, and we just want folks to be adding changelog entries with each PR right now.
* Noted that guides are only available to employees
* Removed redundant User impact section

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -5,32 +5,20 @@ Please fill out the sections below to complete your submission.
 
 We appreciate your contributions!
 -->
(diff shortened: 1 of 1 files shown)


OdNairy (approved):
Can we move `Description of changes` to the top of the Template? I think it's more sense to read meanful description that SLA agreement

macdrevx:
Updated. Removed the heading and combined the `Fixes:` into the comment as well. This will make the template work better with GitHub's behavior that prepends the commit message body to the template content when opening a PR with a single commit.

Issue #1053, opened 2022-01-23
Map content issues after opened the app from background after a couple of hours


nastasiupta:
## Environment

- Xcode version: -
- iOS version: 15.2.1
- Devices affected: 13 Pro Max
- Maps SDK Version: 2.2.0-rc1

## Observed behavior and steps to reproduce

So, I sent an app update in TestFlight and one of the team members moved the app in the background. After a couple of hours he opened, he opened the app, and the map looks as you see in the screenshot above. He had to kill the app and reopen to get to normal

## Expected behavior

Look normal...
![Screenshot 2022-01-23 at 14 47 10](https://user-images.githubusercontent.com/13376064/150679157-f04711a0-c4f6-4e7c-8221-e31947fd5e4f.png)

RodrigoSAlves:
Hey @nastasiupta,

This also happened to two testers in our QA Team, with similar scenarios. They left the app in background in between workdays and when they opened the app again, they experienced this pink color covering many of the map's features.
They also mentioned that after some zooming in / out the tile imagery managed to fix itself, but some of the layers continued to be covered by that pink color and necessitated an app restart in order for them to appear correctly again.
Here's a screenshot we got from them:
![image_1](https://user-images.githubusercontent.com/19390670/150760989-cbb5a272-f93a-49d3-aab3-cf7d6ab862d8.jpeg)

We were actually spinning our wheels trying to figure out what was going on. We'll await feedback from the Mapbox team to see if they have any tips in for fixing this.

ZiZasaurus:
@nastasiaupta thank you for reporting. We will look into what may be causing this behavior.

macdrevx:
@nastasiupta @RodrigoSAlves we made a change in 10.4.0-beta.1 that may resolve this issue. Please try it out and let us know if you are still able to reproduce this bug.

PR #1086, opened 2022-02-03
Pause rendering when entering background

### LINKED ISSUE #978: Purple rectangle layer after a background
## Environment

- Xcode version: 13.2
- iOS version: 15.2
- Devices affected: iPhone 13 pro
- Maps SDK Version: 10.2

After awhile in the background, when I put back the app in foreground. All layers are represented with purple rectangle.
I got the same behavior in the latest mapbox navigation

## Expected behavior

My annotations should be back

287A6EC2-600B-4808-911C-91F48D8D090C

evil159:
`MapView` will now listen to either `UIApplication` or `UIScene` lifecycle event notifications(depending on project setup). As application/scene moves to background/foreground, map view's `displayLink` will be paused/resumed accordingly. 

On iOS 11 and 12 `UIApplication` is the only source of these events. Starting from iOS 13 Scene API can be enabled by including [UIApplicationSceneManifest](https://developer.apple.com/documentation/bundleresources/information_property_list/uiapplicationscenemanifest) key to `Info.plist`. 

Additionally this PR adds `reduceMemoryUse()` call when transitioning to background as its documentation advises.

Fixes #978, #1053 

## Pull request checklist:


Sources/MapboxMaps/Foundation/Extensions/UIWindow.swift
```diff
@@ -0,0 +1,9 @@
+import UIKit
+
+@available(iOS 13.0, *)
+extension UIWindow {
+    internal var parentScene: UIScene? {
+        // falling back to ``CPWindow.templateApplicationScene`` to also cover CarPlay
+        return windowScene ?? self.value(forKeyPath: "templateApplicationScene") as? UIScene
```
evil159, line 7: This was a tricky one - we don't really know at the compile time if CarPlay is available. So it seems that resorting to runtime checks is the way to go. 

UIKit provides us only with `windowScene`, which is `nil` in case of CarPlay window. `CPWindow` has a separate scene property called `templateApplicationScene`. 
Thus the return value of this computed property will be `windowScene` in case of `UIWindow`, and `templateApplicationScene` in case of `CPWindow`. 

I'm open to suggestions on how to improve this :)

Sources/MapboxMaps/Foundation/Extensions/UIWindow.swift
```diff
@@ -0,0 +1,16 @@
+import UIKit
+
+@available(iOS 13.0, *)
+extension UIWindow {
+    private static let templateApplicationSceneSelector = Selector(("templateApplicationScene"))
+
+    internal var parentScene: UIScene? {
```
evil159, line 7: This was a tricky one - we don't really know at the compile time if CarPlay is available. So it seems that resorting to runtime checks is the way to go. 

UIKit provides us only with `windowScene`, which is `nil` in case of CarPlay window. `CPWindow` has a separate scene property called `templateApplicationScene`. 
With the help of Objective C runtime we can access here the car play scene(if any).
Thus the return value of this computed property will be `windowScene` in case of `UIWindow`, and `templateApplicationScene` in case of `CPWindow`. 

I'm open to suggestions on how to improve this :)

Sources/MapboxMaps/Foundation/Extensions/UIWindow.swift
```diff
@@ -0,0 +1,16 @@
...
+        var carPlayScene: UIScene? {
+            guard self.responds(to: UIWindow.templateApplicationSceneSelector) else {
+                return nil
+            }
+            return self.value(forKey: UIWindow.templateApplicationSceneSelector.description) as? UIScene
+        }
+        return windowScene ?? carPlayScene
+    }
```
evil159, line 15: This was a tricky one - we don't really know at the compile time if CarPlay is available. So it seems that resorting to runtime checks is the way to go. 

UIKit provides us only with `windowScene`, which is `nil` in case of CarPlay window. `CPWindow` has a separate scene property called `templateApplicationScene`. 
With the help of Objective C runtime we can access here the car play scene(if any).
Thus the return value of this computed property will be `windowScene` in case of `UIWindow`, and `templateApplicationScene` in case of `CPWindow`. 

I'm open to suggestions on how to improve this :)

Tests/MapboxMapsTests/Foundation/MapViewTests.swift
```diff
@@ -187,4 +187,15 @@ final class MapViewTests: XCTestCase {
...
+        mapView.willEnterForeground()
+
+        XCTAssertFalse(displayLink.isPaused)
+    }
+
+    func testDidEnterBackground() {
+        mapView.didEnterBackground()
+
+        XCTAssertTrue(displayLink.isPaused)
+    }
```
macdrevx, line 200: You can stub the isPaused setter to enable verifying that the setter was actually invoked vs. the property just accidentally having the expected value already.

Sources/MapboxMaps/Foundation/MapView+Lifecycle.swift
```diff
@@ -0,0 +1,89 @@
+import Foundation
+import UIKit
+
+@available(iOSApplicationExtension, unavailable)
+extension MapView {
```
macdrevx, line 5: Let's put this in the main MapView implementation so that we can keep more of its interface private.

Sources/MapboxMaps/Foundation/Extensions/UIWindow.swift
```diff
@@ -0,0 +1,16 @@
+import UIKit
+
+@available(iOS 13.0, *)
+extension UIWindow {
+    private static let templateApplicationSceneSelector = Selector(("templateApplicationScene"))
```
macdrevx, line 5: Let's declare constants in the narrowest scope possible. In this case, this can just be a local let constant in carPlayScene.

Sources/MapboxMaps/Foundation/Extensions/UIWindow.swift
```diff
@@ -0,0 +1,16 @@
+import UIKit
```
macdrevx, line 1: Can we rename this file to `UIWindow+ParentScene.swift`

Sources/MapboxMaps/Foundation/Extensions/UIWindow.swift
macdrevx:
```
    internal var parentScene: UIScene? {
        return windowScene ?? (self as? CPWindow)?.templateApplicationScene
    }
```

Does this work?

macdrevx: It's fine to import CarPlay - the framework will be weak-linked automatically if needed based on the deployment target, and we require Xcode 12.5 or later to build, so CarPlay will always be available when linking.

Sources/MapboxMaps/Foundation/MapView+Lifecycle.swift
```diff
@@ -0,0 +1,89 @@
...
+    internal func subscribeToMemoryWarningNotification() {
+        NotificationCenter.default.addObserver(self,
+                                               selector: #selector(didReceiveMemoryWarning),
+                                               name: UIApplication.didReceiveMemoryWarningNotification,
+                                               object: nil)
+    }
+
+    internal func subscribeToLifecycleNotifications() {
+        if Bundle.main.object(forInfoDictionaryKey: MapView.UIApplicationSceneManifestKey) != nil {
+            if #available(iOS 13.0, *) {
```
macdrevx, line 17: What happens if there's an app that uses scenes in iOS 13+ and falls back to older tech when running on older iOS versions?

Sources/MapboxMaps/Foundation/MapView+Lifecycle.swift
```diff
@@ -0,0 +1,89 @@
...
+
+        switch notification.name {
+        case UIScene.willEnterForegroundNotification:
+            willEnterForeground()
+        case UIScene.didEnterBackgroundNotification:
+            didEnterBackground()
+        default:
+            break
+        }
+    }
```
macdrevx, line 65: Unused?

Sources/MapboxMaps/Foundation/MapView+Lifecycle.swift
```diff
@@ -0,0 +1,89 @@
...
+    internal func unsubscribeFromLifecycleNotifications() {
+        if Bundle.main.object(forInfoDictionaryKey: MapView.UIApplicationSceneManifestKey) != nil {
+            if #available(iOS 13.0, *) {
+                NotificationCenter.default.removeObserver(self, name: UIScene.willEnterForegroundNotification, object: nil)
+                NotificationCenter.default.removeObserver(self, name: UIScene.didEnterBackgroundNotification, object: nil)
+            }
+        } else {
+            NotificationCenter.default.removeObserver(self, name: UIApplication.willEnterForegroundNotification, object: nil)
+            NotificationCenter.default.removeObserver(self, name: UIApplication.didEnterBackgroundNotification, object: nil)
+        }
```
macdrevx, line 34: This can be simplified. No need to check for UIApplicationSceneManifestKey. Just unsubscribe from all 4 (or 2 if < iOS 13) every time.

Sources/MapboxMaps/Foundation/MapView.swift
```diff
@@ -274,6 +269,8 @@ open class MapView: UIView {
 
         // Set up managers
         setupManagers()
+
+        subscribeToMemoryWarningNotification()
```
macdrevx, line 273: I would revert this change. We only do this once, so it doesn't seem necessary to create a method for it. Just leave it inline as it was before.

macdrevx (changes requested):
Looking good. Can you also make the change to how the display link is initialized?

Tests will need to be more thorough if possible:
* verify notifications are registered/unregistered correctly
* unit test UIWindow.parentScene

Sources/MapboxMaps/Foundation/Extensions/UIWindow.swift
evil159: Good tip, looks much better now!

Tests/MapboxMapsTests/Foundation/MapViewTests.swift
evil159: Stubbed

Sources/MapboxMaps/Foundation/MapView+Lifecycle.swift
evil159: Moved it back

Sources/MapboxMaps/Foundation/MapView+Lifecycle.swift
evil159: Well spotted! Fixed it.

Sources/MapboxMaps/Foundation/MapView+Lifecycle.swift
evil159: 😳

Sources/MapboxMaps/Foundation/MapView+Lifecycle.swift
evil159: Good point, indeed.

Sources/MapboxMaps/Foundation/MapView.swift
evil159: Reverted

Tests/MapboxMapsTests/Foundation/Mocks/MockNotificationCenter.swift
```diff
@@ -0,0 +1,36 @@
+import Foundation
+@testable import MapboxMaps
+
+final class MockNotificationCenter: NotificationCenterProtocol {
```
macdrevx, line 4: Is it worth it to mock notification center? NotificationCenter.default delivers notifications synchronously already, and if we verify that the expected behaviors happen when a notification is posted, we'll know that the observer was added correctly.

Sources/MapboxMaps/Foundation/MapView.swift
```diff
@@ -164,7 +167,7 @@ open class MapView: UIView {
     }
 
     deinit {
-        NotificationCenter.default.removeObserver(self)
+        notificationCenter.removeObserver(self)
     }
```
macdrevx, line 171: We should just delete this. NotificationCenter keeps weak refs to observers on every OS version we support.

Sources/MapboxMaps/Foundation/MapView.swift
```diff
@@ -326,6 +335,74 @@ open class MapView: UIView {
             mapboxMap: mapboxMap)
     }
 
+    private func subscribeToLifecycleNotifications() {
+        if bundle.infoDictionary?["UIApplicationSceneManifest"] != nil {
+            if #available(iOS 13.0, *) {
```
macdrevx, line 340: You can combine these into a single if statement and avoid the duplication in the else blocks.

Sources/MapboxMaps/Foundation/MapView.swift
```diff
@@ -326,6 +335,74 @@ open class MapView: UIView {
...
+    private func unsubscribeFromLifecycleNotifications() {
+        if #available(iOS 13.0, *) {
+            notificationCenter.removeObserver(self, name: UIScene.willEnterForegroundNotification, object: nil)
+            notificationCenter.removeObserver(self, name: UIScene.didEnterBackgroundNotification, object: nil)
+        }
+        notificationCenter.removeObserver(self, name: UIApplication.willEnterForegroundNotification, object: nil)
+        notificationCenter.removeObserver(self, name: UIApplication.didEnterBackgroundNotification, object: nil)
+    }
+
+    @objc private func didReceiveNotification(_ notification: Notification) {
```
macdrevx, line 380: Can we name this more specifically?

macdrevx: Hmm, I guess I'm saying I'd rather have separate 3 separate handlers for foreground, background, and memory warnings.

Sources/MapboxMaps/Foundation/MapView.swift
```diff
@@ -430,7 +507,9 @@ open class MapView: UIView {
 
         if window != nil {
             validateDisplayLink()
+            subscribeToLifecycleNotifications()
         } else {
+            unsubscribeFromLifecycleNotifications()
             // TODO: Fix this up correctly.
```
macdrevx, line 513: You're going to have conflicts with your other PR. Do you have a preference which one to merge first?

Tests/MapboxMapsTests/Foundation/Extensions/UIWindow+ParentSceneTests.swift
```diff
@@ -0,0 +1,20 @@
...
+import CarPlay
+@testable import MapboxMaps
+
+final class UIWindowParentSizeTests: XCTestCase {
+
+    @available(iOS 13.0, *)
+    func testCarPlayWindowReturnsCorrectParentScene() {
+        let window = CPWindow()
+
+        XCTAssertEqual(window.parentScene, window.templateApplicationScene)
```
macdrevx, line 11: What are the actual values? Are they both nil? If so, that might not be a very good test.

Tests/MapboxMapsTests/Foundation/Extensions/UIWindow+ParentSceneTests.swift
```diff
@@ -0,0 +1,20 @@
...
+    }
+
+    @available(iOS 13.0, *)
```
macdrevx, line 14: Please make sure the tests pass on OS versions < 13

macdrevx: I vaguely recall this not working in tests and needing to do runtime checks in the test implementation.

macdrevx: But test it to make sure I'm not misremembering.

Tests/MapboxMapsTests/Foundation/Extensions/UIWindow+ParentSceneTests.swift
```diff
@@ -0,0 +1,20 @@
...
+    func testUIWindowReturnsCorrectParentScene() {
+        let window = UIWindow()
+
+        XCTAssertEqual(window.parentScene, window.windowScene)
```
macdrevx, line 18: Same question here about the actual values.

Tests/MapboxMapsTests/Foundation/MapViewTests.swift
```diff
@@ -4,6 +4,8 @@ import XCTest
 final class MapViewTests: XCTestCase {
 
     var displayLink: MockDisplayLink!
+    var notificationCenter: MockNotificationCenter!
+    var bundle: MockBundle!
```
macdrevx, line 8: Please nil these in tearDown

Tests/MapboxMapsTests/Foundation/MapViewTests.swift
```diff
@@ -182,9 +188,113 @@ final class MapViewTests: XCTestCase {
...
+        let bundle = MockBundle()
+        bundle.infoDictionaryStub.defaultReturnValue = [:]
+        dependencyProvider.makeNotificationCenterStub.defaultReturnValue = notificationCenter
+        dependencyProvider.makeBundleStub.defaultReturnValue = bundle
+        let mapView = MapView(
+            frame: CGRect(origin: .zero, size: CGSize(width: 100, height: 100)),
+            mapInitOptions: MapInitOptions(),
+            dependencyProvider: dependencyProvider)
+
+        window.addSubview(mapView)
```
macdrevx, line 202: Is this necessary? seems to duplicate what's already happening in setUp

macdrevx: Similarly in the other tests.

Tests/MapboxMapsTests/Foundation/MapViewTests.swift
```diff
@@ -182,9 +188,113 @@ final class MapViewTests: XCTestCase {
...
+
+        XCTAssertTrue(notificationCenter.addObserverStub.invocations.contains(where: { stub in
+            stub.parameters.name == UIApplication.didEnterBackgroundNotification
+        }))
```
macdrevx, line 206: I'd name these arguments `invocation` instead of `stub` to match the type.

Tests/MapboxMapsTests/Foundation/MapViewTests.swift
```diff
@@ -182,9 +188,113 @@ final class MapViewTests: XCTestCase {
...
+            dependencyProvider: dependencyProvider)
+
+        window.addSubview(mapView)
+
+        XCTAssertTrue(notificationCenter.addObserverStub.invocations.contains(where: { stub in
+            stub.parameters.name == UIScene.didEnterBackgroundNotification
+        }))
+        XCTAssertTrue(notificationCenter.addObserverStub.invocations.contains(where: { stub in
+            stub.parameters.name == UIScene.willEnterForegroundNotification
+        }))
```
macdrevx, line 235: We need to verify the `object` as well as the `name` here since an app may contain multiple scenes.

Tests/MapboxMapsTests/Foundation/MapViewTests.swift
```diff
@@ -182,9 +188,113 @@ final class MapViewTests: XCTestCase {
...
+
+        XCTAssertEqual(displayLink.$isPaused.setStub.parameters, [false])
     }
 
+    @available(iOS 13.0, *)
+    func testDisplayLinkResumedWhenSceneMovingToForeground() {
+        bundle.infoDictionaryStub.defaultReturnValue = ["UIApplicationSceneManifest": []]
+        mapView.didMoveToWindow()
+
+        notificationCenter.post(name: UIScene.willEnterForegroundNotification, object: window.parentScene)
```
macdrevx, line 286: Is parentScene nil?

macdrevx (commented):
Still on the right track! Let me know if you want to talk through any of the feedback.

Tests/MapboxMapsTests/Foundation/Mocks/MockNotificationCenter.swift
evil159: I think the advantage of Mocking Notification Center is in that allows us to directly verify when subscription/unsubscription happens, instead of relying on indirects signs.

Tests/MapboxMapsTests/Foundation/Extensions/UIWindow+ParentSceneTests.swift
evil159: `windowScene` seems to be a special property as it gets populated even when creating a window directly. So in this case bot `parentScene` and `windowScene` won't be nil.

Sources/MapboxMaps/Foundation/MapView.swift
evil159: This one for sure, the other PR is smaller and should be easier to resolve conflicts in.

Sources/MapboxMaps/Foundation/MapView.swift
evil159: TIL 😲

Sources/MapboxMaps/Foundation/MapView.swift
evil159: Deleted.

Tests/MapboxMapsTests/Foundation/Extensions/UIWindow+ParentSceneTests.swift
evil159: Good point, indeed it just crashes when running e.g. on iOS 12 with these annotations. Converted them to runtime checks.

Sources/MapboxMaps/Foundation/MapView.swift
```diff
@@ -326,6 +331,64 @@ open class MapView: UIView {
...
+        displayLink?.isPaused = false
+    }
+
+    @objc private func appDidEnterBackground() {
+        displayLink?.isPaused = true
+    }
+
+    @available(iOS 13.0, *)
+    @objc private func sceneWillEnterForeground(_ notification: Notification) {
+        guard notification.object as? UIScene == window?.parentScene else { return }
```
macdrevx, line 375: I don't think these checks are necessary since we only subscribe for a specific object.

macdrevx: Removing them would allow you to share handlers with the app lifecycle notifications

evil159: I added them here to protect against a situation when map view moves to a window without scene attached(there is no clear documentation on specifics of this). So map view will subscribe to lifecycle notifications with `nil` object. And this supposed to be a safeguard against it.

Tests/MapboxMapsTests/Foundation/Extensions/UIWindow+ParentSceneTests.swift
macdrevx: Not for this PR, but I wonder whether we could set up a test suite that runs in a host app with CarPlay entitlements.

Sources/MapboxMaps/Foundation/MapView.swift
macdrevx: So the scenario is, on iOS 13+ with scenes enabled, map view moves to a window that has a nil scene?

In that case, we would basically not get pausing and resuming until the window's scene became non-nil, right?

Not sure if this scenario is even possible.

macdrevx (approved):
Approving, but can you take care of one more thing before merging? https://github.com/mapbox/mapbox-maps-ios/pull/1086#discussion_r801762176

Sources/MapboxMaps/Foundation/MapView.swift
evil159: Yeah, in this scenario we won't get pausing and resuming, it either this or pausing and resuming by notifications from a random scene(if there would be no checks). But this scenario is definitely in the realm of possibility - windowScene is optional readwrite property.

Tests/MapboxMapsTests/Foundation/Extensions/UIWindow+ParentSceneTests.swift
evil159: Good point, this will enable for tests like this to be run properly, let's get back to it next time we touch CarPlay related stuff.

Sources/MapboxMaps/Foundation/MapView.swift
```diff
@@ -326,6 +331,64 @@ open class MapView: UIView {
...
+                                           object: window?.parentScene)
+            notificationCenter.addObserver(self,
+                                           selector: #selector(sceneDidEnterBackground(_:)),
+                                           name: UIScene.didEnterBackgroundNotification,
+                                           object: window?.parentScene)
+        } else {
+            notificationCenter.addObserver(self,
+                                           selector: #selector(appWillEnterForeground),
+                                           name: UIApplication.willEnterForegroundNotification,
+                                           object: nil)
```
1ec5, line 348: Per mapbox/mapbox-gl-native#1869, this may be too early for mbgl to wake up and start rendering again:

> UIApplicationWillEnterForegroundNotification is posted before UIApplicationDidBecomeActiveNotification, while the application is still in the background.

/ref mapbox/mapbox-gl-native#1821

Sources/MapboxMaps/Foundation/MapView.swift
```diff
@@ -326,6 +331,64 @@ open class MapView: UIView {
             mapboxMap: mapboxMap)
     }
 
+    private func subscribeToLifecycleNotifications() {
+        if #available(iOS 13.0, *), bundle.infoDictionary?["UIApplicationSceneManifest"] != nil {
+            notificationCenter.addObserver(self,
+                                           selector: #selector(sceneWillEnterForeground(_:)),
+                                           name: UIScene.willEnterForegroundNotification,
+                                           object: window?.parentScene)
```
1ec5, line 339: As with `UIApplication.willEnterForegroundNotification`, `UIScene.willEnterForegroundNotification` is probably premature as well. `UIScene.didActivateNotification` would be more reliable.

1ec5: #1402 replaces this with `UIApplication.didBecomeActiveNotification`.

Sources/MapboxMaps/Foundation/MapView.swift
1ec5: #1402 replaces this with `UIScene.didActivateNotification`.

PR #1089, opened 2022-02-04
Remove pitch gesture changed angle requirements

macdrevx:
* Once the pitch gesture begins to recognize, there's no value in
  continuing to restrict the angle between the touches or the
  translation angle. Doing so has undesired consequence of making it
  look like the map is freezing and unfreezing when the translation
  or touch angle is crossing their respective thresholds.

## Pull request checklist:

### DIFF
--- a/Sources/MapboxMaps/Gestures/GestureHandlers/PitchGestureHandler.swift
+++ b/Sources/MapboxMaps/Gestures/GestureHandlers/PitchGestureHandler.swift
@@ -45,16 +45,10 @@ internal final class PitchGestureHandler: GestureHandler, UIGestureRecognizerDel
             }
 
             let translation = gestureRecognizer.translation(in: view)
-            let translationAngle = angleOfLine(from: .zero, to: translation)
-
-            // If the angle between the touch locations is less than the maximum
-            // AND the translation angle is more than 60 degrees, update the pitch.
-            if touchAngleIsLessThanMaximum(for: gestureRecognizer), abs(translationAngle) > 60 {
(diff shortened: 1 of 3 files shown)


Sources/MapboxMaps/Gestures/GestureHandlers/PitchGestureHandler.swift
```diff
@@ -45,16 +45,10 @@ internal final class PitchGestureHandler: GestureHandler, UIGestureRecognizerDel
             }
 
             let translation = gestureRecognizer.translation(in: view)
-            let translationAngle = angleOfLine(from: .zero, to: translation)
-
-            // If the angle between the touch locations is less than the maximum
-            // AND the translation angle is more than 60 degrees, update the pitch.
-            if touchAngleIsLessThanMaximum(for: gestureRecognizer), abs(translationAngle) > 60 {
```
aleksigron, line 52: Is there a chance that single-touch gestures are now being recognized since we don't call `touchAngleIsLessThanMaximum` any more, where we checked that number of touches was 2?

OdNairy (approved):

Sources/MapboxMaps/Gestures/GestureHandlers/PitchGestureHandler.swift
macdrevx: That should be prevented by lines 12 and 13. The check in `touchAngleIsLessThanMaximum` is more relevant *after* the gesture begins to recognize as a safeguard to avoid index out of bounds errors when obtaining touches on lines 26 and 27.

PR #1090, opened 2022-02-06
Sync viewport and puck animations

macdrevx:
Previously, the puck position was updated via a 0.5 second animation implemented in `MapboxCoreMaps`. The animation used an ease-in-out timing curve that was not configurable. This approach clashed with the `FollowPuckViewportState`, which updated the camera to match a new location with a 1 second animation with a linear timing curve.

This PR iterates on this situation by driving both animations from the Maps SDK for iOS. To achieve this, location updates are interpolated linearly over a 1 second duration by hooking into `MapView`'s `CADisplayLink`. Pucks (2D and 3D) and the `FollowPuckViewportState` consume these interpolated locations and update their positions accordingly without additional animation. This design guarantees that the pucks and `FollowPuckViewportState` will always remain in sync.

Additionally, this PR fixes an issue in `Puck2D` that prevented unused properties of the `LocationIndicatorLayer` from getting reset if the location accuracy authorization changed. The implementation assumed that the properties dictionary was flat, but in fact the values were nested under top-level `paint` and `layout` keys. The fix was to build a flattened dictionary manually.

Lastly, some performance optimizations were performed to keep this new design from being too costly. First, `Puck2D` now does a lighter computation when the location changes. Second, `MapboxScaleBarOrnamentView` was modified slightly to significantly reduce the time required for it to update when the camera changes.

This PR does not change any public APIs.

## New Componets

- `CancelableContainer`: this is a small class that enables linking a `Cancelable`'s lifetime to the lifetime of the object that owns the container. It helps to simplify the implementations of `Puck2D`, `Puck3D`, and `FollowPuckViewportDataSource`
- `Interpolator` and `WrappingInterpolator` provide basic implementations of interpolating between two `Double`s given a percent. Right now, they both use linear interpolation, but they could be extended in the future to be configured with other curves. `WrappingInterpolator` is used for interpolating latitudes and directions which both wrap. It picks the shortest path between the from and to values.
- `InterpolatedLocation` is a simple struct that holds location data in a format that is more convenient for interpolation than working with `Location`.
- `InterpolatedLocationProducer` observes locations from the `LocationProducer`, drives the interpolation process by hooking into the display link, and provides API to allow the interpolated values to be observed.
- `LocationInterpolator` provides a basic implementation of interpolating between two `InterpolatedLocation` values given a percent`.
- `ObservableInterpolatedLocation` is a thin wrapper around `ObservableValue` that avoids propagating generics into classes that depend on it.

## New Extensions

- `Dictionary` has been extended with a `mapKeys` method, which is helpful when building the update dictionaries in `Puck2D`.

## Pull request checklist:


Sources/MapboxMaps/Foundation/ObservableValue.swift
```diff
@@ -0,0 +1,43 @@
+internal final class ObservableValue<Value> where Value: Equatable {
```
macdrevx, line 1: This was mostly just a refactor, but the amount of changes were significant enough for it to show as a new file.

Sources/MapboxMaps/Location/Puck/Puck2D.swift
```diff
@@ -184,77 +245,3 @@ private extension Puck2DConfiguration {
...
-    static func makeDefault(showBearing: Bool = false) -> Puck2DConfiguration {
-        let shadowImage = UIImage(named: "location-dot-outer", in: .mapboxMaps, compatibleWith: nil)!
-        return Puck2DConfiguration(
-            topImage: UIImage(named: "location-dot-inner", in: .mapboxMaps, compatibleWith: nil)!,
-            bearingImage: showBearing ? makeBearingImage(size: shadowImage.size) : nil,
-            shadowImage: shadowImage)
-    }
-}
-
-private func makeBearingImage(size: CGSize) -> UIImage {
```
macdrevx, line 199: These were moved to PuckType.swift

Sources/MapboxMaps/Location/Puck/Puck2D.swift
```diff
@@ -78,54 +100,60 @@ internal final class Puck2D: NSObject, Puck {
...
             }
-            layer.shadowImage = .constant(.name(Self.shadowImageId))
-            layer.locationTransition = StyleTransition(duration: 0.5, delay: 0)
-            layer.topImageSize = configuration.resolvedScale
-            layer.bearingImageSize = configuration.resolvedScale
-            layer.shadowImageSize = configuration.resolvedScale
-            layer.emphasisCircleRadiusTransition = StyleTransition(duration: 0, delay: 0)
-            layer.bearingTransition = StyleTransition(duration: 0, delay: 0)
+            newLayerLayoutProperties[.shadowImage] = Self.shadowImageId
+            newLayerPaintProperties[.locationTransition] = immediateTransition
```
pengdev, line 125: I think gl-native supports flattened layer properties, so no need to separate the layout and paint layer properties separately.

macdrevx: It's separated here to make the implementation cleaner. The key types are paint and layout specific: https://github.com/mapbox/mapbox-maps-ios/pull/1090/files#diff-49c72aa33067d00c9bc906b7e1e0d4e6d56bd77db95450efcf80115bc157307fR107-R108

Later, the keys are mapped into strings. We could use strings at this step, but it would either mean making the code much more verbose or using raw strings in the implementation.

Sources/MapboxMaps/Location/InterpolatedLocationProducer.swift
```diff
@@ -0,0 +1,89 @@
...
+    }
+}
+
+extension InterpolatedLocationProducer: LocationConsumer {
+    internal func locationUpdate(newLocation: Location) {
+        let currentDate = dateProvider.now
+
+        // as a first iteration, assume a 1s location update interval and use a
+        // slightly longer interpolation duration to avoid pauses between updates
+        let duration: TimeInterval = 1.1
```
pengdev, line 64: LGTM for the first iteration,
would it make sense to make the duration configurable according to the LocationProvider's update interval in the future?

Sources/MapboxMaps/Location/LocationInterpolator.swift
```diff
@@ -0,0 +1,72 @@
...
+           let endHeading = toLocation.heading {
+            heading = directionInterpolator.interpolate(
+                from: startHeading,
+                to: endHeading,
+                percent: percent)
+        } else if let endHeading = toLocation.heading {
+            heading = endHeading
+        } else {
+            heading = nil
+        }
```
pengdev, line 48: do you interpolate both the heading and course at the same time, even if one is not used?
And I assume this LocationInterpolator is shared between viewport and location manager, and by interpolating both, you can achieve setting the puck bearing source separately on viewport and location..

pengdev (approved):
LGTM % few questions
The concept is similar to what we have on Android, would be good to expose some config settings to the interpolator as follow up.

Sources/MapboxMaps/Location/InterpolatedLocationProducer.swift
macdrevx: Yes, something along those lines. I think we might also be able to explore automatically detecting the update rate.

Sources/MapboxMaps/Location/LocationInterpolator.swift
macdrevx: > do you interpolate both the heading and course at the same time, even if one is not used?
Yes
> And I assume this LocationInterpolator is shared between viewport and location manager, and by interpolating both, you can achieve setting the puck bearing source separately on viewport and location..
That's correct.

Sources/MapboxMaps/Location/InterpolatedLocationProducer.swift
```diff
@@ -0,0 +1,85 @@
+import CoreLocation
+
+internal protocol InterpolatedLocationProducerProtocol: AnyObject {
+    var location: InterpolatedLocation? { get }
+    func observe(with handler: @escaping (InterpolatedLocation) -> Bool) -> Cancelable
+}
+
+internal final class InterpolatedLocationProducer: NSObject, InterpolatedLocationProducerProtocol {
+    private var startDate: Date?
```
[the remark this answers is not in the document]
macdrevx: Yep, we have an internal ticket to migrate. We really want to be using the display link's `targetTimestamp` as described here: https://developer.apple.com/documentation/quartzcore/optimizing_promotion_refresh_rates_for_iphone_13_pro_and_ipad_pro

I'm having a conversation with @aleksigron later this week to discuss some topics related to this. I'll add you to the meeting.

Sources/MapboxMaps/Foundation/Interpolator.swift
```diff
@@ -0,0 +1,39 @@
+internal protocol InterpolatorProtocol {
+    func interpolate(from: Double, to: Double, percent: Double) -> Double
```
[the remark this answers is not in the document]
macdrevx: Re: using associated types: I did that initially, but it actually makes the protocol much less useful, IMO, because I can't inject an object of type InterpolatorProtocol anymore if I do that — the consuming class has to become generic.

Sources/MapboxMaps/Location/Puck/Puck2D.swift
```diff
@@ -45,10 +65,12 @@ internal final class Puck2D: NSObject, Puck {
...
     internal init(configuration: Puck2DConfiguration,
                   style: StyleProtocol,
-                  locationProducer: LocationProducerProtocol) {
+                  interpolatedLocationProducer: InterpolatedLocationProducerProtocol) {
         self.configuration = configuration
         self.style = style
-        self.locationProducer = locationProducer
+        self.interpolatedLocationProducer = interpolatedLocationProducer
+        // swiftlint:disable:next force_cast
+        self.encodedScale = (try! JSONSerialization.jsonObject(with: JSONEncoder().encode([configuration.resolvedScale])) as! [Any]).first!
```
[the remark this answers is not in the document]
macdrevx: scale is a Value which can be either a constant or an expression. This converts the value to a json object that we can pass in the properties dictionary. I'll try to make it a bit more safe…

Sources/MapboxMaps/Location/InterpolatedLocation.swift
```diff
@@ -0,0 +1,33 @@
+import CoreLocation
+
+internal struct InterpolatedLocation: Equatable {
```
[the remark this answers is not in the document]
macdrevx: You need a CLHeading to create a Location and CLHeading doesn't have any public constructors. Rather than trying to subclass it to add a constructor, it was simpler to create this alternative structure.

Sources/MapboxMaps/Camera/CameraAnimationsManager.swift
```diff
@@ -137,6 +137,17 @@ public class CameraAnimationsManager: CameraAnimationsManagerProtocol {
...
                      curve: UIView.AnimationCurve = .easeOut,
                      completion: AnimationCompletion? = nil) -> Cancelable? {
+        return internalEase(to: camera, duration: duration, curve: curve, completion: completion)
+    }
+
+    /// Ease to implementation that returns non-optional cancelable. The public API should have
+    /// been like this, but we have to wait until the next major version to change it. This internal API
+    /// allows us to avoid force-unwrapping for internal use.
+    @discardableResult
+    internal func internalEase(to camera: CameraOptions,
```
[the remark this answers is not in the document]
macdrevx: When we expose it publicly, we'll remove the internal prefix from the method name. I prefixed it to avoid ambiguity with the existing public method.

Sources/MapboxMaps/Foundation/Interpolator.swift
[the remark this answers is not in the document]
macdrevx: > Maybe progress, fraction/fractionComplete could be considered here, as percent can be interpreted as value per 100, i.e. 50, not 0.5.

I'm fine with fraction.

Sources/MapboxMaps/Camera/CameraAnimationsManager.swift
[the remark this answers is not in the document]
macdrevx: So, even if we expose it publicly now, the name would change in v11, so it doesn't really any work for folks using it.

Sources/MapboxMaps/Location/Puck/Puck2D.swift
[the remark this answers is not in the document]
macdrevx: Actually, this isn't any less safe than before. Previously, resolvedScale was set as the value for LocationIndicatorLayer's topImageSize, bearingImageSize, and shadowImageSize properties and then we were converting the LocationIndicatorLayer to a JSON object using `try! layer.jsonObject()`.

The only way this line would trap is if resolvedScale throws an error when it gets encoded. If the encoding succeeds, we can trust that decoding to a JSON object will also succeed unless JSONEncoder itself produces invalid JSON (we trust that this won't happen), and if decoding to a JSON object succeeds, then we know that the force cast to `[Any]` will succeed since we hard-code the input to be an array, and if the force cast succeeds, we know that the array will contain a single element (again since we hard-code that in the input), so the force-unwrap of `.first` will always succeed.

The reason that the input is wrapped in an array is that encoding and decoding JSON fragments aren't supported on some older versions of iOS that we still support.

That said, maybe I can wrap this up into a nicer abstraction :)

Sources/MapboxMaps/Location/Puck/Puck2D.swift
```diff
@@ -160,14 +194,41 @@ internal final class Puck2D: NSObject, Puck {
...
+            case .course:
+                layerProperties[LocationIndicatorLayer.PaintCodingKeys.bearing.rawValue] = location.course
+            }
+        case .reducedAccuracy:
+            fallthrough
+        @unknown default:
+            break
+        }
+
+        try! style.setLayerProperties(for: Self.layerID, properties: layerProperties)
```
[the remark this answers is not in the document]
macdrevx: On second thought, if this crashes, it Swift already logs the error that was thrown.

macdrevx:
@evil159 feedback addressed. Please take another look.

Sources/MapboxMaps/Location/Puck/Puck2D.swift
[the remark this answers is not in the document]
evil159: True that.

Sources/MapboxMaps/Location/Puck/Puck2D.swift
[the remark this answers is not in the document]
evil159: Thank you for the detailed explanation! Initially, I failed to notice that `resolvedScale` is a `Value<Double>`, not just a `Double` type. Very good reasoning about JSON encoding and decoding as well. I love how it's nicely tucked away into an extension now 😍

Sources/MapboxMaps/Camera/CameraAnimationsManager.swift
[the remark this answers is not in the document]
evil159: Sorry, I meant something like this - to expose it with the final name and a signature. And deprecate `ease(to...` with optional return value in v11.
```swift
    @discardableResult
    public func ease(to camera: CameraOptions,
                     duration: TimeInterval,
                     curve: UIView.AnimationCurve = .easeOut,
                     completion: AnimationCompletion? = nil) -> Cancelable? {

    @discardableResult
    public func ease(to camera: CameraOptions,
                               duration: TimeInterval,
                               curve: UIView.AnimationCurve,
                               completion: AnimationCompletion?) -> Cancelable {
```

Sources/MapboxMaps/Foundation/Interpolator.swift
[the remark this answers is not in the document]
evil159: Re: using associated types: Good point, in swift protocols with associated types can, indeed, be quite notorious to work with. If not taken under control generics could spread like a plague.

Sources/MapboxMaps/Camera/CameraAnimationsManager.swift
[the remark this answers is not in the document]
evil159: This would be introducing some issues with swift copier, not going forward with it.

evil159 (approved):

Sources/MapboxMaps/Location/InterpolatedLocationProducer.swift
```diff
@@ -0,0 +1,89 @@
...
+    private var startLocation: InterpolatedLocation?
+    private var endLocation: InterpolatedLocation?
+
+    private let observableInterpolatedLocation: ObservableInterpolatedLocationProtocol
+    private let locationInterpolator: LocationInterpolatorProtocol
+    private let dateProvider: DateProvider
+
+    internal var location: InterpolatedLocation? {
+        observableInterpolatedLocation.value
+    }
```
bamx23, line 20: @mapbox/maps-ios Is it okay, that this `location` is `nil` until the first call of `participate()` method even if `startLocation` and `endLocation` is set? It prevents configuration of Puck3D here: https://github.com/mapbox/mapbox-maps-ios/blob/8cf3f937cfd0df60fe741ce83a089296c85b2546/Sources/MapboxMaps/Location/Puck/Puck3D.swift#L56

macdrevx: Puck3D will be configured as soon as participate is called (on the next display link) because it observes the interpolated location producer: https://github.com/mapbox/mapbox-maps-ios/blob/8cf3f937cfd0df60fe741ce83a089296c85b2546/Sources/MapboxMaps/Location/Puck/Puck3D.swift#L11-L16

Are you running into an issue with this behavior?

bamx23: Yes, we have a test in Nav SDK that verifies [immediate creation of the layer for a 3D puck](https://github.com/mapbox/mapbox-navigation-ios/blob/9e42a853023d9ff1c0ed5535fd693a1b8ceb8411/Tests/MapboxNavigationTests/NavigationViewControllerTests.swift#L359-L372). Do you say this new behavior is intended? In this case we should stop testing this or update the test to work in async way.

macdrevx: The new behavior is intended.

Sources/MapboxMaps/Location/InterpolatedLocationProducer.swift
```diff
@@ -0,0 +1,89 @@
...
+    internal init(observableInterpolatedLocation: ObservableInterpolatedLocationProtocol,
+                  locationInterpolator: LocationInterpolatorProtocol,
+                  dateProvider: DateProvider,
+                  locationProducer: LocationProducerProtocol,
+                  displayLinkCoordinator: DisplayLinkCoordinator) {
+        self.observableInterpolatedLocation = observableInterpolatedLocation
+        self.locationInterpolator = locationInterpolator
+        self.dateProvider = dateProvider
+        super.init()
+        locationProducer.add(self)
```
rkapshuk-ring, line 31: Adding self as consumer for location producer instantly starts location updates.
That's said MapView will always start location updates using CLLocationManager event if user's location is not shown on map

macdrevx: https://github.com/mapbox/mapbox-maps-ios/issues/1251

macdrevx: @rkapshuk-ring the fix for this issue was released today in v10.4.3.

PR #1094, opened 2022-02-07
Animate once when FollowPuckViewportState starts

macdrevx:
* Changes the default animation duration to 0.3 to minimize visibility of remaining jump.
* Mimics Android implementation
* Use constant for interpolation duration in InterpolatedLocationProducer

## Pull request checklist:

### DIFF
--- a/Sources/MapboxMaps/Viewport/States/FollowPuck/FollowPuckViewportState.swift
+++ b/Sources/MapboxMaps/Viewport/States/FollowPuck/FollowPuckViewportState.swift
@@ -44,10 +44,27 @@ extension FollowPuckViewportState: ViewportState {
         guard updatingCameraCancelable == nil else {
             return
         }
-        updatingCameraCancelable = dataSource.observe { [mapboxMap] cameraOptions in
-            mapboxMap.setCamera(to: cameraOptions)
+        var animationStarted = false
+        var animationComplete = false
+
+        let compositeCancelable = CompositeCancelable()
+
+        updatingCameraCancelable = compositeCancelable
+
+        compositeCancelable.add(dataSource.observe { [mapboxMap, cameraAnimationsManager, options] cameraOptions in
+            if animationComplete {
+                mapboxMap.setCamera(to: cameraOptions)
+            } else if !animationStarted {
+                animationStarted = true
+                compositeCancelable.add(cameraAnimationsManager.ease(
+                    to: cameraOptions,
+                    duration: options.animationDuration,
+                    curve: .linear) { _ in
+                        animationComplete = true
+                    }!)
+            }
             return true
-        }
+        })
     }
 
     public func stopUpdatingCamera() {
--- a/Sources/MapboxMaps/Viewport/States/FollowPuck/FollowPuckViewportState.swift
+++ b/Sources/MapboxMaps/Viewport/States/FollowPuck/FollowPuckViewportState.swift
@@ -56,12 +56,12 @@ extension FollowPuckViewportState: ViewportState {
                 mapboxMap.setCamera(to: cameraOptions)
             } else if !animationStarted {
                 animationStarted = true
-                compositeCancelable.add(cameraAnimationsManager.ease(
+                compositeCancelable.add(cameraAnimationsManager.internalEase(
                     to: cameraOptions,
                     duration: options.animationDuration,
                     curve: .linear) { _ in
                         animationComplete = true
-                    }!)
+                    })
             }
             return true
         })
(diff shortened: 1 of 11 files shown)


Sources/MapboxMaps/Viewport/States/FollowPuck/FollowPuckViewportState.swift
```diff
@@ -44,10 +44,27 @@ extension FollowPuckViewportState: ViewportState {
...
+            if animationComplete {
+                mapboxMap.setCamera(to: cameraOptions)
+            } else if !animationStarted {
+                animationStarted = true
+                compositeCancelable.add(cameraAnimationsManager.ease(
+                    to: cameraOptions,
+                    duration: options.animationDuration,
+                    curve: .linear) { _ in
+                        animationComplete = true
+                    }!)
```
macdrevx, line 64: This force-unwrap is "safe" since `ease(to:…` never returns `nil`. We have a ticket to make the return type non-optional in v11.

Sources/MapboxMaps/Viewport/States/FollowPuck/FollowPuckViewportStateOptions.swift
```diff
@@ -9,7 +9,7 @@
                 zoom: CGFloat? = 16.35,
                 bearing: FollowPuckViewportStateBearing? = .heading,
                 pitch: CGFloat? = 45,
-                animationDuration: TimeInterval = 1) {
+                animationDuration: TimeInterval = 0.3) {
```
macdrevx, line 12: @pengdev okay with you if we change this default value to 0.3? When set to 1 second, I'm still seeing a pretty noticeable jump at the end of the DefaultViewportTransition.

macdrevx:
@pengdev I view this as more of a temporary workaround. It seems to me that the root cause is that DefaultViewportTransition doesn't yet handle the moving target problem. I will create a ticket for us to address that in a future version, at which point, I think we should be able to remove the initial animation from `FollowPuckViewportState`.

We'll probably either need to do that before removing the experimental designation from this API or in the next major version since it'd mean we can remove `animationDuration` from `FollowPuckViewportStateOptions`.

Sources/MapboxMaps/Viewport/States/FollowPuck/FollowPuckViewportStateOptions.swift
pengdev: I think that depends on the moving speed of the puck, right? if the puck is not moving fast, this might not be a concern? Anyway I think we have provided users the flexibility to adjust it based on their use case, wdyt?

pengdev:
> It seems to me that the root cause is that DefaultViewportTransition doesn't yet handle the moving target problem.

Yes, I tried to fix it from the `DefaultViewportTransition` by adjusting the end target on the fly and recalculating the interpolated center position. 

However, by design, in the DefaultViewportTransition, the center animation ends earlier than other animators, and in the end, there will always be some gap between  the target's location and the end camera of `DefaultViewportTransition`, and similar happens to bearing animator as well. If we want to keep the `DefaultViewportTransition`'s behaviour, I thought the current approach is the only way we can smooth it out.

Sources/MapboxMaps/Viewport/States/FollowPuck/FollowPuckViewportStateOptions.swift
macdrevx: That's true. It does depend on a lot of factors. I can revert it to 1.

Sources/MapboxMaps/Viewport/States/FollowPuck/FollowPuckViewportState.swift
evil159: Good that you highlighted this! The exclamation mark is quite hard to notice there :) It will take some time to get v11 out, maybe let's stick to the safer side here meanwhile and unwrap this. There will be a compiler error anyway when the return type is non-optional so it won't be forgotten.

macdrevx: How do you propose handling the case where it returns nil?

pengdev (approved):

Sources/MapboxMaps/Viewport/States/FollowPuck/FollowPuckViewportState.swift
macdrevx: I have an idea for how to clean this up. Will push shortly…

PR #1095, opened 2022-02-08
Refactor display link

evil159:
This PR consists of two states(split into two commits):
* move display link creation exclusively to `didMoveToWindow`(collab with @macdrevx)
* introduce `EmptyDisplayLink` to prevent having optional display link property

The latter is probably more like a suggestion/discussion point, but it should be good enough to be merged as well.

## Pull request checklist:


macdrevx:
Found this:

> Because you need to know which screen the window is on before creating the display link, you call this method when UIKit calls your view’s [didMoveToWindow()](https://developer.apple.com/documentation/uikit/uiview/1622527-didmovetowindow) method. UIKit calls this method the first time the view is added to a window and when the view is moved to another screen.

Source: https://developer.apple.com/documentation/metal/drawable_objects/creating_a_custom_metal_view

Sources/MapboxMaps/Foundation/MapView.swift
```diff
@@ -401,45 +358,30 @@ open class MapView: UIView {
...
+
+        // Checking Swift version as a proxy for iOS SDK version to enable
+        // building with iOS SDKs < 15
+#if swift(>=5.5)
+        if #available(iOS 15.0, *) {
+            displayLink.preferredFrameRateRange = oldDisplayLink.preferredFrameRateRange
         }
-    }
+#endif
+        displayLink.preferredFramesPerSecond = oldDisplayLink.preferredFramesPerSecond
```
macdrevx, line 382: What happens if these properties on MapView were never set? Will this overwrite the default values for the screen?

Tests/MapboxMapsTests/Foundation/MapViewTests.swift
```diff
@@ -102,27 +111,44 @@ final class MapViewTests: XCTestCase {
...
-    #endif
 
-    func testDisplayLinkTimestampIsNilWhenDisplayLinkIsNil() {
+    func testPreferredFrameRateRangeTransferredToNewDisplayLink() throws {
+        guard #available(iOS 15.0, *) else {
+            throw XCTSkip("Test requires iOS 15 or higher.")
+        }
+
+        let newWindow = UIWindow()
+        let frameRateRange = CAFrameRateRange(minimum: 0, maximum: 120, __preferred: 80)
```
macdrevx, line 121: I think there's another constructor without the underscores

Tests/MapboxMapsTests/Foundation/MapViewTests.swift
```diff
@@ -102,27 +111,44 @@ final class MapViewTests: XCTestCase {
...
+    func testDisplayLinkTimestampIsNilWhenDisplayLinkIsEmpty() {
+        dependencyProvider.makeDisplayLinkStub.defaultReturnValue = nil
         mapView.removeFromSuperview()
 
         XCTAssertNil(mapView.displayLinkTimestamp)
     }
 
-    func testDisplayLinkTimestampWhenDisplayLinkIsNonNil() {
+    func testDisplayLinkTimestampWhenDisplayLinkIsNotEmpty() {
+        dependencyProvider.makeDisplayLinkStub.defaultReturnValue = nil
```
macdrevx, line 138: This line seems unnecessary

Sources/MapboxMaps/Foundation/DisplayLinkProtocol.swift
```diff
@@ -14,4 +14,30 @@ internal protocol DisplayLinkProtocol: AnyObject {
...
+    // store the value as an `Any` in `_untypedPreferredFrameRateRange` below
+    // and make this a computed property.
+    @available(iOS 15.0, *)
+    var preferredFrameRateRange: CAFrameRateRange {
+        get { return (_untypedPreferredFrameRateRange as? CAFrameRateRange) ?? .default }
+        set { _untypedPreferredFrameRateRange = newValue }
+    }
+
+    private var _untypedPreferredFrameRateRange: Any?
+#endif
```
macdrevx, line 32: There's some non-trivial logic here. Please add unit tests to confirm default values and preferredFrameRate getter and setter.

Sources/MapboxMaps/Foundation/DisplayLinkProtocol.swift
```diff
@@ -14,4 +14,30 @@ internal protocol DisplayLinkProtocol: AnyObject {
     func invalidate()
 }
 
+internal final class EmptyDisplayLink: DisplayLinkProtocol {
+    var timestamp: CFTimeInterval { .nan }
+    var duration: CFTimeInterval { .nan }
```
macdrevx, line 19: These could also be implemented as `let` stored properties.

Sources/MapboxMaps/Foundation/MapView.swift
```diff
@@ -98,59 +98,29 @@ open class MapView: UIView {
...
     @_spi(Metrics) public var displayLinkTimestamp: CFTimeInterval? {
-        return displayLink?.timestamp
+        return displayLink.timestamp.isNaN ? nil : displayLink.timestamp
     }
 
     /// The `duration` from the underlying `CADisplayLink` if it exists, otherwise `nil`
     @_spi(Metrics) public var displayLinkDuration: CFTimeInterval? {
-        return displayLink?.duration
+        return displayLink.duration.isNaN ? nil : displayLink.duration
     }
```
macdrevx, line 124: We should add some simple unit tests for these.

macdrevx (commented):
Looking good. https://github.com/mapbox/mapbox-maps-ios/pull/1095#discussion_r801229209 is my biggest concern.

Sources/MapboxMaps/Foundation/MapView.swift
evil159: Yes indeed, when map view did move to window happens frame rate values for a new 'CADisplayLink' will be overridden with corresponding values from 'EmptyDisplayLink'.

evil159: The display link is created via the screen as that's the only way to have the link linked to the correct screen in multiscreen setup. The default values in `EmptyDisplayLink` are the same as ones from newly created `CADisplayLink`, so overriding should not affect anything.

Sources/MapboxMaps/Foundation/MapView.swift
evil159: I think `testDisplayLinkTimestampIsNilWhenDisplayLinkIsEmpty`, `testDisplayLinkTimestampWhenDisplayLinkIsNotEmpty`, `testDisplayLinkDurationIsNilWhenDisplayLinkIsEmpty` and `testDisplayLinkDurationWhenDisplayLinkIsNotEmpty` should cover these.

Tests/MapboxMapsTests/Foundation/MapViewTests.swift
evil159: Removed

Sources/MapboxMaps/Foundation/DisplayLinkProtocol.swift
evil159: Indeed, added `EmptyDisplayLinkTests` to test this.

Tests/MapboxMapsTests/Foundation/EmptyDisplayLinkTests.swift
```diff
@@ -0,0 +1,48 @@
...
+        try super.setUpWithError()
+
+        displayLink = EmptyDisplayLink()
+    }
+
+    override func tearDownWithError() throws {
+        try super.tearDownWithError()
+
+        displayLink = nil
+    }
```
macdrevx, line 17: Since there's nothing that can throw, let's just use setUp/tearDown

Sources/MapboxMaps/Foundation/DisplayLinkProtocol.swift
```diff
@@ -15,4 +15,31 @@ internal protocol DisplayLinkProtocol: AnyObject {
     func invalidate()
 }
 
+internal final class EmptyDisplayLink: DisplayLinkProtocol {
+    let timestamp: CFTimeInterval = .nan
+    let duration: CFTimeInterval = .nan
+    var isPaused: Bool = false
+    var preferredFramesPerSecond: Int = 0
+#if swift(>=5.5)
```
macdrevx, line 23: It occurs to me that this is somewhat problematic because it means these APIs won't be available to anyone using the direct download installation method. We have to build those XCFrameworks with our min supported xcode to ensure bitcode compatibility. I don't think we should change anything about this, but it's something to be aware of in case anyone runs into trouble.

cc: @ZiZasaurus

Tests/MapboxMapsTests/Foundation/MapViewTests.swift
```diff
@@ -105,32 +114,48 @@ final class MapViewTests: XCTestCase {
...
+
+        let newWindow = UIWindow()
+        let frameRateRange = CAFrameRateRange(minimum: 0, maximum: 120, preferred: 80)
+        mapView.preferredFrameRateRange = frameRateRange
+
+        newWindow.addSubview(mapView)
+
+        XCTAssertEqual(mapView.preferredFrameRateRange, frameRateRange)
+    }
+#endif
```
macdrevx, line 136: indentation mismatch?

Sources/MapboxMaps/Foundation/MapView.swift
```diff
@@ -464,47 +421,33 @@ open class MapView: UIView {
...
+
+        // Checking Swift version as a proxy for iOS SDK version to enable
+        // building with iOS SDKs < 15
+#if swift(>=5.5)
+        if #available(iOS 15.0, *) {
+            displayLink.preferredFrameRateRange = oldDisplayLink.preferredFrameRateRange
         }
-    }
+#endif
+        displayLink.preferredFramesPerSecond = oldDisplayLink.preferredFramesPerSecond
```
macdrevx, line 446: On line 443, we set `preferredFrameRateRange`, and here, we set `preferredFramesPerSecond`. Do they interact in any way? What happens if, on a real CADisplayLink, you set conflicting values? Does the order in which you set the conflicting values matter?

macdrevx (commented):
Left a few more thoughts, but https://github.com/mapbox/mapbox-maps-ios/pull/1095#discussion_r806404155 is the only one that would block my ✅

Sources/MapboxMaps/Foundation/MapView.swift
evil159: It seems that we can't avoid checking if the preferred frame rate is overridden. The `EmptyDisplayLink` concept doesn't seem to work in this setup after all. Let's scrap it and move on with the original refactoring then.

macdrevx (approved):

scripts/api-compatibility-check/.baseapi
```diff
@@ -1 +1 @@
-v10.1.0
\ No newline at end of file
+81618929724d94fcc64d4be24caed15357e98446
```
evil159, line 1: @macdrevx @OdNairy It seems that because of `MapView.didMoveToSuperview` and `MapView.willMove(toWindow:)` overrides were removed from `MapView` in this PR - it no longer passes the API compatibility check.
Not sure what is the process in this case(it seems the baseline has never been updated since the introduction). I did update the baseline to be the top of this branch. I think one way to go would be to merge this and then update the baseline when next release(or maybe as soon as beta/rc) is out.

Another thing to consider would be to exclude public overrides from API compatibility check. Not sure if `swift-api-digester` supports that(seems not). @OdNairy do you have any insight on this?

macdrevx: Should it be considered a breaking change? Seems like not since method dispatch would happen via objc message passing in this case, right? If we wanted, we could leave the overrides in place and have them just call super.

evil159: This shouldn't be considered a breaking change. I guess, `swift-api-digester` reports false positives for overridden public methods. Let's update the baseline so there are no dangling overrides.

PR #1113, opened 2022-02-10
Add Viewport Examples

macdrevx:
Ported from corresponding examples on Android ([ViewportShowcaseActivity.kt](https://github.com/mapbox/mapbox-maps-android/blob/main/app/src/main/java/com/mapbox/maps/testapp/examples/viewport/ViewportShowcaseActivity.kt), [AdvancedViewportGesturesExample.kt](https://github.com/mapbox/mapbox-maps-android/blob/main/app/src/main/java/com/mapbox/maps/testapp/examples/viewport/AdvancedViewportGesturesExample.kt)) with minor adjustments.

Also fixes an issue where the single tap and double tap to zoom in gestures could recognize simultaneously.

### DIFF
--- a/Apps/Examples/Examples/All Examples/AdvancedViewportGesturesExample.swift
+++ b/Apps/Examples/Examples/All Examples/AdvancedViewportGesturesExample.swift
@@ -1,6 +1,12 @@
 import UIKit
 @_spi(Experimental) import MapboxMaps
 
+// This example configures the viewport so that when it's in the follow puck state,
+// zoom and pitch gestures work alongside updates coming from the state itself.
+// Single tap on the map to switch to the overview state.
+//
+// When trying this example in the simulator, choose Features > Location > Freeway Drive
+// to get a good sense of the resulting user experience.
 @objc(AdvancedViewportGesturesExample)
 final class AdvancedViewportGesturesExample: UIViewController, ExampleProtocol {
 
--- a/Apps/Examples/Examples/All Examples/AdvancedViewportGesturesExample.swift
+++ b/Apps/Examples/Examples/All Examples/AdvancedViewportGesturesExample.swift
@@ -1,0 +1,139 @@
+import UIKit
+@_spi(Experimental) import MapboxMaps
+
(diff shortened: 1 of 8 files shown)


evil159 (approved):
Looking good! Maybe there can be some instructions on how to make these examples come to life. Not sure where to put them though. This could be beneficial especially for developers new to maps. I think having more detailed instructions here is more important than in other examples where the outcome is immediately visible. 

Things that can be pointed out:
- How to launch drive/run simulation in iOS simulator
- How to switch between viewport states in the advanced example

macdrevx:
These are good suggestions. We do this sort of thing on docs.mapbox.com for certain examples.

For now, I'll add some comments.

Apps/Examples/Examples/All Examples/AdvancedViewportGesturesExample.swift
```diff
@@ -0,0 +1,139 @@
...
+        mapView.mapboxMap.setCamera(to: CameraOptions(center: cupertino, zoom: 14))
+
+        mapView.location.options.puckType = .puck2D(.makeDefault(showBearing: true))
+        mapView.location.options.puckBearingSource = .course
+
+        followPuckViewportState = mapView.viewport.makeFollowPuckViewportState(
+            options: FollowPuckViewportStateOptions(
+                bearing: .course))
+
+        overviewViewportState = mapView.viewport.makeOverviewViewportState(
```
pengdev, line 39: seems you are using the real device's location for this example, while on Android, we have a custom LocationProvider implementation to emits fake location updates on a route. 
I think it is ok to be different on iOS, as the main showcase is how to customise gestures to work with the viewport, but would be good to consider adding the route simulator later on.

pengdev (approved):
LGTM with a comment.

Apps/Examples/Examples/All Examples/AdvancedViewportGesturesExample.swift
macdrevx: Added some comments to these examples explaining how to simulate location in the iOS simulator

PR #1118, opened 2022-02-14
Fix camera change events being fired after map has stopped moving

evil159:
### The setup

Mapview continues moving on its own after flicking it with a quick pan gesture. The velocity with which the map moves is seeded with the end velocity of the pan gesture. The velocity decreases every millisecond by `GestureOptions.panDecelerationFactor`. 

### The issue
Towards the very end of deceleration, the map view moves so briefly that it appears almost static. While these slight movements still generate e.g. `.cameraDidChange` events. Which gives the impression that camera events keep coming in even after the map has stopped moving.

### The fix
~This PR introduces a faster deceleration rate if combined map movement by two axes is less than a threshold. It addresses the issue by dramatically decreasing the period in which the map doesn't have any perceivable motion.  Both the faster deceleration rate(`0.95`) and its threshold(`20 pt/s`) are hand-picked hardcoded values. The deceleration rate won't be adjusted in case a provided deceleration rate is quick enough.~
The deceleration cut-off threshold increased to `20 pt/s`.

## Pull request checklist:
 - [x] Add any new public symbols to the Jazzy config's `custom_categories` (scripts/doc-generation/.jazzy.yaml)


Sources/MapboxMaps/Camera/GestureDecelerationCameraAnimator.swift
```diff
@@ -66,6 +66,18 @@ internal final class GestureDecelerationCameraAnimator: NSObject, CameraAnimator
 
         locationChangeHandler(location, toLocation)
 
+        let decelerationFactor: CGFloat
+
+        // check whether the end of animation is approaching
+        // if so - apply steeper deceleration
+        if sqrt(pow(velocity.x, 2) + pow(velocity.y, 2)) >= 20 {
```
macdrevx, line 73: How's the performance here? Is there some cheaper proxy for this calculation?

macdrevx (commented):
How many frames do you get at the steeper deceleration?

Could you achieve a similar effect by just increasing the constants at the end of the method? (which should be converted to an `if`, by the way — not sure why I used a `guard` at the end of a method).

evil159:
@macdrevx A very good point about just increasing deceleration cut-off constants! It felt different when run separately, but now when I compared them side-by-side - there's almost no visible difference between increasing the cut-off threshold vs faster deceleration rate at the end. 

https://user-images.githubusercontent.com/1478430/154027387-b1ec5749-788a-4130-bcb6-b801386976c9.mp4


https://user-images.githubusercontent.com/1478430/154027400-2b7df0cc-1ffb-4ecd-8d1e-c8511e218f5a.mp4

OdNairy (approved):

PR #1092, opened 2022-02-07
Add GestureOptions.pinchPanEnabled and GestureOptions.pinchZoomEnabled

ZiZasaurus:
Allows developers to isolate pan and zoom with pinch gesture.

Includes significant refactoring to `PanGestureHandler`. Now each combination of `pinchPanEnabled`, `pinchRotateEnabled`, and `pinchZoomEnabled` is backed by a separate `PinchBehavior` implementation. While it does introduce a larger number of classes, each new class is simple and easy to test, and the main `PinchGestureHandler` implementation is greatly simplified as well.

## Pull request checklist:
 - [x] Write tests for all new functionality. If tests were not written, please explain why.
 - [x] Add documentation comments for any added or updated public APIs.
 - [x] Describe the changes in this PR, especially public API changes.
 - [x] Add a changelog entry to to bottom of the relevant section (typically the `## main` heading near the top).
 - [x] Review and agree to the Contributor License Agreement ([CLA](https://github.com/mapbox/mapbox-maps-ios/blob/main/CONTRIBUTING.md#contributor-license-agreement)).

PR #1122, opened 2022-02-15
Zoom focal point

### LINKED ISSUE #879: Ability to specify anchor for Zoom gestures
## New Feature

We need to be able to specify anchor, from where the map will Zoom In or Out when using gestures. 
E.g.  Specify map center as the Zoom Gestures anchor, to always keep the center on the same location.

## Why

We are developing a Ride Booking application (read Uber). So we need to keep the location the user specified in center when he/she zooms In or Out.

P.S. The same request applies to the Android SDK, should I open a feature request there as well?

Thanks. 🙏

evil159:
This PR introduces the ability to override the focal point for zoom gestures. If the focal point is set e.g. to the center of the map, then zoom will anchor on that point, disregarding the actual touch location of any zoom gesture.

List of gestures supporting focal point override:
- double-tap to zoom in
- double touch to zoom out
- quick zoom
- pinch

Fixes #879 

## Pull request checklist:


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ Mapbox welcomes participation and contributions from everyone.
 * Prevent rendering in background by pausing/resuming display link in response to application or scene lifecycle events ([#1086](https://github.com/mapbox/mapbox-maps-ios/pull/1086))
 * Sync viewport and puck animations. ([#1090](https://github.com/mapbox/mapbox-maps-ios/pull/1090))
 * Add puckBearingEnabled property for location ([#1107](https://github.com/mapbox/mapbox-maps-ios/pull/1107))
+* Zoom focal point ([#1122](https://github.com/mapbox/mapbox-maps-ios/pull/1122))
```
OdNairy, line 10: Can you provide a more readable changelog entry?

Sources/MapboxMaps/Gestures/GestureHandlers/DoubleTapToZoomInGestureHandler.swift
```diff
@@ -1,11 +1,15 @@
 import UIKit
 
+internal protocol DoubleTapToZoomInGestureHandlerProtocol: GestureHandler {
+    var focalPoint: CGPoint? { get set }
```
OdNairy, line 4: Wherefrom this term come? It sounds wrongly placed a bit for me. 
Does `anchorPoint` work?

Sources/MapboxMaps/Gestures/GestureHandlers/DoubleTapToZoomInGestureHandler.swift
```diff
@@ -1,11 +1,15 @@
 import UIKit
 
+internal protocol DoubleTapToZoomInGestureHandlerProtocol: GestureHandler {
```
OdNairy, line 3: I don't think that the protocol name should say `DoubleTap` as there is nothing double-tap-related in it. 
Should it be just `ZoomGestureHandler`? 🤔

Sources/MapboxMaps/Gestures/GestureManager.swift
```diff
@@ -41,6 +45,8 @@ public final class GestureManager: GestureHandlerDelegate {
             gestureOptions.quickZoomEnabled = quickZoomGestureRecognizer.isEnabled
             gestureOptions.panMode = panGestureHandler.panMode
             gestureOptions.panDecelerationFactor = panGestureHandler.decelerationFactor
+#warning("TODO: which of many handlers should be source of truth here")
+            gestureOptions.focalPoint = doubleTapToZoomInGestureHandler.focalPoint
```
evil159, line 49: This is something that I'm not sure about. Gesture options are dynamically derived based on settings in gesture handlers. And since the focal point option is shared, there are multiple handlers that can populate this option. 
On the other hand, since the focal point option is passed to gesture handlers only through this property, then it can be populated from any of them as all of them should have the same value.

CHANGELOG.md
evil159: It made sense in my head 😀 Should be better now!

Sources/MapboxMaps/Gestures/GestureHandlers/PinchGestureHandlerImpl1.swift
```diff
@@ -40,7 +41,7 @@ internal final class PinchGestureHandlerImpl1: PinchGestureHandlerImpl {
         guard let view = gestureRecognizer.view else {
             return
         }
-        let pinchMidpoint = gestureRecognizer.location(in: view)
+        let pinchMidpoint = focalPoint ?? gestureRecognizer.location(in: view)
```
macdrevx, line 44: `pinchMidpoint` is used for other purposes as well. This change should only apply when it's used for anchor.

Sources/MapboxMaps/Gestures/GestureManager.swift
macdrevx: Up until now, we'd intentionally avoided having options that applied to more than one gesture on iOS. It's helped to keep the implementation simple and has avoided this problem.

You're right that there's no way for their values to diverge in the current design, so it's probably fine. If we wanted to, we could split this into 4 options, one for each gesture that supports it. The naming would follow the existing pattern:

* <gestureName><Option>
* doubleTapToZoomInFocalPoint
* doubleTouchToZoomOutFocalPoint
* quickZoomFocalPoint
* pinchFocalPoint

This would give greater granularity, but it would diverge from Android.

It would also pave the way towards reorganizing gesture options in the future if we wanted to replace the somewhat monolithic `GestureManager` with a per-gesture object, each with its own `gestureRecgonizer` and `options`.

@pengdev I'd like your input on this since it was a topic we'd discussed last year prior to GA.

macdrevx:
I'm working on doing away with the two pinch gesture implementations. It probably won't land until later this week or early next week, but I wanted to give you a heads-up about it since pan impl 2 is going away.

macdrevx (commented):
Holding further feedback until we discuss these initial comments.

Sources/MapboxMaps/Gestures/GestureHandlers/PinchGestureHandlerImpl1.swift
evil159: True, `pinchMidpoint` is used for zoom/rotation and dragging the map here. I think it is expected that the ability to pan with pinch gesture would be disabled when specifying a focal point(so to say drag happens from/to the specified focal point).
If we proceed with applying the focal point option only to zoom in pinch gesture. Then, in order to meet the feature request, we would need to add a separate option to disable panning while pinching. 

@pengdev how focal point handled on Android in this regard?

Sources/MapboxMaps/Gestures/GestureManager.swift
pengdev: l would suggest to have an aligned approach here, and in v11, we could add granular control for each gesture.

So we could use the single focal point that control affect all the gestures. On Android, we default to null, which means the focal point will always be the user's touch point.

Sources/MapboxMaps/Gestures/GestureHandlers/PinchGestureHandlerImpl1.swift
pengdev: I think the focal point doesn't affect pan gesture. 
And since we use one focal point for all gestures, we could always use the `pinchMidPoint` as the focal point?

Sources/MapboxMaps/Gestures/GestureManager.swift
macdrevx: Sounds good, let's proceed with a single option and counting on the handlers staying in sync.

Sources/MapboxMaps/Gestures/GestureHandlers/PinchGestureHandlerImpl1.swift
macdrevx: @ZiZasaurus is already working on adding a flag to disable panning during the pinch gesture: https://github.com/mapbox/mapbox-maps-ios/pull/1092

I agree that it wouldn't make much sense to use a non-nil focalPoint when panning is enabled. Should we add some logging when a gesture begins if we detect that condition?

Sources/MapboxMaps/Gestures/GestureOptions.swift
```diff
@@ -73,5 +73,8 @@ public struct GestureOptions: Equatable {
     /// Defaults to `UIScrollView.DecelerationRate.normal.rawValue`
     public var panDecelerationFactor: CGFloat = UIScrollView.DecelerationRate.normal.rawValue
 
+    /// By default, gestures rotate and zoom around the center of the gesture. Set this property to rotate and zoom around a fixed point instead.
```
macdrevx, line 76: Let's assume that https://github.com/mapbox/mapbox-maps-ios/pull/1092 will land for the beta and update this comment to encourage setting pinchPanEnabled to false when using a non-nil focalPoint.

Sources/MapboxMaps/Gestures/GestureHandlers/QuickZoomGestureHandler.swift
```diff
@@ -33,7 +35,8 @@ internal final class QuickZoomGestureHandler: GestureHandler {
             let distance = location.y - initialLocation.y
             // change by 1 zoom level per 75 points of translation
             let newZoom = initialZoom + distance / 75
-            mapboxMap.setCamera(to: CameraOptions(anchor: initialLocation, zoom: newZoom))
+            let anchor = focalPoint ?? initialLocation
+            mapboxMap.setCamera(to: CameraOptions(anchor: anchor, zoom: newZoom))
```
macdrevx, line 39: There's a pattern in these gesture handlers where we capture configuration options when the gesture starts so that if the options change during the gesture, the changes won't apply until the next gesture. Let's follow that pattern for focalPoint as well. See `initialRotateEnabled` in the pinch gesture handler for an example.

Sources/MapboxMaps/Gestures/GestureHandlers/PinchGestureHandlerImpl1.swift
macdrevx: and focalPoint should only apply to the final of the 3 setCameras that happen in the `.changed` case of this handler.

Sources/MapboxMaps/Gestures/GestureHandlers/ZoomGestureHandlerProtocol.swift
```diff
@@ -0,0 +1,5 @@
+import Foundation
+
+internal protocol ZoomGestureHandlerProtocol: GestureHandler {
```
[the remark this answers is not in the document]
macdrevx: Can we choose a name that reflects that? Maybe something like `FocusableGestureHandlerProtocol` 😬

Sources/MapboxMaps/Gestures/GestureHandlers/PinchGestureHandlerImpl1.swift
evil159: Made it so the focal point option affects only zoom and rotation. Added logging as well.

Sources/MapboxMaps/Gestures/GestureOptions.swift
evil159: Good point, added a sentence about it.

Sources/MapboxMaps/Gestures/GestureHandlers/PinchBehaviors/PinchBehaviorProvider.swift
```diff
@@ -24,18 +26,21 @@ internal final class PinchBehaviorProvider: PinchBehaviorProviderProtocol {
...
+                focalPoint: focalPoint,
                 mapboxMap: mapboxMap)
         case (true, true, false):
             return PanZoomPinchBehavior(
                 initialCameraState: initialCameraState,
                 initialPinchMidpoint: initialPinchMidpoint,
+                focalPoint: focalPoint,
                 mapboxMap: mapboxMap)
         case (true, false, true):
             return PanRotatePinchBehavior(
```
macdrevx, line 46: I think PanRotatePinchBehavior needs the focalPoint too

macdrevx: Or maybe PanZoomRotatePinchBehavior and PanZoomPinchBehavior should *not* have focalPoint

Sources/MapboxMaps/Gestures/GestureOptions.swift
```diff
@@ -60,5 +60,11 @@ public struct GestureOptions: Equatable {
...
+    ///
+    /// In order to force the map to only revolve around the specified focal point, disable
+    /// ``GestureOptions/pinchPanEnabled`` and ``GestureOptions/panEnabled``.
```
macdrevx, line 66: I don't think it's necessary to disable `GestureOptions.panEnabled`

Sources/MapboxMaps/Gestures/GestureHandlers/PinchGestureHandler.swift
```diff
@@ -86,13 +91,21 @@ internal final class PinchGestureHandler: GestureHandler, PinchGestureHandlerPro
         guard let view = gestureRecognizer.view else {
             return
         }
+
+        if panEnabled, focalPoint != nil {
+            Log.warning(
+                forMessage: "Possible pinch gesture recognizer misconfiguration: the specified focal point may be ignored when pinching. In order for the focal point to work, pinch pan has to be disabled.",
+                category: "Gestures")
+        }
+
```
macdrevx, line 100: Could we move this warning to the GestureManager.options setter?

Tests/MapboxMapsTests/Gestures/GestureHandlers/Mocks/MockDoubleTapToZoomInGestureHandler.swift
```diff
@@ -0,0 +1,5 @@
+@testable import MapboxMaps
+
+final class MockDoubleTapToZoomInGestureHandler: GestureHandler, FocusableGestureHandlerProtocol {
+    var focalPoint: CGPoint?
+}
```
macdrevx, line 5: Could we just have a single `MockFocusableGestureHandler`

Sources/MapboxMaps/Gestures/GestureHandlers/PinchBehaviors/PinchBehaviorProvider.swift
evil159: It sounds to me that it probably is a question of whether the focal point affects rotation or not. This PR assumes rotation is affected, from this perspective `PanRotatePinchBehavior` missing the `focalPoint` option is an oversight.

Sources/MapboxMaps/Gestures/GestureOptions.swift
evil159: Leaving regular pan enabled will enable the visible portion of the map to be moved relative to the specified focal point. So a completely different location might end up under the focal point instead of the location shown initially.

Sources/MapboxMaps/Gestures/GestureHandlers/PinchGestureHandler.swift
evil159: That was my first place to put it :) I put it here to avoid having false positives in case user configures options one by one: 
```swift
mapView.gestures.options.focalPoint = .zero
mapView.gestures.options.pinchPanEnabled = false
```

Sources/MapboxMaps/Gestures/GestureHandlers/PinchBehaviors/PinchBehaviorProvider.swift
macdrevx: What's the behavior if you're panning and rotating simultaneously with a fixed focal point?

Sources/MapboxMaps/Gestures/GestureOptions.swift
macdrevx: @pengdev any thoughts on this?

Tests/MapboxMapsTests/Gestures/GestureHandlers/DoubleTapToZoomInGestureHandlerTests.swift
```diff
@@ -61,4 +61,16 @@ final class DoubleTapToZoomInGestureHandlerTests: XCTestCase {
         easeToCompletion(.end)
         XCTAssertEqual(delegate.animationEndedStub.parameters, [.doubleTapToZoomIn])
     }
+
+    func testFocalPoint() {
+        let focalPoint = CGPoint(x: 1000, y: 1000)
+        gestureHandler.focalPoint = focalPoint
+        gestureRecognizer.getStateStub.defaultReturnValue = .recognized
+        mapboxMap.cameraState = .random()
```
macdrevx, line 69: Is setting the cameraState to random relevant to this test?

Tests/MapboxMapsTests/Gestures/GestureHandlers/DoubleTapToZoomInGestureHandlerTests.swift
```diff
@@ -61,4 +61,16 @@ final class DoubleTapToZoomInGestureHandlerTests: XCTestCase {
...
+
+        gestureRecognizer.sendActions()
+
+        XCTAssertEqual(cameraAnimationsManager.easeToStub.invocations.count, 1)
+        XCTAssertEqual(cameraAnimationsManager.easeToStub.parameters[0].camera.anchor, focalPoint)
```
macdrevx, line 74:
```suggestion
        XCTAssertEqual(cameraAnimationsManager.easeToStub.parameters.first?.camera.anchor, focalPoint)
```
Avoid crashing the tests if easeTo isn't invoked.

macdrevx: There are several other places in the PR that access array elements without checking the array length. I'm only going to comment on this one, but let's address all of them.

Sources/MapboxMaps/Gestures/GestureHandlers/PinchBehaviors/PinchBehaviorProvider.swift
evil159: Well... It behaves as if the focal point is not specified.

Tests/MapboxMapsTests/Gestures/GestureHandlers/PinchBehaviors/PinchBehaviorProviderTests.swift
```diff
@@ -12,7 +12,8 @@ final class PinchBehaviorProviderTests: XCTestCase {
                 rotateEnabled: true,
                 initialCameraState: .random(),
                 initialPinchMidpoint: .random(),
-                initialPinchAngle: .random(in: 0..<2 * .pi))
+                initialPinchAngle: .random(in: 0..<2 * .pi),
+                focalPoint: .random())
```
macdrevx, line 16:
```suggestion
                focalPoint: .random(.random()))
```

This will give you nils sometimes too. Probably not super important for this test since we expect it to just be a pass-through, but I wanted to make sure you were aware of it.

Tests/MapboxMapsTests/Gestures/GestureManagerTests.swift
```diff
@@ -466,4 +468,38 @@ final class GestureManagerTests: XCTestCase {
...
+
+        let secondFocalPoint = CGPoint.random()
+        doubleTapToZoomInGestureHandler.focalPoint = secondFocalPoint
+
+        XCTAssertEqual(gestureManager.options.focalPoint, doubleTapToZoomInGestureHandler.focalPoint)
+
+        doubleTapToZoomInGestureHandler.focalPoint = nil
+
+        XCTAssertNil(gestureManager.options.focalPoint)
+
```
macdrevx, line 503: I would recommend omitting this aspect of the test. Unlike setting `isEnabled` on a publicly-exposed gesture recognizer, we don't expose a way for developers to modify focalPoint on doubleTapToZoomInGestureHandler — it's just an implementation detail.

The other part of the test is the important part.

Sources/MapboxMaps/Gestures/GestureHandlers/PinchBehaviors/PinchBehaviorProvider.swift
macdrevx: Notes from our call:

Add `didSet` to `GestureOptions.pinchPanEnabled` that sets `focalPoint` to `nil` and logs a warning if `pinchPanEnabled` is `true` and `focalPoint != nil`. Update the documentation to describe this relationship.

Do the same to `GestureOptions.focalPoint`.

Remove `focalPoint` from all `Pan*PinchBehavior` classes

Sources/MapboxMaps/Gestures/GestureOptions.swift
macdrevx: I think if we're going to mention `panEnabled` it should be in the context of keeping a map *coordinate* fixed at specific focal point. That's a slightly different use case than forcing rotation and zoom to happen at a specified point.

Tests/MapboxMapsTests/Gestures/GestureHandlers/PinchBehaviors/PinchBehaviorProviderTests.swift
evil159: Thanks, I wasn't aware of this indeed!

Sources/MapboxMaps/Gestures/GestureHandlers/PinchBehaviors/PinchBehaviorProvider.swift
evil159: Closing this as we decided to disallow having pinch pan enabled and a fixed focal point at the same time.

Tests/MapboxMapsTests/Gestures/GestureHandlers/DoubleTapToZoomInGestureHandlerTests.swift
evil159: Not really, a copy-paste thing most likely.

Tests/MapboxMapsTests/Gestures/GestureHandlers/DoubleTapToZoomInGestureHandlerTests.swift
evil159: Good point, fixed it.

Tests/MapboxMapsTests/Gestures/GestureManagerTests.swift
evil159: Fair point, removed.

CHANGELOG.md
```diff
@@ -13,6 +13,8 @@ Mapbox welcomes participation and contributions from everyone.
 * Update to MapboxCoreMaps 10.4.0-beta.1 and MapboxCommon 21.2.0-beta.1. ([#1126](https://github.com/mapbox/mapbox-maps-ios/pull/1126))
 * Exposed APIs to allow positioning of other views relative to the logoView, compassView, scaleBarView and attributionButton. ([#1130](https://github.com/mapbox/mapbox-maps-ios/pull/1130))
 * Add `GestureOptions.pinchPanEnabled` and `.pinchZoomEnabled`. ([#1092](https://github.com/mapbox/mapbox-maps-ios/pull/1092))
+* Zoom focal point ([#1122](https://github.com/mapbox/mapbox-maps-ios/pull/1122))
+* Add focalPoint property to zoom and rotate gestures ([#1122](https://github.com/mapbox/mapbox-maps-ios/pull/1122))
```
pengdev, line 17: nit: duplicated change log entries for the same PR, probably we want to remove the first one?

Sources/MapboxMaps/Gestures/GestureHandlers/PinchBehaviors/PanRotatePinchBehavior.swift
```diff
@@ -1,3 +1,5 @@
+import CoreGraphics
```
pengdev, line 1: nit: I'm not familiar with swift, but is it needed as there's no other change in this class besides this import.

Sources/MapboxMaps/Gestures/GestureOptions.swift
```diff
@@ -34,7 +35,18 @@ public struct GestureOptions: Equatable {
...
     /// Whether pan is enabled for the pinch gesture.
     /// Defaults to `true`.
-    public var pinchPanEnabled: Bool = true
+    ///
+    /// Setting this property to `true` sets ``GestureOptions/focalPoint`` to `nil`.
+    public var pinchPanEnabled: Bool = true {
+        didSet {
+            if focalPoint != nil, pinchPanEnabled {
+                Log.warning(
+                    forMessage: "Gesture options misconfiguration detected: pinch pan cannot be enabled when having a fixed focal point. Focal point option will be set to nil.",
```
pengdev, line 44: if the default value for the `pinchPanEnabled` is true, what if the user set the focal point later when the `pinchPanEnabled` is already enabled?

Is there anyway we can handle it internally without changing `focalPoint` to nil when enabling `pinchPanEnabled`?

Sources/MapboxMaps/Gestures/GestureOptions.swift
```diff
@@ -60,5 +72,20 @@ public struct GestureOptions: Equatable {
...
+    ///
+    /// Setting this property to any valid point sets ``GestureOptions/pinchPanEnabled`` to `false`.
+    public var focalPoint: CGPoint? {
+        didSet {
+            if pinchPanEnabled, focalPoint != nil {
+                Log.warning(
+                    forMessage: "Gesture options misconfiguration detected: a fixed focal point cannot be specified with pinch pan enabled. Pinch pan will be disabled.",
+                    category: "Gestures")
+
+                pinchPanEnabled = false
```
pengdev, line 85: This behaviour might be a bit confusing? Can we cache the focal point while the pinchPan is in progress and restore it when the pinchPan is finished?

Sources/MapboxMaps/Gestures/GestureHandlers/PinchBehaviors/PanRotatePinchBehavior.swift
macdrevx: Xcode has started auto-importing frameworks which is leading to a lot of things like this happening. Probably there were other changes to this file in a previous iteration of this PR which led to this import which was then not reverted.

Sources/MapboxMaps/Gestures/GestureHandlers/PinchGestureHandler.swift
```diff
@@ -1,9 +1,11 @@
 import UIKit
+@_implementationOnly import MapboxCommon_Private
```
macdrevx, line 2: This import can be removed now

Tests/MapboxMapsTests/Gestures/GestureManagerTests.swift
macdrevx: Seems it's not fully removed

Sources/MapboxMaps/Gestures/GestureOptions.swift
evil159: Made it so 👍

Tests/MapboxMapsTests/Gestures/GestureManagerTests.swift
evil159: Now it should be 🤞

CHANGELOG.md
evil159: Well spotted, rebase artefacts, removed it.

Sources/MapboxMaps/Gestures/GestureOptions.swift
evil159: Resolved according to https://github.com/mapbox/mapbox-maps-ios/pull/1122#discussion_r812119694

Sources/MapboxMaps/Gestures/GestureOptions.swift
evil159: Resolved according to https://github.com/mapbox/mapbox-maps-ios/pull/1122#discussion_r812119694

Sources/MapboxMaps/Gestures/GestureHandlers/PinchGestureHandler.swift
evil159: This one stays, but the one from `GestureOptions` goes :)

Sources/MapboxMaps/Gestures/GestureHandlers/PinchBehaviors/PanRotatePinchBehavior.swift
evil159: @macdrevx exactly! I'm willing to keep it, Xcode has a point - we do use `CoreGrapics` types here.

Sources/MapboxMaps/Gestures/GestureOptions.swift
```diff
@@ -60,5 +60,10 @@ public struct GestureOptions: Equatable {
     /// Defaults to `UIScrollView.DecelerationRate.normal.rawValue`
     public var panDecelerationFactor: CGFloat = UIScrollView.DecelerationRate.normal.rawValue
 
+    /// By default, gestures rotate and zoom around the center of the gesture. Set this property to rotate and zoom around a fixed point instead.
+    ///
+    /// This property will be ignored if ``GestureOptions/pinchPanEnabled`` is set to `true`.
```
macdrevx, line 65:
```suggestion
    /// This property will be ignored by the pinch gesture if ``GestureOptions/pinchPanEnabled`` is set to `true`.
```

macdrevx (approved):

PR #1139, opened 2022-02-21
API breakage allowlist

OdNairy:
PR https://github.com/mapbox/mapbox-maps-ios/pull/1095 introduced legit breaking changes that we would like to ignore. This pull request adds functionality to allow some breaks from the filelist. 

## Pull request checklist:

### Useful links:
- https://github.com/apple/swift/blob/main/test/api-digester/breakage-allowlist.swift


scripts/api-compatibility-check/.baseapi
```diff
@@ -1 +1 @@
-81618929724d94fcc64d4be24caed15357e98446
+v10.3.0
```
macdrevx, line 1: This was lower before, right? 10.1 or something like that?

macdrevx (commented):
Question about the baseline

scripts/api-compatibility-check/.baseapi
OdNairy: We have to use the latest public release every time. 
I would like to automate this part this week.

scripts/api-compatibility-check/breakage_allowlist.txt
```diff
@@ -0,0 +1,7 @@
+// https://github.com/mapbox/mapbox-maps-ios/pull/1095
+Func MapView.didMoveToSuperview() has been removed
+Func MapView.willMove(toWindow:) has been removed
+
+// _spi(Experimental)
+Enum PinchGestureBehavior has been removed
+Var GestureOptions.pinchBehavior has been removed
```
OdNairy, line 7: @macdrevx I'd like to name each breaking change in this file (and remove them just after the new public release). What is the best name for these functions, WDYT?

macdrevx (approved):

PR #1141, opened 2022-02-21
Disable check-api-compatibility for PRs

macdrevx:
* Disabling this job temporarily to unblock PRs since it is yielding
  false positives.


.circleci/config.yml
```diff
@@ -117,7 +117,7 @@ workflows:
           xcode: "12.5.0"
           configuration: "Release"
 
-      - check-api-compatibility
+      # - check-api-compatibility
```
OdNairy, line 120: Can you instead of workflow fix, change the actual `check-breaking-api` step? It would temporary disable checks for nightly build as well.

macdrevx: Updated

OdNairy (approved):

PR #1154, opened 2022-02-25
Add CameraOptionsInterpolator

macdrevx:
Adds a new interpolator for CameraOptions that will be used in a subsequent PR to create a new type of animator (for internal use) to help solve the moving target problem in `DefaultViewportTransition`.

## Pull request checklist:


Sources/MapboxMaps/Camera/CameraOptionsInterpolator.swift
```diff
@@ -0,0 +1,68 @@
...
+internal final class CameraOptionsInterpolator: CameraOptionsInterpolatorProtocol {
+    private let optionalInterpolator = OptionalInterpolator()
+    private let coordinateInterpolator: CoordinateInterpolatorProtocol
+    private let uiEdgeInsetsInterpolator: UIEdgeInsetsInterpolatorProtocol
+    private let interpolator: InterpolatorProtocol
+    private let bearingInterpolator: InterpolatorProtocol
+
+    internal init(coordinateInterpolator: CoordinateInterpolatorProtocol,
+                  uiEdgeInsetsInterpolator: UIEdgeInsetsInterpolatorProtocol,
+                  interpolator: InterpolatorProtocol,
```
evil159, line 19: Should this parameter name indicate what kind of interpolator is expected? It seems it is used to interpolate values on two different scales(zoom 0-22 and pitch 0-85).

Sources/MapboxMaps/Foundation/Interpolators/CoordinateInterpolator.swift
```diff
@@ -0,0 +1,34 @@
...
+    func interpolate(from: CLLocationCoordinate2D,
+                     to: CLLocationCoordinate2D,
+                     fraction: Double) -> CLLocationCoordinate2D
+}
+
+internal final class CoordinateInterpolator: CoordinateInterpolatorProtocol {
+    private let interpolator: InterpolatorProtocol
+    private let longitudeInterpolator: InterpolatorProtocol
+
+    internal init(interpolator: InterpolatorProtocol,
```
evil159, line 13: I suggest using `latitudeInterpolator` as a name here as well.

macdrevx:
@evil159 These points seem to be driving toward a question of how do we name injected dependencies? Is it based on…

* What they are, or
* How they're used?

Up to this point, I've been naming them based on what they are.

In this case `interpolator: InterpolatorProtocol` denotes an interpolator without any special behaviors that operates on `Double`s. `longitudeInterpolator: InterpolatorProtocol` or `bearingInterpolator: InterpolatorProtocol` denote interpolators that are specifically suited to do the wrapping required by those situations.

I think what makes it confusing is that all three parameters have the same type. I'm going to think about whether we could use the type system to make it more clear.

macdrevx:
@evil159 I pushed some changes that should make this clearer. They also confine the ranges for wrapping longitude and direction values to a single place. It'd been bothering me that I had to write those ranges any time I wanted to create a wrapping interpolator.

Thanks for the feedback, and please take another look!

Sources/MapboxMaps/Camera/CameraOptionsInterpolator.swift
macdrevx: Most of this comment is addressed in my replies and changes below, but one additional note here is that zoom and pitch can use the same interpolator since the interpolation works the same way for both (interpolation of double values without any special wrapping or other logic).

evil159 (approved):
Thanks @macdrevx, now it's definitely clearer! I like how direction and longitude interpolation ranges are defined once to be reused elsewhere. On the flip side interpolars became more complex, though I think clarity improvements overweights this increase. 

As for naming things: I pointed those names out because it's hard to tell what kind or interpolator supposed to passed to as a parameter when its name and type is quite generic, when looking just at the interface, not the implementation. 

Certainly, the presence of other interpolators is the key factor why it needed to be addressed. I think it would have been fine to use `interpolator` as a name if that was the only parameter of such type.

PR #1157, opened 2022-02-28
Metal view explicit drawing

evil159:
This PR enables explicit drawing behaviour for `MTKView`. When core asks to repaint the map, instead of calling `setNeedsDisplay()` on the next display link callback, we call directly ask the metal view to redraw itself(`draw()`).

### Before

Core would call schedule repaint on the MapView. Only after the next Vsync(on `CADisplayLink` callback) `MapView` would notify the metal view that its content needs to be redrawn. Then the `MTKView` view is actually redrawn at "the next drawing cycle"(probably at the next VSync) according to the docs.

With this setup we have at least one frame lag on updating the metal view from the moment core requested it.

```mermaid
sequenceDiagram
Core->>MapView: schedule repaint
participant VSync
MapView->>MTKView: set needs display
MTKView->>"the next drawing cycle": draw
```

### After

~Core calls schedule repaint on the `MapView`. `MapView` immediately asks `MTKView` to redraw its content. No lag.~

Core calls schedule repaint on the `MapView`. After the next VSync`MapView` asks `MTKView` to redraw its content.

```mermaid
sequenceDiagram
Core->>MapView: schedule repaint
participant VSync Again
MapView->>MTKView: draw
```
**Both diagrams above assume all invocations are performed instantly.*

## Pull request checklist:

### DIFF
--- a/Sources/MapboxMaps/Foundation/MapView.swift
+++ b/Sources/MapboxMaps/Foundation/MapView.swift
@@ -81,7 +81,6 @@ open class MapView: UIView {
     internal private(set) var resourceOptions: ResourceOptions!
 
     private var needsDisplayRefresh: Bool = false
-    private var displayCallback: (() -> Void)?
     private var displayLink: DisplayLinkProtocol?
 
     /// Holding onto this value that comes from `MapOptions` since there is a race condition between
@@ -467,7 +466,7 @@ open class MapView: UIView {
 
         if needsDisplayRefresh {
             needsDisplayRefresh = false
-            displayCallback?()
+            metalView?.draw()
         }
     }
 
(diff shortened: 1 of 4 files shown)


Sources/MapboxMaps/Foundation/MapView.swift
```diff
@@ -582,7 +578,7 @@ extension MapView: DelegatingMapClientDelegate {
         metalView.isOpaque = isOpaque
         metalView.layer.isOpaque = isOpaque
         metalView.isPaused = true
-        metalView.enableSetNeedsDisplay = true
+        metalView.enableSetNeedsDisplay = false
```
OdNairy, line 581: `enableSetNeedsDisplay` default value is false. I suppose we can just remove it.

OdNairy (approved):

astojilj:
Could we add a reference to an issue, or an explanation, why this feature is required? Thanks.

macdrevx (approved):

evil159:
@astojilj I updated the description of this PR with more info on before and after(on the second try). The main motivation for this change was that it seems redundant to use metal view's draw notifications behaviour(`setNeedsDisplay`) together with draw notifications from Core/GL Native(`scheduleRepaint`).

astojilj (approved):
> With this setup we have at least one frame lag on updating the metal view from the moment core requested it.

PR #1162, opened 2022-03-01
Clarify Style.transition usage for changing styles

macdrevx:
Relates to https://github.com/mapbox/mapbox-maps-android/pull/1174

### DIFF
--- a/Sources/MapboxMaps/Style/Style.swift
+++ b/Sources/MapboxMaps/Style/Style.swift
@@ -281,11 +281,14 @@ public final class Style: StyleProtocol {
     /// Get or set the map style's transition options.
     ///
     /// By default, the style parser will attempt to read the style default
-    /// transition, if any, falling back to an immediate transition otherwise.
+    /// transition, if any, falling back to a 0.3 s transition otherwise.
     ///
     /// Overridden transitions are reset once a new style has been loaded.
     /// To customize the transition used when switching styles, set this
-    /// property during `MapEvents.EventKind.styleDataLoaded`.
+    /// property after `MapEvents.EventKind.styleDataLoaded` where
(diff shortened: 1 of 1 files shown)


Sources/MapboxMaps/Style/Style.swift
```diff
@@ -281,13 +281,12 @@ public final class Style: StyleProtocol {
     /// Get or set the map style's transition options.
     ///
     /// By default, the style parser will attempt to read the style default
-    /// transition options, if any, falling back to an immediate transition
-    /// otherwise.
+    /// transition, if any, falling back to an immediate transition otherwise.
```
pengdev, line 284: I think the style's default transition is not immediate, but transition with 300ms duration https://docs.mapbox.com/mapbox-gl-js/style-spec/transition/#duration

Sources/MapboxMaps/Style/Style.swift
```diff
@@ -281,13 +281,12 @@ public final class Style: StyleProtocol {
...
     ///
-    /// The style transition is re-evaluated when a new style is loaded.
-    ///
-    /// - Attention:
-    ///     Overridden transition options are reset once a new style has been loaded.
+    /// Overridden transitions are reset once a new style has been loaded.
+    /// To customize the transition used when switching styles, set this
+    /// property during `MapEvents.EventKind.styleDataLoaded`.
```
pengdev, line 288:
```suggestion
    /// property after `MapEvents.EventKind.styleDataLoaded`.
```

pengdev: And should we also clarify to set it before style is fully loaded(sprite and sources are loaded)

pengdev (approved):

PR #1163, opened 2022-03-02
Add support for non-HTML attribution

### LINKED ISSUE #1164: Attribution parser doesn't show source attributions
## Environment

- Xcode version:
- iOS version:
- Devices affected: All
- Maps SDK Version: v10

When attribution is added using source (for eg. TileJson). It should show on the attribution modal. this is to track the issue raised here https://github.com/mapbox/mapbox-maps-internal/issues/1262 

## Expected behavior

Attribution related to source should show on attribution dialog.

evil159:
Currently there is no attribution in case the provided source attribution is in form of plain text(`{attribution: "OpenStreetMap contributors, CC-BY-SA"}`. This PR adds support for plain text attributions.

| Before | v1 | v2 |
| ---- | ---- | ---- |
|   |  |   |

Initially I went with the same approach as on Android https://github.com/mapbox/mapbox-maps-android/pull/1087 to disable alert's actions without URLs(see v1 image above), but it does not look good on iOS due to platform differences. Since there can only be a single plain text attribution I decided put it into alert's message instead(v2).

Fixes #1164

## Pull request checklist:


Sources/MapboxMaps/Style/AttributionDialogManager.swift
```diff
@@ -131,11 +132,18 @@ extension AttributionDialogManager: InfoButtonOrnamentDelegate {
         let bundle = Bundle.mapboxMaps
 
         if let attributions = dataSource?.attributions() {
-            for attribution in attributions {
-                let action = UIAlertAction(title: attribution.title, style: .default) { _ in
-                    self.delegate?.attributionDialogManager(self, didTriggerActionFor: attribution)
+
+            // Non actionable single item get's displayed as alert's title
```
macdrevx, line 136:
```suggestion
            // Non actionable single item gets displayed as alert's title
```

Sources/MapboxMaps/Style/AttributionDialogManager.swift
```diff
@@ -131,11 +132,18 @@ extension AttributionDialogManager: InfoButtonOrnamentDelegate {
...
+            if attributions.count == 1, let attribution = attributions.first, attribution.url == nil {
+                alert.message = attribution.title
+            } else {
+                for attribution in attributions {
+                    let action = UIAlertAction(title: attribution.title, style: .default) { _ in
+                        self.delegate?.attributionDialogManager(self, didTriggerActionFor: attribution)
+                    }
```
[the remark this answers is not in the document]
macdrevx: I actually wonder if we wouldn't be better off changing the signature of the delegate method to take a non-optional URL instead of an Attribution and moving the whole guard statement I linked into here.

macdrevx (commented):
I like the UI! Nice job.

What changes do we need to make to add really good tests for this part of the SDK? Let's incorporate refactoring for testability into every change we make unless there's a really good reason not to.

Sources/MapboxMaps/Foundation/MapView+Attribution.swift
```diff
@@ -7,7 +7,10 @@ extension MapView: AttributionDialogManagerDelegate {
     }
 
     func attributionDialogManager(_ attributionDialogManager: AttributionDialogManager, didTriggerActionFor attribution: Attribution) {
-        let url: URL = attribution.isFeedbackURL ? mapboxFeedbackURL() : attribution.url
+        guard let url: URL = attribution.isFeedbackURL ? mapboxFeedbackURL() : attribution.url else {
```
OdNairy, line 10: Do we need an explicit `URL` type in this line?

Sources/MapboxMaps/Style/AttributionDialogManager.swift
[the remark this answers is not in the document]
evil159: Yeah, with optional url it looks quite ugly, I introduced `Attribution.Kind` to help with this.

evil159:
> What changes do we need to make to add really good tests for this part of the SDK? Let's incorporate refactoring for testability into every change we make unless there's a really good reason not to.

Thanks for calling me out on this 😅 With the changes to `Attribution` I added tests covering almost all of it. 

A bit trickier with `AttributionDialogManager` as it has the logic to generate correct attribution dialog using `UIAlertController`. An option I considered was to mirror `UIAlertController`/`UIAlertAction` infrastructure with our own data classes/builders to make it easier to test the logic separately. But, it turned out that `UIAlertController` fares reasonably well in unit tests.

Sources/MapboxMaps/Foundation/MapView+Attribution.swift
evil159: No need, indeed. It went away after a refactoring related to https://github.com/mapbox/mapbox-maps-ios/pull/1163#discussion_r817311749

Tests/MapboxMapsTests/Style/AttributionDialogTests.swift
```diff
@@ -0,0 +1,217 @@
...
+        mockDataSource = nil
+        mockDelegate = nil
+    }
+
+    func testShowTelemetryDialogMetricsEnabled() throws {
+        let viewController = UIViewController()
+        let bundle = Bundle.mapboxMaps
+        let window = UIWindow()
+        window.makeKeyAndVisible()
+        window.rootViewController = viewController
```
macdrevx, line 32: If this ordering results in a warning logged to the console, you may want to try setting rootViewController before calling makeKeyAndVisible

Tests/MapboxMapsTests/Style/AttributionTests.swift
```diff
@@ -0,0 +1,161 @@
...
+
+    override func setUp() {
+        super.setUp()
+        Bundle.mapboxMapsMetadata.version = "AttributionTests"
+    }
+
+    override func tearDown() {
+        super.tearDown()
+        Bundle.mapboxMapsMetadata.version = sdkVersion
+    }
```
macdrevx, line 18: I was surprised that `mapboxMapsMetadata` was settable. That looks like a bug in my opinion. Could you avoid setting it here and use string interpolation to create the expected value below?

macdrevx (approved):
Just a couple additional suggestions. Approving!

PR #1183, opened 2022-03-09
App extensions support

evil159:
This PR makes possible to use `MapView` inside an application extension 🎉 

### Preconditions

The limitation of being able to only mark certain methods/entities being unavailable in application extensions with `@available(iOSApplicationExtension, unavailable)`(not vice versa). Plus the absence of run/compile -time app extension check (`if #available(iOSApplicationExtension)`). Has certainly made an impact on the design of the affected public API. 

### Changes

In order to be able to isolate `UIApplication`-specific behaviour, the map view is injected with `InterfaceOrientationProvider` and `AttributionURLOpener`. 

`InterfaceOrientationProvider` is a protocol the map view uses to obtain the current interface orientation. The interface orientation is used to ensure the map view gets correct heading data from its location producer.
`AttributionURLOpener` is a protocol the map view uses to open attribution-related web pages.

In order to preserve the existing map view's public initialiser, I created internal `DefaultAttributionURLOpener`, as well as `UIApplicationInterfaceOrientationProvider` and `DefaultInterfaceOrientationProvider`. 

`UIApplicationInterfaceOrientationProvider` and `DefaultAttributionURLOpener` are using `UIApplication.shared`, thus they are available only in app context. 

In case the map view is used in an app extension context - developers are required to provide their own `AttributionURLOpener`(utilising `NSExtensionContext.open(_:completionHandler:)`). And their own `InterfaceOrientationProvider` in case they are targeting iOS 11 or 12.

And finally this PR removes all `@available(iOSApplicationExtension, unavailable)` annotations from the top level. Instead it marks the existing initialiser with being unavailable in an app extension. And creates two alternative initialisers to be used in app extensions.

## Pull request checklist:

### DIFF
--- a/Sources/MapboxMaps/Foundation/InterfaceOrientationProvider.swift
+++ b/Sources/MapboxMaps/Foundation/InterfaceOrientationProvider.swift
@@ -2,8 +2,18 @@ import Foundation
 import UIKit
 import CoreLocation
 
+/// A protocol that supplies current interface orientation for the specified view.
+///
+/// Use this protocol when the map view is used in non-application target(e.g. application extension target).
 @available(iOS, deprecated: 13)
 public protocol InterfaceOrientationProvider {
+
+    /// Asks the provider for the interface orientation of the provided view.
+    ///
+    /// When a device is rotated map view passes current interface orientation to its location producer in order to ensure heading is displayed correctly.
+    /// - Parameters:
+    ///   - view: The view to get interface orientation from.
+    /// - Returns: The interface orientation for the provided view.
     func interfaceOrientation(for view: UIView) -> UIInterfaceOrientation?
 }
 
--- a/Sources/MapboxMaps/Foundation/InterfaceOrientationProvider.swift
+++ b/Sources/MapboxMaps/Foundation/InterfaceOrientationProvider.swift
@@ -33,8 +33,14 @@ extension InterfaceOrientationProvider {
 @available(iOS, deprecated: 13)
 @available(iOSApplicationExtension, unavailable)
 internal final class UIApplicationInterfaceOrientationProvider: InterfaceOrientationProvider {
+    private let application: UIApplicationProtocol
+
+    init(application: UIApplicationProtocol = UIApplication.shared) {
+        self.application = application
+    }
+
     func interfaceOrientation(for view: UIView) -> UIInterfaceOrientation? {
-        return UIApplication.shared.statusBarOrientation
+        return application.statusBarOrientation
     }
 }
 
--- a/Sources/MapboxMaps/Foundation/InterfaceOrientationProvider.swift
+++ b/Sources/MapboxMaps/Foundation/InterfaceOrientationProvider.swift
@@ -1,0 +1,57 @@
+import Foundation
+import UIKit
+import CoreLocation
+
+@available(iOS, deprecated: 13)
+public protocol InterfaceOrientationProvider {
+    func interfaceOrientation(for view: UIView) -> UIInterfaceOrientation?
+}
+
+extension InterfaceOrientationProvider {
+    internal func headingOrientation(for view: UIView) -> CLDeviceOrientation? {
+        // locationProvider.headingOrientation should be adjusted based on the
+        // current UIInterfaceOrientation of the containing window, not the
+        // device orientation
+        guard let interfaceOrientation = interfaceOrientation(for: view) else {
+            return nil
+        }
+
+        return CLDeviceOrientation(interfaceOrientation: interfaceOrientation)
+    }
+}
+
+@available(iOS, deprecated: 13)
+@available(iOSApplicationExtension, unavailable)
+internal final class UIApplicationInterfaceOrientationProvider: InterfaceOrientationProvider {
+    func interfaceOrientation(for view: UIView) -> UIInterfaceOrientation? {
+        return UIApplication.shared.statusBarOrientation
+    }
+}
+
+@available(iOS 13.0, *)
+internal final class DefaultInterfaceOrientationProvider: InterfaceOrientationProvider {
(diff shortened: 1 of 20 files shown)


Sources/MapboxMaps/Foundation/InterfaceOrientationProvider.swift
```diff
@@ -0,0 +1,67 @@
+import Foundation
+import UIKit
+import CoreLocation
+
+/// A protocol that supplies current interface orientation for the specified view.
+///
+/// Use this protocol when the map view is used in non-application target(e.g. application extension target).
```
macdrevx, line 7:
```suggestion
/// Use this protocol when the map view is used in non-application target (e.g. application extension target).
```

Sources/MapboxMaps/Foundation/InterfaceOrientationProvider.swift
```diff
@@ -0,0 +1,67 @@
...
+    internal func headingOrientation(for view: UIView) -> CLDeviceOrientation? {
+        // locationProvider.headingOrientation should be adjusted based on the
+        // current UIInterfaceOrientation of the containing window, not the
+        // device orientation
+        guard let interfaceOrientation = interfaceOrientation(for: view) else {
+            return nil
+        }
+
+        return CLDeviceOrientation(interfaceOrientation: interfaceOrientation)
+    }
```
macdrevx, line 30: Let's add tests for this

Sources/MapboxMaps/Foundation/MapView.swift
```diff
@@ -174,30 +173,90 @@ open class MapView: UIView {
...
+                            urlOpener: AttributionURLOpener) {
+        self.init(frame: frame,
+                  mapInitOptions: mapInitOptions,
+                  dependencyProvider: MapViewDependencyProvider(),
+                  orientationProvider: DefaultInterfaceOrientationProvider(),
+                  urlOpener: urlOpener)
+    }
+
+    @available(iOSApplicationExtension, unavailable)
     required public init?(coder: NSCoder) {
```
macdrevx, line 236: All the public init methods except for this one are now convenience initializers. Does that impact the ability to subclass MapView from another module?

macdrevx, line 236: Can you use xibs or storyboards together with App Extensions? If so, what would happen if you tried to do so with an instance of MapView?

macdrevx (commented):
Looks great! Just a few minor comments and questions.

Sources/MapboxMaps/Foundation/InterfaceOrientationProvider.swift
evil159: Added `testHeadingOrientationWrapperCallsInterfaceOrientation` for this

Sources/MapboxMaps/Foundation/InterfaceOrientationProvider.swift
```diff
@@ -0,0 +1,67 @@
...
+        case .landscapeLeft:
+            self = .landscapeRight
+        case .landscapeRight:
+            self = .landscapeLeft
+        case .portraitUpsideDown:
+            self = .portraitUpsideDown
+        default:
+            self = .portrait
+        }
+    }
```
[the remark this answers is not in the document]
evil159: Added `testInterfaceToDeviceOrientationConversion` for this.

Tests/MapboxMapsTests/Foundation/MapViewTests.swift
```diff
@@ -310,4 +328,38 @@ final class MapViewTests: XCTestCase {
...
+        XCTAssertEqual(attributionURLOpener.openAttributionURLStub.invocations.count, 1)
+        XCTAssertEqual(attributionURLOpener.openAttributionURLStub.invocations.first?.parameters, url)
+    }
+}
+
+extension MapViewTests: AttributionDataSource {
+    func attributions() -> [Attribution] {
+        return []
+    }
 }
```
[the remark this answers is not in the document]
evil159: I somehow overlooked that there already are mocks for this, thanks for pointing it out!

Sources/MapboxMaps/Foundation/MapView.swift
evil159: Hmmm... Good point, with this setup it would pretty much by impossible to subclass it.

evil159: I made two newly added public initialisers designated to address this.

Sources/MapboxMaps/Foundation/MapView.swift
```diff
@@ -174,30 +173,94 @@ open class MapView: UIView {
...
+
     /// Initialize a MapView
     /// - Parameters:
     ///   - frame: frame for the MapView.
     ///   - mapInitOptions: `MapInitOptions`; default uses
     ///    `ResourceOptionsManager.default` to retrieve a shared default resource option, including the access token.
-    public init(frame: CGRect, mapInitOptions: MapInitOptions = MapInitOptions()) {
-        dependencyProvider = MapViewDependencyProvider()
+    @available(iOSApplicationExtension, unavailable)
+    public convenience init(frame: CGRect, mapInitOptions: MapInitOptions = MapInitOptions()) {
```
macdrevx, line 185: We may need to make this one a designated initializer as well to avoid breaking existing subclasses that use this when calling super.init.

Come to think of it, we could add a MapView subclass for each designated initializer to the test target to avoid regressions in this area.

evil159: Brilliant idea! I think one `MapView` subclass overriding all designated public initialisers should do the job as well. Is there a reason to have a subclass per each designated initializer?

macdrevx: Nope, that works.

macdrevx (approved):

PR #1182, opened 2022-03-09
Simplify FlyToCameraAnimator init interface

macdrevx:
This refactoring is part two of a series that will unlock better unit testability for CameraAnimationsManager. (part 1: #1181)

* 3 of the parameters could be obtained via the injected MapboxMap

## Pull request checklist:
 - [x] Write tests for all new functionality. If tests were not written, please explain why.
 - [x] Describe the changes in this PR, especially public API changes.
 - [x] Review and agree to the Contributor License Agreement ([CLA](https://github.com/mapbox/mapbox-maps-ios/blob/main/CONTRIBUTING.md#contributor-license-agreement)).

PR #1189, opened 2022-03-10
Refactor basic camera animator

macdrevx:
This refactoring is part three of a series that will unlock better unit testability for CameraAnimationsManager. (part 2: https://github.com/mapbox/mapbox-maps-ios/pull/1182)

* Updates `BasicCameraAnimator` to use the internal impl pattern.
* The first commit is configured to make the following more clear for review purposes:
  * The former `BasicCameraAnimator` is most closely related to the new `BasicCameraAnimatorImpl`
  * The new `BasicCameraAnimator` is a thin wrapper around `BasicCameraAnimatorImpl`
  * The former `BasicCameraAnimatorTests` is most closely related to the new `BasicCameraAnimatorImplTests`
* The second commit updates the file names based on their final contents.

## Pull request checklist:


Tests/MapboxMapsTests/Camera/BasicCameraAnimatorTests2.swift
```diff
@@ -0,0 +1,166 @@
...
+
+    override func tearDown() {
+        delegate = nil
+        animator = nil
+        impl = nil
+        super.tearDown()
+    }
+
+    func testOwner() {
+        impl.owner = .random()
```
evil159: Should the opposite to be tested as well? 

The current test tests the setter of the property on impl side and the getter on animator side.

While the opposite would cover the setter of the property on the animator side and the getter on impl side.

```swift 
animator.owner = .random()
XCTAsseartEqual(animation.owner, impl.owner)
```

Sources/MapboxMaps/Camera/BasicCameraAnimator2.swift
```diff
@@ -0,0 +1,105 @@
+import UIKit
+
+public final class BasicCameraAnimator: NSObject, CameraAnimator, CameraAnimatorProtocol {
```
evil159: What is the reasoning to create this wrapper? 

Is it to preserve backwards compatibility? I wonder if converting `BasicCameraAnimator` to a protocol could work in this case 🤔.

Tests/MapboxMapsTests/Camera/BasicCameraAnimatorTests2.swift
macdrevx: This test is one-way since `animator.owner` is read-only.

Sources/MapboxMaps/Camera/BasicCameraAnimator2.swift
macdrevx: This change will allow isolating `CameraAnimationsManager` and `BasicCameraAnimatorImpl` in unit tests (will come in the next PR). Without this change, testing `CameraAnimationsManager` will be difficult since we wouldn't be able to use mocks for `BasicCameraAnimator`.

For internal code, yes, we'd probably just write `CameraAnimationsManager` in terms of `BasicCameraAnimatorProtocol`, but as we push toward v11, I'm recommending that we align on only making protocols public in situations where we would expect end-developers to create their own implementations. Sticking to concrete types in other cases gives us more flexibility to evolve the API without major version bumps.

evil159: Thanks for the great explanation!

Issue #788, opened 2021-10-25
MapboxMap.point(for: CLLocationCoordinate2D) -> CGPoint doesn't work correctly


Ogerets:
## Environment

- Xcode version: 12.5.1
- iOS version: 14.2
- Devices affected: tested on iPhone 6s, iPhone X
- Maps SDK Version: v10.0.1

## Observed behavior and steps to reproduce

When there is a non zero camera pitch, `MapboxMap.point(for: CLLocationCoordinate2D) -> CGPoint` returns totally incorrect CGPoint. When pitch is zero, it's working correctly.
Also zoom level seems to have some influence on this bug. As well as location of camera relative to coordinates.

 [Here](https://github.com/Ogerets/mapbox-convert-issue/tree/v10) is a sample project representing an issue with 100% reproduce rate.
 These are results of converting the **same** camera with different pitch:
 <img width="887" alt="Screenshot 2021-10-25 at 18 06 42" src="https://user-images.githubusercontent.com/20406383/138721428-91b9b1b9-5e52-4fd3-8f10-f12cee82beef.png">

## Expected behavior

Correctly converted point on view from coordinates even with non zero pitch

## Notes

I can see this bug in app with Mapbox v6.4.0. [Here](https://github.com/mapbox/mapbox-gl-native-ios/issues/638) is the similar issue for that version. I think a lot of guys on v6 will appreciate this bug being fixed on older version too.

## Additional links and references

 [Here](https://github.com/Ogerets/mapbox-convert-issue) is a sample project representing an issue

endanke:
@Ogerets originally you mentioned that the issue was first noticed in `v10.0.1`. Do you also have a sample to reproduce it there? We can backport the fix if needed, but primarily we would need to verify the issue in the newer version.

Ogerets:
@endanke Yes, sure. [Here](https://github.com/Ogerets/mapbox-convert-issue/tree/v10) it is. Sorry, I put the link to main branch in the issue description instead of `v10` branch.

endanke:
@Ogerets we've investigated this a bit further, thank you for the examples! Your illustration is on point, it seems like this is pretty much the reason for the issue. Unfortunately it's not a new bug, but a design limitation and it wasn't properly communicated in the documentation of this API.

Currently it's not clear how much effort would it take to rework the implementation of this function. Could you please share some details about your use-case? Maybe we could think about some alternatives to help you out.

Ogerets:
[...]

PR #1195, opened 2022-03-15
Limit MapboxMap.point(for: CLLocationCoordinate2D) to the bounds of map view

endanke:
Related to: https://github.com/mapbox/mapbox-maps-ios/issues/788

Currently we cannot guarantee that the returned point of `MapboxMap.point(for: CLLocationCoordinate2D)` is valid outside of the MapView bounds. Originally the underlying implementation was designed to convert points only on the visible parts of the map. Until the GL-Native implementation is reworked we have to limit the public API. 
https://github.com/mapbox/mapbox-gl-native-internal/issues/3062

## Pull request checklist:

### DIFF
--- a/Sources/MapboxMaps/Foundation/MapboxMap.swift
+++ b/Sources/MapboxMaps/Foundation/MapboxMap.swift
@@ -436,9 +436,11 @@ public final class MapboxMap: MapboxMapProtocol {
 
     /// Converts a map coordinate to a `CGPoint`, relative to the `MapView`.
     /// - Parameter coordinate: The coordinate to convert.
-    /// - Returns: A `CGPoint` relative to the `UIView`.
+    /// - Returns: A `CGPoint` relative to the `UIView`. If the point is outside of the bounds
+    ///     of `MapView` the returned point contains `-1.0` for both coordinates.
     public func point(for coordinate: CLLocationCoordinate2D) -> CGPoint {
-        return __map.pixelForCoordinate(for: coordinate).point
+        let point = __map.pixelForCoordinate(for: coordinate).point
+        return CGRect(origin: .zero, size: size).contains(point) ? point : CGPoint(x: -1.0, y: -1.0)
     }
 
     /// Converts map coordinates to an array of `CGPoint`, relative to the `MapView`.
--- a/Tests/MapboxMapsTests/Foundation/MapboxMapsFoundationTests.swift
+++ b/Tests/MapboxMapsTests/Foundation/MapboxMapsFoundationTests.swift
@@ -55,10 +55,23 @@ class MapboxMapsFoundationTests: XCTestCase {
 
     func testCoordinateToPoint() {
         let centerCoordinate = CLLocationCoordinate2D(latitude: 0, longitude: 0)
-        let convertedPoint = mapView.mapboxMap.point(for: centerCoordinate)
+        var convertedPoint = mapView.mapboxMap.point(for: centerCoordinate)
 
         XCTAssertEqual(convertedPoint.x, mapView.bounds.midX, accuracy: 0.01)
         XCTAssertEqual(convertedPoint.y, mapView.bounds.midY, accuracy: 0.01)
+
+        let maxPoint = CGPoint(x: mapView.bounds.maxX, y: mapView.bounds.maxY)
+        let boundaryCoordinate = mapView.mapboxMap.coordinate(for: maxPoint)
+        convertedPoint = mapView.mapboxMap.point(for: boundaryCoordinate)
+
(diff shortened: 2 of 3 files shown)


endanke:
This would most likely need an Android SDK port too. cc: @mapbox/maps-android

Sources/MapboxMaps/Foundation/MapboxMap.swift
```diff
@@ -436,9 +436,14 @@ public final class MapboxMap: MapboxMapProtocol {
...
-    /// - Returns: A `CGPoint` relative to the `UIView`.
+    /// - Returns: A `CGPoint` relative to the `UIView`. If the point is outside of the bounds
+    ///     of `MapView` the returned point contains `CGFloat.nan` for both coordinates.
     public func point(for coordinate: CLLocationCoordinate2D) -> CGPoint {
-        return __map.pixelForCoordinate(for: coordinate).point
+        let point = __map.pixelForCoordinate(for: coordinate).point
+        if point.x < 0 || point.y < 0 || point.x > size.width || point.y > size.height {
+            return CGPoint(x: CGFloat.nan, y: CGFloat.nan)
+        }
+        return point
```
macdrevx, line 446:
```suggestion
        return CGRect(origin: .zero, size: size).contains(point) ? point : CGPoint(x: .nan, y: .nan)
```

Tests/MapboxMapsTests/Foundation/MapboxMapsFoundationTests.swift
```diff
@@ -55,10 +55,23 @@ class MapboxMapsFoundationTests: XCTestCase {
...
+        convertedPoint = mapView.mapboxMap.point(for: boundaryCoordinate)
+
+        XCTAssertEqual(convertedPoint.x, maxPoint.x, accuracy: 0.01)
+        XCTAssertEqual(convertedPoint.y, maxPoint.y, accuracy: 0.01)
+
+        let outOfBoundsCoordinate = CLLocationCoordinate2D(latitude: boundaryCoordinate.latitude + 1,
+                                                           longitude: boundaryCoordinate.longitude + 1)
+        convertedPoint = mapView.mapboxMap.point(for: outOfBoundsCoordinate)
+        XCTAssert(convertedPoint.x.isNaN)
+        XCTAssert(convertedPoint.y.isNaN)
```
macdrevx, line 74:
```suggestion
        XCTAssertTrue(convertedPoint.x.isNaN)
        XCTAssertTrue(convertedPoint.y.isNaN)
```

macdrevx (approved):
2 minor suggestions. Thanks for the PR!

pengdev:
@endanke if this logic should be fixed for both iOS and Android, wouldn't it be easier to fix on the gl-native side? Adding such logic in the sdk wrapper doesn't seem very good practice.

endanke:
@pengdev 
> If this logic should be fixed for both iOS and Android, wouldn't it be easier to fix on the gl-native side?  Adding such logic in the sdk wrapper doesn't seem very good practice. 

Yeah that's also an option. Technically in some cases we could use the value beyond the bounds of the view. (For example if the pitch is zero. Or if the point is above the bottom side of the MapView.) So the advantage of the platform side limit is that we can use this API in the future for internal purposes if we keep those constraints in mind. But for the public API it's safer to have the limit.
 
> And on Android, we couldn't construct a ScreenCoordinate with null value, we would either need to use -1 to represent invalid input or throw an exception, which way do you think is preferred in this case?

I think -1 would be better, it requires less changes on the customer's side. I could change it here too.

pengdev:
Android counterpart here https://github.com/mapbox/mapbox-maps-android/pull/1226

kushsolitary:
Hello, is there a way of getting back the previous functionality? We rely heavily on out-of-bounds points to make an interaction work, and this is a breaking change. https://twitter.com/HelloSupersonic/status/1518631632038547456

endanke:
@kushsolitary unfortunately we had to disable this functionality for the time being because we cannot guarantee correct behaviour in all cases.
Sorry for the inconvenience, eventually I think this feature will be supported again but there's no estimated date for delivery.
Could you please try to implement a workaround using the suggestion here? https://github.com/mapbox/mapbox-maps-ios/issues/788#issuecomment-960565313

kushsolitary:
> @kushsolitary unfortunately we had to disable this functionality for the time being because we cannot guarantee correct behaviour in all cases. Sorry for the inconvenience, eventually I think this feature will be supported again but there's no estimated date for delivery. Could you please try to implement a workaround using the suggestion here? [#788 (comment)](https://github.com/mapbox/mapbox-maps-ios/issues/788#issuecomment-960565313)

Okay, will check it out. Thanks for the explanation! :)

PR #1200, opened 2022-03-18
Avoid animator-related leaks

### LINKED ISSUE #1119: MapboxMap not deallocated
## Environment

- Xcode version: 13.2
- iOS version: 15.3
- Devices affected: iPhone XS real device
- Maps SDK Version:  v10.3.0

`MapboxMap` is retained in memory if you try to dismiss / replace UIViewController while map has some ongoing animations.

## Expected behavior

`MapboxMap` is not retained in memory and it's `deinit` method is called.

I was able to fix this problem by calling `_mapView.camera.cancelAnimations()` in view models `deinit` function.

macdrevx:
Fixes #1119

* Update `MapView` to disable animations whenever it does not have a display link. See analysis here https://github.com/mapbox/mapbox-maps-ios/issues/1119#issuecomment-1061076964 for more information.
* Update `CameraAnimatorsRunner` to cancel newly-added animators immediately if animations are disabled.
* Introduce `EnablableProtocol` to allow separate components to independently control whether animations are enabled without needing to worry about balancing disable/enable calls. `MapView` and `AnyTouchGestureHandler` are each injected with a unique `MutableEnablableProtocol` that they can use to disable animations, while `CameraAnimatorsRunner` is injected with an `EnablableProtocol` (concretely, a `CompositeEnablable`) that provides it with a view onto the final enabled/disabled decision.
* Relax some constraints on `CameraAnimator` to make it safe for animators to be canceled synchronously when they are being created.
    * Ignore calls to `BasicCameraAnimator.startAnimation()` when the animator is already complete (instead of fatalError).
    * Ignore calls to `BasicCameraAnimator.startAnimation(afterDelay:)` when the animator is running or already complete (instead of fatalError). Calling this method on a paused animator still results in fatalError (matches UIViewPropertyAnimator behavior).
    * Ignore calls to `BasicCameraAnimator.pauseAnimation()` when the animator is already complete (instead of fatalError).
    * Update the documentation for `CameraAnimator.stopAnimation()` and `CameraAnimatorProtocol.startAnimation()` to describe how they should behave when called on animators that are already complete.
    * Update `FlyToCameraAnimator` and `GestureDecelerationCameraAnimator` to behave according to these updated expectations.
* Update `MapViewDependencyProvider` to start to abstract away more of the object graph creation from `MapView`.
* Accept Xcode 13.3 upgrade checks

## Pull request checklist:

### DIFF
--- a/Sources/MapboxMaps/Foundation/MapView.swift
+++ b/Sources/MapboxMaps/Foundation/MapView.swift
@@ -29,6 +29,7 @@ open class MapView: UIView {
     /// The `camera` object manages a camera's view lifecycle.
     public private(set) var camera: CameraAnimationsManager!
     private var cameraAnimatorsRunner: CameraAnimatorsRunnerProtocol!
+    private let cameraAnimatorsRunnerEnablable: MutableEnablableProtocol
 
     /// The `location`object handles location events of the map.
     public private(set) var location: LocationManager!
@@ -194,8 +195,9 @@ open class MapView: UIView {
         self.dependencyProvider = MapViewDependencyProvider()
         self.interfaceOrientationProvider = orientationProvider
         self.attributionUrlOpener = DefaultAttributionURLOpener()
-        notificationCenter = dependencyProvider.makeNotificationCenter()
-        bundle = dependencyProvider.makeBundle()
+        notificationCenter = dependencyProvider.notificationCenter
+        bundle = dependencyProvider.bundle
+        cameraAnimatorsRunnerEnablable = dependencyProvider.cameraAnimatorsRunnerEnablable
         super.init(frame: frame)
         commonInit(mapInitOptions: mapInitOptions, overridingStyleURI: nil)
     }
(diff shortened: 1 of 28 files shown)


Sources/MapboxMaps/Foundation/MapView.swift
```diff
@@ -574,16 +586,29 @@ open class MapView: UIView {
...
             window: window,
             target: ForwardingDisplayLinkTarget { [weak self] in
                 self?.updateFromDisplayLink(displayLink: $0)
             },
             selector: #selector(ForwardingDisplayLinkTarget.update(with:)))
+
+        guard let displayLink = displayLink else {
+            cameraAnimatorsRunner.cancelAnimations()
+            cameraAnimatorsRunnerEnablable.isEnabled = false
+            return
```
maios, line 605: https://github.com/mapbox/mapbox-maps-ios/pull/1200/files#diff-5be8ad5aa3482beb6c6ec989ea84a88e38d60e6a294963b0051c9b9f3eddedbeR613
this line will not be executed now if `displayLink`is nil, if I am not mistaken the previous implementation will still execute this line and subscribe MapView to life cycle events. Is this change intended?

PR #1222, opened 2022-03-29
Update node dependencies

macdrevx:
Fixes dependabot alerts


OdNairy (approved):
What do we need `node` for? 🤔

PR #1215, opened 2022-03-28
Sync annotation examples with android

evil159:
This PR updates the "Annotation" section of Examples app to match with the examples from the same section on Android.

Fixes: https://github.com/mapbox/mapbox-maps-ios-internal/issues/1097

## Pull request checklist:
 - [x] Write tests for all new functionality. If tests were not written, please explain why.
 - [x] Describe the changes in this PR, especially public API changes.
 - [x] Add a changelog entry to to bottom of the relevant section (typically the `## main` heading near the top).
 - [x] Review and agree to the Contributor License Agreement ([CLA](https://github.com/mapbox/mapbox-maps-ios/blob/main/CONTRIBUTING.md#contributor-license-agreement)).

PR #1225, opened 2022-03-30
Add example of map view with debug options

maios:
This pull request add an example of map view with settings for `debugOptions`.

## Pull request checklist:
 - [x] Update the guides (internal access only), README.md, and DEVELOPING.md if their contents are impacted by these changes.

### DIFF
--- a/Apps/Examples/Examples/All Examples/DebugMapExample.swift
+++ b/Apps/Examples/Examples/All Examples/DebugMapExample.swift
@@ -12,28 +12,31 @@ private struct MapDebugOptionSetting {
 
 final class DebugMapExample: UIViewController, ExampleProtocol, DebugOptionSettingsDelegate {
 
-    private var map: MapboxMap!
-
-    override func loadView() {
-        let mapView = MapView(frame: .zero)
-        map = mapView.mapboxMap
-
-        view = mapView
-    }
+    private var mapView: MapView!
 
     override func viewDidLoad() {
         super.viewDidLoad()
 
-        let rightBarButtonItems = navigationItem.rightBarButtonItems ?? []
+        mapView = MapView(frame: view.bounds)
+        view.addSubview(mapView)
+
+        mapView.translatesAutoresizingMaskIntoConstraints = false
+        NSLayoutConstraint.activate([
+            mapView.topAnchor.constraint(equalTo: view.safeAreaLayoutGuide.topAnchor),
+            mapView.leftAnchor.constraint(equalTo: view.safeAreaLayoutGuide.leftAnchor),
+            mapView.bottomAnchor.constraint(equalTo: view.safeAreaLayoutGuide.bottomAnchor),
+            mapView.rightAnchor.constraint(equalTo: view.safeAreaLayoutGuide.rightAnchor),
+        ])
+
         let debugOptionsBarItem = UIBarButtonItem(
             barButtonSystemItem: .edit,
             target: self,
             action: #selector(openDebugOptionsMenu(_:)))
-        navigationItem.rightBarButtonItems = [debugOptionsBarItem] + rightBarButtonItems
+        navigationItem.rightBarButtonItems?.insert(debugOptionsBarItem, at: 0)
     }
 
     @objc private func openDebugOptionsMenu(_ sender: UIBarButtonItem) {
-        let settingsViewController = SettingsViewController(debugOptions: map.debugOptions)
+        let settingsViewController = SettingsViewController(debugOptions: mapView.mapboxMap.debugOptions)
         settingsViewController.delegate = self
 
         let navigationController = UINavigationController(rootViewController: settingsViewController)
@@ -45,7 +48,7 @@ final class DebugMapExample: UIViewController, ExampleProtocol, DebugOptionSetti
 
     fileprivate func debugOptionSettingsDidChange(_ controller: SettingsViewController) {
         controller.dismiss(animated: true, completion: nil)
-        map.debugOptions = controller.enabledDebugOptions
+        mapView.mapboxMap.debugOptions = Array(controller.enabledDebugOptions)
     }
 }
 
@@ -54,7 +57,7 @@ private final class SettingsViewController: UIViewController, UITableViewDataSou
     weak var delegate: DebugOptionSettingsDelegate?
     private var listView: UITableView!
 
-    private(set) var enabledDebugOptions: [MapDebugOptions]
+    private(set) var enabledDebugOptions: Set<MapDebugOptions>
     private let allSettings: [MapDebugOptionSetting] = [
         MapDebugOptionSetting(debugOption: .collision, displayTitle: "Debug collision"),
         MapDebugOptionSetting(debugOption: .depthBuffer, displayTitle: "Show depth buffer"),
@@ -67,7 +70,7 @@ private final class SettingsViewController: UIViewController, UITableViewDataSou
     ]
 
     init(debugOptions: [MapDebugOptions]) {
-        enabledDebugOptions = debugOptions
+        enabledDebugOptions = Set(debugOptions)
         super.init(nibName: nil, bundle: nil)
     }
 
@@ -80,6 +83,7 @@ private final class SettingsViewController: UIViewController, UITableViewDataSou
 
         listView = UITableView()
         listView.dataSource = self
+        listView.register(DebugOptionCell.self, forCellReuseIdentifier: String(describing: DebugOptionCell.self))
 
         view.addSubview(listView)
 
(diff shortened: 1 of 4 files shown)


Apps/Examples/Examples/All Examples/DebugMapExample.swift
```diff
@@ -0,0 +1,179 @@
...
+final class DebugMapExample: UIViewController, ExampleProtocol, DebugOptionSettingsDelegate {
+
+    private var map: MapboxMap!
+
+    override func loadView() {
+        let mapView = MapView(frame: .zero)
+        map = mapView.mapboxMap
+
+        view = mapView
+    }
```
macdrevx, line 22: These view controllers are configured to allow their view to extend under the navigation bar, which means we can't see the top of the map. You'll either need to change that config on this view controller or move this setup to viewDidLoad and add the map as a subview of the view controller's root view instead of making it the root view itself.

Apps/Examples/Examples/All Examples/DebugMapExample.swift
```diff
@@ -0,0 +1,179 @@
...
+}
+
+private struct MapDebugOptionSetting {
+    let debugOption: MapDebugOptions
+    let displayTitle: String
+}
+
+final class DebugMapExample: UIViewController, ExampleProtocol, DebugOptionSettingsDelegate {
+
+    private var map: MapboxMap!
```
macdrevx, line 15: Almost everywhere we refer to this as mapboxMap instead of just map to avoid ambiguity with the `map` function and with things like [`Map`](https://developer.android.com/reference/java/util/Map) on android.

Apps/Examples/Examples/All Examples/DebugMapExample.swift
```diff
@@ -0,0 +1,179 @@
...
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+
+        let rightBarButtonItems = navigationItem.rightBarButtonItems ?? []
+        let debugOptionsBarItem = UIBarButtonItem(
+            barButtonSystemItem: .edit,
+            target: self,
+            action: #selector(openDebugOptionsMenu(_:)))
+        navigationItem.rightBarButtonItems = [debugOptionsBarItem] + rightBarButtonItems
```
macdrevx, line 32:
```suggestion
        let debugOptionsBarItem = UIBarButtonItem(
            barButtonSystemItem: .edit,
            target: self,
            action: #selector(openDebugOptionsMenu(_:)))
        navigationItem.rightBarButtonItems?.insert(debugOptionsBarItem, at: 0)
```

Apps/Examples/Examples/All Examples/DebugMapExample.swift
```diff
@@ -0,0 +1,179 @@
...
+        delegate?.debugOptionSettingsDidChange(self)
+    }
+
+    func tableView(_ tableView: UITableView, numberOfRowsInSection section: Int) -> Int {
+        allSettings.count
+    }
+
+    func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
+        let cellID = String(describing: DebugOptionCell.self)
+        let cell = tableView.dequeueReusableCell(withIdentifier: cellID) as? DebugOptionCell ?? DebugOptionCell()
```
macdrevx, line 116: I recommend registering the cell class for the cell reuse identifier and then using [`dequeueReusableCell(withIdentifier:for:)`](https://developer.apple.com/documentation/uikit/uitableview/1614878-dequeuereusablecell) which returns a non-optional cell. You'll still need a cast, but I think a force-cast would be fine since you control all the inputs.

Apps/Examples/Examples/All Examples/DebugMapExample.swift
```diff
@@ -0,0 +1,179 @@
...
+        map.debugOptions = controller.enabledDebugOptions
+    }
+}
+
+private final class SettingsViewController: UIViewController, UITableViewDataSource {
+
+    weak var delegate: DebugOptionSettingsDelegate?
+    private var listView: UITableView!
+
+    private(set) var enabledDebugOptions: [MapDebugOptions]
```
macdrevx, line 57: Consider using `Set` instead of `Array`

Apps/Examples/Examples/All Examples/DebugMapExample.swift
```diff
@@ -0,0 +1,179 @@
...
+
+// MARK: Cell
+
+private class DebugOptionCell: UITableViewCell {
+
+    private let titleLabel = UILabel()
+    private let toggle = UISwitch()
+
+    init() {
+        super.init(style: .default, reuseIdentifier: String(describing: Self.self))
```
macdrevx, line 140: To do https://github.com/mapbox/mapbox-maps-ios/pull/1225/files#r838898209, you'll probably also need to override `init(style:reuseIdentifier:)` instead since it's the class's designated initializer and will likely be called by UITableView when it needs a new instance.

macdrevx (changes requested):
I like it! Does Android have a comparable example? Our goal is to have the same set of examples on both platforms.

Apps/Examples/Examples/Models/Examples.swift
```diff
@@ -64,6 +64,9 @@ public struct Examples {
         Example(title: "Display a map view using storyboard",
                 description: "Create and display a map using a storyboard.",
                 type: StoryboardMapViewExample.self),
+        Example(title: "Debug Map",
+                description: "This example shows how the map looks like with different debug options",
```
macdrevx, line 68:
```suggestion
                description: "This example shows how the map looks with different debug options",
```

Apps/Examples/Examples/All Examples/DebugMapExample.swift
```diff
@@ -0,0 +1,179 @@
...
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    func configure(with setting: MapDebugOptionSetting, isOptionEnabled: Bool) {
+        titleLabel.text = setting.displayTitle
+        toggle.isOn = isOptionEnabled
+    }
+
+    private var _onToggled: ((Bool) -> Void)?
```
evil159, line 170: Maybe a different name can be used to avoid underscoring the variable? E.g. `onToggledHandler`. Underscored properties are uncharacteristic for Swift code.

evil159:
> I like it! Does Android have a comparable example? Our goal is to have the same set of examples on both platforms.

Yes, Android has an example called "Debug mode" that has similar functionality to toggle debug options.

Apps/Examples/Examples/All Examples/DebugMapExample.swift
maios: I think using name with underscore prefix is fine in Swift, but if you can send me some materials to read more about this it would be great. I do like your suggested name better :-D I made this change, thank you.

Apps/Examples/Examples/All Examples/DebugMapExample.swift
maios: Change suggested in https://github.com/mapbox/mapbox-maps-ios/pull/1225#discussion_r838886458 is made, and now we will access `debugOptions` with `mapView.mapboxMap.debugOptions`.

Apps/Examples/Examples/All Examples/DebugMapExample.swift
evil159: The API Design Guidelines covers this quite well, though it is quite vague sometimes.

I think this convention from the [guidelines](https://www.swift.org/documentation/api-design-guidelines/#general-conventions) applies here:
> Follow case conventions. Names of types and protocols are UpperCamelCase. Everything else is lowerCamelCase.

Apps/Examples/Examples/All Examples/DebugMapExample.swift
```diff
@@ -0,0 +1,183 @@
...
+            mapView.leftAnchor.constraint(equalTo: view.safeAreaLayoutGuide.leftAnchor),
+            mapView.bottomAnchor.constraint(equalTo: view.safeAreaLayoutGuide.bottomAnchor),
+            mapView.rightAnchor.constraint(equalTo: view.safeAreaLayoutGuide.rightAnchor),
+        ])
+
+        let debugOptionsBarItem = UIBarButtonItem(
+            barButtonSystemItem: .edit,
+            target: self,
+            action: #selector(openDebugOptionsMenu(_:)))
+        navigationItem.rightBarButtonItems?.insert(debugOptionsBarItem, at: 0)
```
evil159, line 35: One more thing - don't forget to call `finish()` the end of `viewDidLoad()`. Unit tests expect `finish()` to be called sometime after the example is set up, and changes in #1215 will fail the test if this expectation is not met.

evil159 (approved):

macdrevx (approved):

Issue #207, opened 2021-03-24
Expose utility functions for CoordinateBounds


kiryldz:
## New Feature

Expose functions: `getCenter`, `contains`, `union`, `intersect`, `containsLatitude`, `containsLongitude`, `include`, `getLatitudeSpan`, `getLongitudeSpan` as extension functions for `CoordinateBounds`.

Full definition of Android `LatLngBounds`(ancestor for `CoordinateBounds`) could be found [here](https://github.com/mapbox/mapbox-gl-native-android/blob/main/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/geometry/LatLngBounds.java) on Android platform. iOS equivalent for it in previous Maps version is `MGLCoordinateBounds` that also was exposing some functions but not truly aligned with Android.

## Why

Better parity with Android platform.

NasH5169:
Hi all,

Any news on this topic?

Thanks!

tobrun:
@macdrevx this landed upstream, is there any additional work left?

macdrevx:
Yes, we intend to refine some of the new utilities for Swift.

PR #1226, opened 2022-03-30
Expose coordinate bounds methods

evil159:
This PR exposes some utility methods/properties for `CoordinateBounds` to public.

Additionally to the methods outlined in https://github.com/mapbox/mapbox-maps-ios/issues/207, `east`, `west`, `north`, `south`, `isEmpty` were exposed together with equality and hashability support.

Fixes: https://github.com/mapbox/mapbox-maps-ios/issues/207

## Pull request checklist:


Sources/MapboxMaps/Foundation/Extensions/Core/CoordinateBounds.swift
```diff
@@ -5,27 +5,106 @@ import MapboxCoreMaps
...
+
+    /// Returns whether the bounds are empty or not.
+    var isEmpty: Bool { __isEmpty() }
+
+    /// The absolute distance, in degrees, between the north and south boundaries of these bounds.
+    ///
+    /// One degree of latitude is approximately 111 kilometers (69 miles).
+    var latitudeSpan: CLLocationDegrees {
+        __latitudeSpan()
+    }
```
macdrevx, line 34: inconsistent formatting. please be consistent about whether to include `return` and whether to wrap or write all on one line

Sources/MapboxMaps/Foundation/Extensions/Core/CoordinateBounds.swift
```diff
@@ -5,27 +5,106 @@ import MapboxCoreMaps
...
-        for coordinate in coordinates {
-            if latitudeRange.contains(coordinate.latitude) || longitudeRange.contains(coordinate.longitude) {
-                return true
-            }
+    /// Extends these bounds to include `point`.
+    /// - Parameter coordinate: The coordinate to be included.
+    /// - Returns: The extended bounds.
+    func include(_ coordinate: CLLocationCoordinate2D) -> CoordinateBounds {
+        extend(forPoint: coordinate)
+    }
```
macdrevx, line 67: Can you confirm whether these functions return new instances or mutate self?

I'd also like to workshop on these names with you a bit to think through consistency with Android and adherence to the Swift API Design Guidelines.

Sources/MapboxMaps/Foundation/Extensions/Core/CoordinateBounds.swift
```diff
@@ -5,27 +5,106 @@ import MapboxCoreMaps
...
+
+    override func isEqual(_ object: Any?) -> Bool {
+        guard let other = object as? CoordinateBounds else {
+            return false
         }
-        return false
+
+        return self == other
+    }
```
macdrevx, line 102: I'm wary of overriding methods in extensions. Is it safe?

macdrevx: Seems it's allowed only for extensions of Objective-C types. What's the correct mental model for how it works? What happens if the same type has its own implementation?

Sources/MapboxMaps/Foundation/Extensions/Core/CoordinateBounds.swift
evil159: They return new instances.

Yes, the inconsistency between `union` and `include` is puzzling, given that they perform similar things.

Sources/MapboxMaps/Foundation/Extensions/Core/CoordinateBounds.swift
```diff
@@ -5,27 +5,106 @@ import MapboxCoreMaps
...
+    /// - Parameter coordinate: The coordinate to be included.
+    /// - Returns: The extended bounds.
+    func include(_ coordinate: CLLocationCoordinate2D) -> CoordinateBounds {
+        extend(forPoint: coordinate)
+    }
+
+    /// Returns a new `CoordinateBounds` that is the intersection of this with another `CoordinateBounds`.
+    /// - Parameter coordinateBounds: The `CoordinateBounds` to intersect with.
+    /// - Returns: The bounds representing the intersection of the two bounds or null if there is no intersection.
+    func intersect(_ coordinateBounds: CoordinateBounds) -> CoordinateBounds? {
```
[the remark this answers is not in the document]
evil159: It's here https://github.com/mapbox/mapbox-gl-native-android/blob/d25d1335f97cfcdbb45deab50d384cc4a5ecf8ad/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/geometry/LatLngBounds.java#L454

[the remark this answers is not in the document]
macdrevx: I believe that's a link to android v9

Sources/MapboxMaps/Foundation/Extensions/Core/CoordinateBounds.swift
evil159: Removed as a result of our workshop.

Sources/MapboxMaps/Foundation/Extensions/Core/CoordinateBounds.swift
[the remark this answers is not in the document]
evil159: 🤦

Sources/MapboxMaps/Foundation/Extensions/Core/CoordinateBounds.swift
evil159: Not that safe it seems:

> If the name of a method declared in a category is the same as a method in the original class, or a method in another category on the same class (or even a superclass), the behavior is undefined as to which method implementation is used at runtime. This is less likely to be an issue if you’re using categories with your own classes, but can cause problems when using categories to add methods to standard Cocoa or Cocoa Touch classes.

From [here](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/ProgrammingWithObjectiveC/CustomizingExistingClasses/CustomizingExistingClasses.html), removed them.

macdrevx (approved):

PR #1227, opened 2022-03-31
Add SimpleCameraAnimator

macdrevx:
`SimpleCameraAnimator` is a camera animator that offers cubic bezier easing, delayed start, and a dynamically-updatable target camera.

This animator has some overlap in functionality with `BasicCameraAnimator`; however, since it uses direct interpolation rather than interpolation via `CameraView`, it has a simpler implementation and enables more advanced use cases like dynamically updating `SimpleCameraAnimator.to`.

This animator will be used in a subsequent PR to solve the moving target problem in the Viewport API.

## Pull request checklist:

### DIFF
--- a/Sources/MapboxMaps/Camera/SimpleCameraAnimator.swift
+++ b/Sources/MapboxMaps/Camera/SimpleCameraAnimator.swift
@@ -172,9 +172,8 @@ internal final class SimpleCameraAnimator: NSObject, SimpleCameraAnimatorProtoco
     }
 
     private func invokeCompletionBlocks(with position: UIViewAnimatingPosition) {
-        let handlers = completionHandlers
-        for handler in handlers {
-            handler(position)
+        for completionHandler in completionHandlers {
+            completionHandler(position)
         }
         completionHandlers.removeAll()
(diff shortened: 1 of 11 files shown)


Sources/MapboxMaps/Camera/SimpleCameraAnimator.swift
```diff
@@ -0,0 +1,214 @@
...
+        case .initial, .running:
+            completionHandlers.append(completion)
+        case .final:
+            // already stopped, so do nothing
+            break
+        }
+    }
+
+    private func invokeCompletionBlocks(with position: UIViewAnimatingPosition) {
+        let handlers = completionHandlers
```
evil159, line 175: What is the reason to capture handlers here? Multithreading?

evil159 (approved):

Sources/MapboxMaps/Camera/SimpleCameraAnimator.swift
macdrevx: Good call. It's unnecessary.

PR #1234, opened 2022-04-01
Add high-level API to set `Light`

evil159:
This PR adds a method to set `Light` to `Style`, as well as a method to remove it. Additionally, `BuildingExtrusionsExample` is updated with buttons to change light source and colour, keeping the example in sync with Android.

Fixes: https://github.com/mapbox/mapbox-maps-ios-internal/issues/1090

## Pull request checklist:


Sources/MapboxMaps/Style/Style.swift
```diff
@@ -744,6 +744,26 @@ public final class Style: StyleProtocol {
...
+        try setLight(properties: lightDictionary)
+    }
+
+    /// Removes light from style if it was set.
+    public func removeLight() throws {
+        try handleExpected {
+            styleManager.setStyleLightForProperties([:])
+        }
+    }
+
```
macdrevx, line 766: Light cannot be removed from a style. See https://github.com/mapbox/mapbox-gl-native-internal/issues/2649#issuecomment-989932206

macdrevx (changes requested):
Need to remove `removeLight()` since removing light is not supported.

PR #1235, opened 2022-04-01
Update MapboxCoreMaps and MapboxCommon

evil159:
#### Breaking changes ⚠️
* Update automatic transition between the globe and mercator projection to appear visually more subtle. ([#3049](https://github.com/mapbox/mapbox-gl-native-internal/pull/3049))
* Add support for parsing map projection from the style ([#3119](https://github.com/mapbox/mapbox-gl-native-internal/pull/3119))

#### Features ✨ and improvements 🏁
* Decrease time used for render three preparation ([#2907](https://github.com/mapbox/mapbox-gl-native-internal/pull/2907))
* Avoid unnecessary layer properties transition calculation if the layer is transitioning to the same constant value or if transitioning from/to data-driven property ([#2845](https://github.com/mapbox/mapbox-gl-native-internal/pull/2845))
* Introduce a new line layer paint property '{"line-trim-offset", [trim-start, trim-end]}', which takes the line trim-off percentage range based on the whole line range [0.0, 1.0]. The property will only be effective when 'line-gradient' property is set. The line part between [trim-start, trim-end] will be marked as transparent to make a line gradient a vanishing effect. If either 'trim-start' or 'trim-end' offset is out of valid range, the default range [0.0, 0.0] will be set.  ([#2780](https://github.com/mapbox/mapbox-gl-native-internal/pull/2780))
* Add target frame timestamp to metal backend ([#2967](https://github.com/mapbox/mapbox-gl-native-internal/pull/2967))
* Revamp globe view controls for more intuitive interaction with touch controls ([#3059](https://github.com/mapbox/mapbox-gl-native-internal/pull/3059))
* Add new map debug option to display model bounding boxes. ([#3104](https://github.com/mapbox/mapbox-gl-native-internal/pull/3104))
* Update mapbox-common to v21.3.0-beta.2 ([#3137](https://github.com/mapbox/mapbox-gl-native-internal/pull/3137))
* [Experimental] Request localized map content resources when language and/or worldview settings are set ([#2769](https://github.com/mapbox/mapbox-gl-native-internal/pull/2769))
* Added terrain wireframe debug option ([#3133](https://github.com/mapbox/mapbox-gl-native-internal/pull/3133))
* Add OfflineRegion::getStatus() API, so that the clients can get the completion status and the local size of the existing legacy offline regions. ([#3168](https://github.com/mapbox/mapbox-gl-native-internal/pull/3168))

#### Bug fixes 🐞
* Dispatched in-flight events will not be delivered if 'unsubscribe' is called before an event is delivered ([#2976](https://github.com/mapbox/mapbox-gl-native-internal/pull/2976))
* Fix an issue where some of the visible tiles could be erroneously culled during transition between globe and mercator projection. ([#3047](https://github.com/mapbox/mapbox-gl-native-internal/pull/3047))
* Disable map LOD for the camera pitch less than 30 degrees and thus avoid map content missing on maps with insets  ([#3099](https://github.com/mapbox/mapbox-gl-native-internal/pull/3099))
* Fixes issues where camera appears under terrain, or map gets bumpy repositioning after exaggeration change. ([#3082](https://github.com/mapbox/mapbox-gl-native-internal/pull/3082))
* Disable terrain rendering if GPU does not support Vertex Texture Fetch ([#3123](https://github.com/mapbox/mapbox-gl-native-internal/pull/3123))
* Fixes a bug that occasionally prevents symbols from loading. ([#3174](https://github.com/mapbox/mapbox-gl-native-internal/pull/3174))

#### Features ✨ and improvements 🏁

* [Swift] Add Swift extension infrastructure ([#2528](https://github.com/mapbox/mapbox-sdk-common/pull/2528))
* Enable events service API by default in the Common SDK so that the Native SDKs like GL Native and Navigation Native can start using to send turnstile and feedback events to our events endpoint. ([#2507](https://github.com/mapbox/mapbox-sdk-common/pull/2507))
* [settings] Implement persistent/nonpersistent key/value storage API (Settings API) ([#2408](https://github.com/mapbox/mapbox-sdk-common/pull/2408))
* [settings] Add common keys for localization ([#2568](https://github.com/mapbox/mapbox-sdk-common/pull/2568))
* Enable i18n support for TileStore ([#2367](https://github.com/mapbox/mapbox-sdk-common/pull/2367))

#### Bug fixes 🐞

* [logging, linux] Fix retrieval of the name of the process ([#2544](https://github.com/mapbox/mapbox-sdk-common/pull/2544))
* [http, linux] Fix flaky `HttpService.OfflineSwitchOfflineOnline` test case ([#2539](https://github.com/mapbox/mapbox-sdk-common/pull/2539))
* [telemetry] Do not include empty model property in the payload ([#2517](https://github.com/mapbox/mapbox-sdk-common/pull/2517))

#### 💫️ Other

* Update Mapbox Bindgen to 23.1.1. Noteworthy changes since the last update include a fix for extracted Java interfaces access modifiers, generation of type annotation for MBXExpected in Objective-C and a fix for documentation generation for callbacks and variants in C++. See the [full changelog](https://github.com/mapbox/mapbox-bindgen/blob/334af6f17f468b85dad018a3537150e04e60dbf4/CHANGELOG.md) for more details. ([#2580](https://github.com/mapbox/mapbox-sdk-common/pull/2580))

## Pull request checklist:


CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ Mapbox welcomes participation and contributions from everyone.
 * Update annotation examples. ([#1215](https://github.com/mapbox/mapbox-maps-ios/pull/1215))
 * Remove `FollowPuckViewportStateOptions.animationDuration`, a workaround for the moving target problem. ([#1228](https://github.com/mapbox/mapbox-maps-ios/pull/1228))
 * Deprecate `FollowPuckViewportStateOptions.animationDuration`, a workaround for the moving target problem. ([#1228](https://github.com/mapbox/mapbox-maps-ios/pull/1228))
+* Update to MapboxCoreMaps 10.5.0-beta.1 and MapboxCommon 21.3.0-beta.2. ([]())
```
macdrevx, line 18: Remember to finish adding the link

Sources/MapboxMaps/Foundation/Extensions/Core/DownloadStatus.swift
```diff
@@ -40,7 +40,7 @@ extension DownloadStatus {
                             downloadOptions: DownloadOptions,
                             httpResult: Result<HttpResponseData, HttpRequestError>?) {
 
-        let expected: Expected<AnyObject, AnyObject>?
+        let expected: Expected<HttpResponseData, HttpRequestError>?
```
macdrevx, line 43: Is there a reason we need to combine the dependency version upgrade and adoption of lightweight generics? I would like to do some more thorough testing around lightweight generics before we adopt it, and I don't want to block the upgrade while we do that.

Sources/MapboxMaps/Foundation/MapProjection.swift
```diff
@@ -6,7 +6,12 @@
 /// Describes the projection used to render the map.
 ///
 /// Mapbox map supports Mercator and Globe projections.
-@_spi(Experimental) public enum MapProjection: Codable, Hashable {
+@_spi(Experimental) public enum MapProjection: Codable, Hashable, CaseIterable {
+    public static var allCases: [MapProjection] = [.mercator(), .globe()]
+    enum CodingKeys: String, CodingKey {
+        case name
+    }
+
```
macdrevx, line 14: Why are these changes necessary?

Sources/MapboxMaps/Foundation/MapboxMap.swift
```diff
@@ -1008,15 +1008,17 @@ extension MapboxMap {
 
 extension MapboxMap {
     /// Errors related to MapProjection API
+    @available(iOS, obsoleted: 11.0)
     @_spi(Experimental) public enum MapProjectionError: Error {
         case unsupportedProjection
     }
```
macdrevx, line 1014: What's the benefit of `obsoleted` vs just deleting it?

Sources/MapboxMaps/Style/Style.swift
```diff
@@ -964,6 +964,38 @@ extension Style {
...
+    ///
+    /// - Returns:
+    ///     `MapProjection` map is using.
+    /// - Throws: `MapProjectionError.unsupportedProjection` if the projection is not compatible with the SDK.
+    @_spi(Experimental) public func mapProjection() throws -> MapProjection {
+        let namePropertyName = MapProjection.CodingKeys.name.rawValue
+        let projectionName = styleManager.getStyleProjectionProperty(forProperty: namePropertyName).value as? String
+
+        guard let projection = MapProjection.allCases.first(where: { projectionName == $0.name }) else {
+            throw MapProjectionError.unsupportedProjection
```
macdrevx, line 991: This is marked as obsolete. Maybe it should just be transferred into Style?

Sources/MapboxMaps/Foundation/MapProjection.swift
macdrevx: I see now. I opened a discussion with GL Native team about this. Even if we don't change GL Native for the beta, I'd like to avoid adding the allCases as public API (let it be an implementation detail in `Style.mapProjection()`) since this enum is designed to be able to be extended with additional projections that may contain configuration options.

Sources/MapboxMaps/Foundation/MapboxMap.swift
```diff
@@ -1008,15 +1008,17 @@ extension MapboxMap {
...
 
     /// Set map projection for Mapbox map.
     /// - Parameter mode: The `MapProjection` to be used by the map.
     /// - Throws: Errors during encoding or `MapProjectionError.unsupportedProjection` if the supplied projection is not compatible with the SDK.
+    @available(*, deprecated, message: "Use `Style.setMapProjection(_:)` instead")
```
macdrevx, line 1019: Would `renamed:` work in this situation?

macdrevx (commented):
Done with comments.

Sources/MapboxMaps/Foundation/Extensions/Core/DownloadStatus.swift
evil159: After the update every single `Expected<AnyObject, AnyObject>` was causing a compilation error, e.g.:

> Cannot convert value of type '(Expected<AnyObject, AnyObject>?) -> Void' to expected argument type 'TileRegionGeometryCallback' (aka '(Expected<Geometry, TileRegionError>) -> ()')

As now Common requires `Expected` specialised with concrete types rather than `AnyObject`.

Sources/MapboxMaps/Foundation/MapboxMap.swift
evil159: Well spotted, I committed this accidentally after experimenting with `deprecated` vs `obsoleted` for these symbols. Changed it back to deprecated.

Sources/MapboxMaps/Style/Style.swift
evil159: There already is a top level `MapProjectionError` in `MapProjection.swift` file, which is used here.

Sources/MapboxMaps/Foundation/MapboxMap.swift
evil159: Interesting

> renamed with a keypath to another API; __when provided, Xcode provides an automatic “fix-it”__

This would definitely be useful here

evil159: Works like a charm - Xcode displays a warning that the method is deprecated and, indeed, provides a "fix-it" 🔥

scripts/release/packager/project.yml
```diff
@@ -18,7 +18,7 @@ targets:
         DEFINES_MODULE: "YES"
         PRODUCT_NAME: $(TARGET_NAME:c99extidentifier)
         CODE_SIGN_STYLE: Automatic
-        SWIFT_TREAT_WARNINGS_AS_ERRORS: "YES"
+        SWIFT_TREAT_WARNINGS_AS_ERRORS: "NO"
```
evil159, line 21: I disabled this as it was causing the `build-sdk-for-api-compatibility-check` step to fail  https://app.circleci.com/pipelines/github/mapbox/mapbox-maps-ios/6562/workflows/5e6fe8c2-94ba-4bc2-831c-4529b57e322f/jobs/61993

Sources/MapboxMaps/Foundation/MapProjection.swift
```diff
@@ -1,118 +0,0 @@
...
-/// Conceptually it is the undistorted and unskewed “ground truth” view of the map
-/// that preserves true proportions between different areas of the map.
-///
-/// Some layers are not supported when map is in Globe projection:
-///  - custom
-///  - fill extrusion
-///  - location indicator
-///
-/// If Globe projection is set it will be switched automatically to Mercator projection
-/// when passing `GlobeMapProjection.transitionZoomLevel` during zooming in.
```
macdrevx, line 82: I'm not seeing this behavior anymore. @mpulkki-mapbox @kiryldz can you confirm that this is no longer relevant?

Sources/MapboxMaps/Foundation/MapProjection.swift
```diff
@@ -1,118 +0,0 @@
...
-    ///
-    /// If `MapboxMap` is using `MapProjection.globe` and current map zoom level is >= `GlobeMapProjection.transitionZoomLevel` -
-    /// map will use `MapProjection.mercator` and `MapboxMap.getMapProjection` will return `MapProjection.mercator`.
-    ///
-    /// If `MapboxMap` is using `MapProjection.globe` and current map zoom level is < `GlobeMapProjection.transitionZoomLevel` -
-    /// map will use `MapProjection.globe` and `MapboxMap.getMapProjection` will return `MapProjection.globe`.
-    ///
-    /// If `MapboxMap` is using `MapProjection.mercator` - map will use `MapProjection.mercator` for any zoom level and
-    /// `MapboxMap.getMapProjection` will return `MapProjection.mercator`.
-    public static let transitionZoomLevel: CGFloat = 5.0
```
macdrevx, line 117: Removed. See https://github.com/mapbox/mapbox-maps-android/pull/1255#discussion_r843197159

Sources/MapboxMaps/Foundation/MapboxMap.swift
```diff
@@ -972,68 +970,38 @@ extension MapboxMap {
...
-    ///
-    /// - Returns:
-    ///     `MapProjection` map is using.
-    /// - Throws: Errors during decoding
-    @_spi(Experimental) public func mapProjection() throws -> MapProjection {
-        let data = try JSONSerialization.data(withJSONObject: __map.getMapProjection(), options: [])
-        return try JSONDecoder().decode(MapProjection.self, from: data)
-    }
-}
-
```
macdrevx, line 1036: Removing instead of deprecating in keeping with Android. https://github.com/mapbox/mapbox-maps-android/pull/1255#discussion_r842709350

Sources/MapboxMaps/Style/Style.swift
```diff
@@ -964,6 +964,34 @@ extension Style {
...
+
+    /// The current projection.
+    public var projection: StyleProjection {
+        let projectionName = styleManager.getStyleProjectionProperty(
+            forProperty: StyleProjection.CodingKeys.name.rawValue)
+        if projectionName.kind == .undefined {
+            return StyleProjection(name: .mercator)
+        } else {
+            return StyleProjection(name: StyleProjectionName(rawValue: projectionName.value as! String))
+        }
```
macdrevx, line 991: matches Android impl

Sources/MapboxMaps/Style/StyleProjection.swift
```diff
@@ -0,0 +1,44 @@
...
+    public var name: StyleProjectionName
+
+    /// Initializes a projection
+    public init(name: StyleProjectionName) {
+        self.name = name
+    }
+
+    internal enum CodingKeys: String, CodingKey {
+        case name
+    }
```
macdrevx, line 14: not necessary for Codable conformance, but I wanted to refer to its raw value elsewhere

scripts/release/packager/project.yml
macdrevx: I don't see any other way around this

macdrevx (approved):
Approving. @evil159 please review the changes I made before merging.

cc: @kiryldz I think the iOS implementation is a bit closer to Android now too.

Sources/MapboxMaps/Style/StyleProjection.swift
```diff
@@ -0,0 +1,44 @@
...
+
+    /// The name of the globe projection
+    ///
+    /// Globe projection is a custom map projection mode for rendering the map wrapped around a full 3D globe.
+    /// Conceptually it is the undistorted and unskewed “ground truth” view of the map
+    /// that preserves true proportions between different areas of the map.
+    ///
+    /// Some layers are not supported when map is in globe projection:
+    ///  - custom
+    ///  - fill extrusion
```
kiryldz, line 41: discussed this with @mpulkki-mapbox and fill extrusion is supported already, we need to remove it from here and I'll also add similar docs for Android impl as we're missing that

Sources/MapboxMaps/Style/StyleProjection.swift
```diff
@@ -0,0 +1,44 @@
...
+        self.name = name
+    }
+
+    internal enum CodingKeys: String, CodingKey {
+        case name
+    }
+}
+
+/// The name of a style projection as defined in the Mapbox Style Spec
+public struct StyleProjectionName: RawRepresentable, Hashable, Codable {
```
maios, line 18: Is there a reason we are using struct instead of enum here?

Sources/MapboxMaps/Foundation/MapProjection.swift
kiryldz: time-based transition at `transitionZoomLevel` was replaced with zoom-based transition at certain zoom levels (from zoom 5.0 to 6.0) but main difference here is that getter in globe projection previously did return mercator when zoom > transitionZoomLevel but now globe is returned in any case. Thus exposing that constant is no longer relevant as it's useless in any kind. Perhaps in the future we could make that zoom range transition configurable somehow but not now for sure.

Sources/MapboxMaps/Style/StyleProjection.swift
macdrevx: struct should be easier to extend without breaking changes. adding a case to an non-frozen-enum is source-breaking.

Sources/MapboxMaps/Foundation/MapProjection.swift
```diff
@@ -7,6 +7,10 @@
 ///
 /// Mapbox map supports Mercator and Globe projections.
 @_spi(Experimental) public enum MapProjection: Codable, Hashable {
```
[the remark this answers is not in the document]
macdrevx: StyleProjection is Codable to give us the flexibility to switch to a codable-based implementation in the future if we end up adding any config options besides name. Protocol conformances need to be introduced in the same release as the conforming type to avoid breaking changes.

PR #1245, opened 2022-04-05
Add initial solution to moving target problem

macdrevx:
This PR updates the DefaultViewportTransition to animate using SimpleCameraAnimator and update the animator's target camera dynamically during the animation. This approach results in arriving at exactly the right location at the end of the transition.

Before:
https://user-images.githubusercontent.com/205658/161695243-fd8c1204-278e-4c48-86fe-46cb8639f5e6.mp4

After:
https://user-images.githubusercontent.com/205658/161695273-eb3b3467-21be-4c0f-9d12-9b0821f7c5e1.mp4

The high-to-low and low-to-high animation spec providers are not yet tested. Those tests will be added in a subsequent PR.

## Pull request checklist:

### DIFF
--- a/Sources/MapboxMaps/Viewport/Transitions/Default/CameraOptionsComponent.swift
+++ b/Sources/MapboxMaps/Viewport/Transitions/Default/CameraOptionsComponent.swift
@@ -1,0 +1,21 @@
+internal protocol CameraOptionsComponentProtocol {
+    var cameraOptions: CameraOptions { get }
+    func updated(with cameraOptions: CameraOptions) -> CameraOptionsComponentProtocol?
+}
+
+internal struct CameraOptionsComponent<T>: CameraOptionsComponentProtocol {
+    internal let keyPath: WritableKeyPath<CameraOptions, T?>
+    internal let value: T
+
+    internal var cameraOptions: CameraOptions {
+        var cameraOptions = CameraOptions()
+        cameraOptions[keyPath: keyPath] = value
+        return cameraOptions
+    }
+
+    internal func updated(with cameraOptions: CameraOptions) -> CameraOptionsComponentProtocol? {
+        cameraOptions[keyPath: keyPath].map {
+            CameraOptionsComponent(keyPath: keyPath, value: $0)
+        }
+    }
+}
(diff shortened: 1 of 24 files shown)


Sources/MapboxMaps/Viewport/Transitions/Default/CameraOptionsComponent.swift
```diff
@@ -0,0 +1,21 @@
+internal protocol CameraOptionsComponentProtocol {
+    var cameraOptions: CameraOptions { get }
+    func updated(with cameraOptions: CameraOptions) -> CameraOptionsComponentProtocol?
+}
+
+internal struct CameraOptionsComponent<T>: CameraOptionsComponentProtocol {
+    internal let keyPath: WritableKeyPath<CameraOptions, T?>
```
maios, line 7: does it have to be a key path to an optional property of `CameraOptions`?  I see that CameraOptions has everything as Optional now, but if in the future we decide to add new property that is non-optional then this will not work for those properties.

Sources/MapboxMaps/Viewport/Transitions/Default/DefaultViewportTransition.swift
```diff
@@ -26,13 +26,21 @@ extension DefaultViewportTransition: ViewportTransition {
     public func run(to toState: ViewportState,
                     completion: @escaping (Bool) -> Void) -> Cancelable {
         let resultCancelable = CompositeCancelable()
+        var animation: DefaultViewportTransitionAnimationProtocol?
```
maios, line 29: If we know that we are going to intialize animation later before we use it, can we use `!` here instead?

maios (approved):

evil159 (approved):
Nice architecture!

Sources/MapboxMaps/Viewport/Transitions/Default/CameraOptionsComponent.swift
macdrevx: It wouldn't have to be. Doing it this way allows `value` (on the next line) to be non-optional.

The `*Options` naming pattern in GL Native typically indicates a type where all of its fields are optional. You pass instances to a setter and non-nil fields are updated and any nil fields are not modified. Given that, it's probably unlikely that we'll add non-optional fields to `CameraOptions` in the future.

PR #1246, opened 2022-04-05
[i18n] Add server side localization check and example

evil159:
This PR adds a warning when attempting to localise the map via `Style.localiseLabels(into:forLayerIds:)` while having the server-side localization enabled via `SettingsServiceInterface.set(key:value:)`. 

Additionally `LocalizationSettingsExample` has been added to demonstrate how to turn on the server-side localization.

## Pull request checklist:


Sources/MapboxMaps/Style/Style+Localization.swift
```diff
@@ -31,6 +36,23 @@ extension Style {
         }
     }
 
+    private func isServerSideLocalizationSet() -> Bool {
+        let settings = SettingsServiceFactory.getInstance(storageType: .nonPersistent)
```
evil159, line 40: This check doesn't work when mixing settings storage types. i.e. when language/worldview is set to persistent storage on the app side. In this case this check returns `false`. I wonder if we need to check both storage types here 🤔 
@alexshalamov any thoughts?

maios (approved):

ank27:
Just a note : this will not be included in 10.5.0-beta.1 cc @evil159 @macdrevx

Sources/MapboxMaps/Style/Style+Localization.swift
```diff
@@ -31,6 +36,23 @@ extension Style {
...
+            return true
+        }
+
+        let worldviewResult = settings.has(key: MapboxCommonSettings.worldview)
+
+        if let isWorldviewSet = try? worldviewResult.get(), isWorldviewSet {
+            return true
+        }
+
+        return false
```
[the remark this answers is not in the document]
OdNairy: Can we than write it this way?

```swift
return settings.has(key: MapboxCommonSettings.language) == .success(true) || 
       settings.has(key: MapboxCommonSettings.worldview) == .success(true)
```

Apps/Examples/Examples/All Examples/Localization/LocalizationSettingsExample.swift
```diff
@@ -0,0 +1,175 @@
...
+/// parameters are supported.
+///
+/// In order to get localized tiles (the server side localization), create a Setting service instance using SettingsServiceFactory,
+/// and set `MapboxCommonSettings.language` or `MapboxCommonSettings.worldview`.
+///
+/// `SettingsServiceInterface` stores the provided language/worldview as a key:value pair in `UserDefaults`
+/// and try to apply it to the tiles. If the values are supported, the localized tiles will be loaded onto the map,
+/// otherwise `MapEvents.EventKind.mapLoadingError` event will be emitted.
+///
+/// The language parameter should be one of the BCP-47 tags and the worldview parameter one of the ISO 3166-1 alpha-2 codes.
```
OdNairy, line 15: > should be one of the BCP-47 tags

Can we make an example here? I don't expect many developers known what is the BCP-47 is 🙂 

> one of the ISO 3166-1 alpha-2 codes

Same here

Apps/Examples/Examples/All Examples/Localization/LocalizationSettingsExample.swift
```diff
@@ -0,0 +1,175 @@
...
+            return UIAlertAction(title: title, style: .default) { _ in
+                do {
+                    // set the worldview setting according to the language selected by the user
+                    try settings.set(key: MapboxCommonSettings.worldview, value: worldview).get()
+                    sender.setTitle("Worldview: \(worldview)", for: .normal)
+                } catch {
+                    print("Failed to set the worldview, error: \(error.localizedDescription)")
+                }
+            }
+        }.forEach(alert.addAction(_:))
```
OdNairy, line 169: [discussion, nit] 
It's probably just me, but I'm not a big fun of the combination `map + forEach`.
Wouldn't common for-in look a bit cleaner?

evil159:
not required now

PR #1260, opened 2022-04-07
Use literal in PointAnnotationManager.textFont

macdrevx:
`[String]` is interpreted as an expression unless it is wrapped in a literal expression.

## Pull request checklist:

### DIFF
--- a/Sources/MapboxMaps/Annotations/Generated/PointAnnotationManager.swift
+++ b/Sources/MapboxMaps/Annotations/Generated/PointAnnotationManager.swift
@@ -317,10 +317,10 @@ public class PointAnnotationManager: AnnotationManagerInternal {
     /// Font stack to use for displaying text.
     public var textFont: [String]? {
         get {
-            return layerProperties["text-font"] as? [String]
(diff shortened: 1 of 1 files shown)


Sources/MapboxMaps/Annotations/Generated/PointAnnotationManager.swift
```diff
@@ -317,10 +317,10 @@ public class PointAnnotationManager: AnnotationManagerInternal {
     /// Font stack to use for displaying text.
     public var textFont: [String]? {
         get {
-            return layerProperties["text-font"] as? [String]
+            return (layerProperties["text-font"] as? [Any])?[1] as? [String]
```
evil159, line 320: Is it safe to assume that the array contains at least 2 elements?

macdrevx: Yes. The value for this key is only set in the corresponding setter to either nil or a length-2 array.

evil159 (approved):

PR #1269, opened 2022-04-13
Resizable image support

evil159:
This PR provides an API to add a resizable image to the style without calculating `stretchX` and `stretchY`. 

The resizable image can be defined either in Xcode asset catalog or programmatically, e.g.:
```swift
let image = UIImage(named: "circle")! // a 25x25 pt image of a colour filled circle
    .resizableImage(withCapInsets: UIEdgeInsets(top: 12, left: 12, bottom: 12, right: 12))
```

Which would produce the following result:
 

https://user-images.githubusercontent.com/1478430/163143890-6910f068-fabe-475f-a578-ba4b33adeeee.mp4

### Public API additions to Style

The method to add a resizable image
```swift
public func addImageResizable(
    _ image: UIImage,
    id: String,
    sdf: Bool = false,
    content: ImageContent? = nil) throws
```

A convenience method for adding regular images
```swift
public func addImage(
    _ image: UIImage,
    id: String,
    sdf: Bool = false,
    content: ImageContent? = nil) throws
```

## Pull request checklist:
     Tests are not written as Style is not particularly testable at this point


ZiZasaurus (approved):
LGTM! I also like the conscious change to `addImageResizeable` as it will make it easier for customers to encounter in the autocomplete.

Apps/Examples/Examples.xcodeproj/project.pbxproj
```diff
@@ -421,6 +424,14 @@
 			path = Radar;
 			sourceTree = "<group>";
 		};
+		7406E36E2806B659002CC41F /* Lab */ = {
```
macdrevx, line 427: What does this mean?

macdrevx (commented):
Do we need separate methods for addImage and addImageResizable? What if we just offered addImage and made it respect cap insets by default (and potentially had it take a flag to disable that behavior)?

Apps/Examples/Examples/Models/Examples.swift
```diff
@@ -149,6 +153,13 @@ public struct Examples {
 
     ]
 
+    // Examples focused on displaying the user's location.
```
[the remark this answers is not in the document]
evil159: Ctrl+C Ctrl+V 😅

kiryldz (approved):

evil159:
> Do we need separate methods for addImage and addImageResizable? What if we just offered addImage and made it respect cap insets by default (and potentially had it take a flag to disable that behavior)?

Good point. Indeed, there's no immediate need to have two separate methods for regular and resizable images. It was a conscious decision. Adding resizable image support to the existing method might be considered a breaking change, as it would lead to changes in behaviour of the method when a resizable image is supplied.

I think it's a sound idea to add an opt out flag, however I feel that will make the "addImage(..." method even more crowded. 
```swift
func addImage(_ image: UIImage,
                id: String,
                sdf: Bool = false,
                enableResizing: Bool = true,
                stretchX: [ImageStretches],
                stretchY: [ImageStretches],
                content: ImageContent? = nil) throws
```

If I understand your suggestion correctly, one would need to explicitly state that they want image to be resized by setting the flag. In that case I think there's very little difference between that and using a distinct method to add a resizable image.

I wonder if we could change the behaviour of `addImage(...` method to support resizable images by default(without a flag) in the next major version. If so, then adding a flag is definitely the way to, so the flag can be removed later on.

Sources/MapboxMaps/Style/Style.swift
```diff
@@ -721,6 +725,71 @@ public final class Style: StyleProtocol {
...
+    /// [`line-pattern`](https://www.mapbox.com/mapbox-gl-js/style-spec/#paint-line-line-pattern).
+    ///
+    /// - Parameters:
+    ///   - image: Image to add.
+    ///   - id: ID of the image.
+    ///   - sdf: Option to treat whether image is SDF(signed distance field) or not.
+    ///         Setting this to `true` allows template images to be recolored. The
+    ///         default value is `false`.
+    ///   - content: An array of four numbers, with the first two specifying the
+    ///         left, top corner, and the last two specifying the right, bottom
```
OdNairy, line 746: > content: An array of four numbers

Seems ambiguous. I see an `ImageContent` type and not the array of numbers.

Sources/MapboxMaps/Style/Style.swift
```diff
@@ -721,6 +725,71 @@ public final class Style: StyleProtocol {
...
+    ///         symbol's text will be fit inside the content box.
+    ///
+    /// - Throws:
+    ///     An error describing why the operation was unsuccessful.
+    public func addImageResizable(_ image: UIImage, id: String, sdf: Bool = false, content: ImageContent? = nil) throws {
+        guard image.capInsets != .zero else {
+            Log.warning(forMessage: "The supplied image is missing cap insets and might be resized incorrectly.")
+            try addImage(image, id: id, stretchX: [], stretchY: [])
+            return
+        }
```
OdNairy, line 777: This guard statement makes me think that we don't need a separate methods. 
It looks like we can introduce united `addImage` method and highlight in the description that this method supports resizable images.

Sources/MapboxMaps/Style/Style.swift
OdNairy: Just saw our comment:
> I decided to leave the image content as a parameter accessible, if needed, to the user. To make the API more Swifty the parameter type could be `UIEdgeInsets` instead of `ImageContent`, as edge insets should be more familiar to iOS developers.

Switching to `UIEdgeInsets` looks legit for me. It makes much more sense for non-SDK developers. 
However, we have to be sure that it wouldn't prevent developers from passing this value to some other method or variables that accepts only `ImageContent` type.

Sources/MapboxMaps/Style/Style.swift
```diff
@@ -721,6 +725,71 @@ public final class Style: StyleProtocol {
...
+    /// Adds an image to be used in the style.
+    ///
+    /// This API can also be used for
+    /// updating an image. If the image id was already added, it gets replaced
+    /// by the new image.
+    ///
+    /// The image can be used in
+    /// [`icon-image`](https://www.mapbox.com/mapbox-gl-js/style-spec/#layout-symbol-icon-image),
+    /// [`fill-pattern`](https://www.mapbox.com/mapbox-gl-js/style-spec/#paint-fill-fill-pattern), and
+    /// [`line-pattern`](https://www.mapbox.com/mapbox-gl-js/style-spec/#paint-line-line-pattern).
```
OdNairy, line 737: Just to confirm – does Xcode autocomplete/documentation support this format?

OdNairy:
> Adding resizable image support to the existing method might be considered a breaking change, as it would lead to changes in behaviour of the method when a resizable image is supplied.

It's an interesting problem on definition a "breaking changes". As for me, we should not consider that kind of changes as a breaking one. I feel like we had a bug which we just fixed.

Sources/MapboxMaps/Style/Style.swift
macdrevx: I support using UIEdgeInsets. We could name the parameter contentInsets as well, which has precedent in UIKit APIs.

Sources/MapboxMaps/Style/Style.swift
macdrevx: Agreed. This is what I had in mind: https://github.com/mapbox/mapbox-maps-ios/pull/1269#pullrequestreview-941701627

Every UIImage has a non-optional capInsets property. It just defaults to zero for non-resizable images.

macdrevx:
My suggestion:

1. Leave the preexisting addImage API (the one that involves imageStretches) as-is.
2. Add a single new method:

```
public func addImage(
    _ image: UIImage,
    id: String,
    sdf: Bool = false,
    contentInsets: UIEdgeInsets = .zero) throws
```

and document that it respects UIImage.capInsets but not [`UIImage.resizingMode`](https://developer.apple.com/documentation/uikit/uiimage/1624157-resizingmode) (the result is always as if `resizingMode == .stretch`)

---

I'd mentioned possibly adding a flag about whether to respect capInsets — that was targeted towards the new method, but thinking about it more, it seems unnecessary. Developers who don't want resizing behavior should just pass images with `capInsets == .zero`.

Sources/MapboxMaps/Style/Style.swift
```diff
@@ -721,6 +725,71 @@ public final class Style: StyleProtocol {
...
+    ///         Setting this to `true` allows template images to be recolored. The
+    ///         default value is `false`.
+    ///   - content: An array of four numbers, with the first two specifying the
+    ///         left, top corner, and the last two specifying the right, bottom
+    ///         corner. If present, and if the icon uses icon-text-fit, the
+    ///         symbol's text will be fit inside the content box.
+    ///
+    /// - Throws:
+    ///     An error describing why the operation was unsuccessful.
+    public func addImageResizable(_ image: UIImage, id: String, sdf: Bool = false, content: ImageContent? = nil) throws {
```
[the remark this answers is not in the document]
evil159: Good point, for some reason it didn't appear in the list of suggestions when I tried it. Now it works like you said.

Sources/MapboxMaps/Style/Style.swift
evil159: Seems like it does

Screenshot 2022-04-19 at 15 47 08

Sources/MapboxMaps/Style/Style.swift
```diff
@@ -721,6 +725,46 @@ public final class Style: StyleProtocol {
...
+    /// regardless of the value in `UIImage.resizingMode`.
+    ///
+    /// - Parameters:
+    ///   - image: Image to add.
+    ///   - id: ID of the image.
+    ///   - sdf: Option to treat whether image is SDF(signed distance field) or not.
+    ///         Setting this to `true` allows template images to be recolored. The
+    ///         default value is `false`.
+    ///   - contentInsets: The distances the edges of content are inset from the image rectangle.
+    ///         If present, and if the icon uses icon-text-fit, the
```
macdrevx, line 740: Is this specific to symbol layer?

Sources/MapboxMaps/Style/Style.swift
```diff
@@ -721,6 +725,46 @@ public final class Style: StyleProtocol {
...
+        let scale = Float(image.scale)
+        let stretchXFirst = Float(image.capInsets.left) * scale
+        let stretchXSecond = Float(image.size.width - image.capInsets.right) * scale
+        let stretchYFirst = Float(image.capInsets.top) * scale
+        let stretchYSecond = Float(image.size.height - image.capInsets.bottom) * scale
+
+        let contentBoxLeft = Float(contentInsets.left) * scale
+        let contentBoxRight = Float(image.size.width - contentInsets.right) * scale
+        let contentBoxTop = Float(contentInsets.top) * scale
+        let contentBoxBottom = Float(image.size.height - contentInsets.bottom) * scale
```
macdrevx, line 755: curious to learn more about how scale works in GL Native here.

macdrevx (approved):

PR #1278, opened 2022-04-15
Fix localization crash on iOS 11 and 12

### LINKED ISSUE #954: Crash under iOS 12 in mapView.mapboxMap.style.localizeLabels()
## Environment

- Xcode version: 13.2
- iOS version: iOS 12.45.5
- Devices affected: iPhone, probably iPad
- Maps SDK Version: V10

After the map style has loaded I call the localize method to get localized map labels (I’ve loaded the „outdoor“ map style):
`self.mapView.mapboxMap.style.localizeLabels(into: Locale.current)`

This works fine under iOS 15 (probably also other iOS releases) but crashes under iOS 12 on my iPhone 5s (where I checked my App how it works on the tiny screen).

> Fatal error: 'try!' expression unexpectedly raised an error: Swift.EncodingError.invalidValue(nil, Swift.EncodingError.Context(codingPath: [], debugDescription: "Top-level Optional<Value<String>> encoded as null JSON fragment.", underlyingError: nil)): file MapboxMaps/Style+Localization.swift, line 28

## Expected behavior

The Map applies the localization if available, if not available it uses the default (as documented).

According to the debugger output, the crash occurs in "Style+Localization.swift“ at line 28 because of the „try!“ which can not be catched  for the call "convertExpressionForLocalization()“. Further analysis reveals that within "convertExpressionForLocalization()" in line 86 the „symbolLayer.textField“ is nil under iOS 12 (at least for one of the layers), which has the result that the JSONEncoder throws an error. I have no idea why the textField of one of the layers is nil under iOS 12 but not under iOS 15. But by changing the non-catchable "try!“ (line 28) into „try“ or "try?“ so that the App won’t crash with a runtime error anymore, the Map shows fine even under iOS 12 perfectly localized.

I have just made my first steps with the mapBox SDK, so I’m not that familar with all the details. Therefore I'm not really sure if the not-catchable „try!“ has a special purpose here. But it seems that if certain layers might not have all data set under all (older) iOS releases, this „try“ should be better „catchable“ to avoid such crashes.

macdrevx:
* Style.localizeLabels(into:forLayerIds:) crashed on iOS 11 and 12 if
  any of the symbol layers had text-field set to nil.

Fixes #954

## Pull request checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ Mapbox welcomes participation and contributions from everyone.
 
 * Add support for runtime source properties. ([#1267](https://github.com/mapbox/mapbox-maps-ios/pull/1267))
 * Start location services lazily. ([#1262](https://github.com/mapbox/mapbox-maps-ios/pull/1262))
+* Fix localization crash on iOS 11 and 12. ([#1278](https://github.com/mapbox/mapbox-maps-ios/pull/1278))
 
 ## 10.5.0-beta.1 - April 7, 2022
 
--- a/Sources/MapboxMaps/Style/Style+Localization.swift
+++ b/Sources/MapboxMaps/Style/Style+Localization.swift
@@ -83,8 +83,8 @@ extension Style {
         let expressionAbbr = try NSRegularExpression(pattern: "\\[\"get\",\\s*\"abbr\"\\]",
                                                      options: .caseInsensitive)
 
-        if var stringExpression = String(data: try JSONEncoder().encode(symbolLayer.textField), encoding: .utf8),
-           stringExpression != "null" {
+        if case .expression(let textField) = symbolLayer.textField,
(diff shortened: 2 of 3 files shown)


macdrevx:
Added a test that exercises the crash when run with the test host on an iOS 12 device and passes with the code change.

evil159 (approved):
I wonder if we should eliminate force unwrapping(lines 28 and 52) here to prevent these kinds of crashes happening in future. Nice thing about crashing is that they are more prominent to the SDK users, so they are more likely to report them. While thrown errors are more likely to be ignored.

macdrevx:
@evil159 these are situations where we have complete control of the inputs, so if it throws, it's our fault and finding out about it through a crash seems reasonable to me. They don't seem like actionable errors to propagate to the public API.

PR #1281, opened 2022-04-19
Update dependencies

macdrevx:
MapboxCoreMaps: v10.5.0-beta.1 -> v10.5.0-rc.1
MapboxCommon: v21.3.0-beta.2 -> v21.3.0-rc.2
Turf: v2.3.0 -> v2.4.0 (only for development & direct download; SPM & CocoaPods dependency version requirements remain set to any 2.x)

## Features ✨ and improvements 🏁
* Avoid repeated tile loading from network (or repeated tile decompression when the tile is fetched from the cache database) and repeated vector tile data allocation and parsing when loading render tiles referring to the same logical tile.
* Switch to use shader to calculate the 'line-trim-offset' property update.  

## Bug fixes 🐞
* Fix issue where internal hsla() function was converted to an invalid rgba expression
* Fix a bug that 'line-trim-offset' calculation did not property cover 'round' or 'square' line cap in line ends.

### DIFF
--- a/Apps/Apps.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/Apps/Apps.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -15,8 +15,8 @@
         "repositoryURL": "https://github.com/mapbox/mapbox-common-ios.git",
         "state": {
           "branch": null,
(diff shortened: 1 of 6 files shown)


macdrevx:
Device tests job failed due to CircleCI timeout, but the actual tests passed on firebase. cc: @OdNairy FYI.

evil159 (approved):

PR #1301, opened 2022-04-25
Call reduceMemoryUse on app enter background

maios:
Call MapboxMap.reduceMemoryUse when application enters the background.
## Pull request checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ Mapbox welcomes participation and contributions from everyone.
 
 * Expose API to check whether an image exists in `Style`. ([#1297](https://github.com/mapbox/mapbox-maps-ios/pull/1297))
 * Invoke animator completion handlers added after completion or cancellation. ([#1305])(https://github.com/mapbox/mapbox-maps-ios/pull/1305))
+* Call `MapboxMap.reduceMemoryUse` when application goes to background. ([#1301](https://github.com/mapbox/mapbox-maps-ios/pull/1301))
 
 ## 10.5.0-rc.1 - April 20, 2022
 
--- a/Sources/MapboxMaps/Foundation/MapView.swift
+++ b/Sources/MapboxMaps/Foundation/MapView.swift
@@ -465,15 +465,6 @@ open class MapView: UIView {
         }
     }
 
-    private func unsubscribeFromLifecycleNotifications() {
-        if #available(iOS 13.0, *) {
-            notificationCenter.removeObserver(self, name: UIScene.willEnterForegroundNotification, object: nil)
-            notificationCenter.removeObserver(self, name: UIScene.didEnterBackgroundNotification, object: nil)
-        }
-        notificationCenter.removeObserver(self, name: UIApplication.willEnterForegroundNotification, object: nil)
-        notificationCenter.removeObserver(self, name: UIApplication.didEnterBackgroundNotification, object: nil)
-    }
-
     @objc private func appWillEnterForeground() {
         displayLink?.isPaused = false
     }
--- a/Sources/MapboxMaps/Foundation/MapView.swift
+++ b/Sources/MapboxMaps/Foundation/MapView.swift
@@ -322,6 +322,7 @@ open class MapView: UIView {
             mapInitOptions: resolvedMapInitOptions,
             mapboxObservableProvider: dependencyProvider.mapboxObservableProvider)
 
+        subscribeToLifecycleNotifications()
         notificationCenter.addObserver(self,
                                        selector: #selector(didReceiveMemoryWarning),
                                        name: UIApplication.didReceiveMemoryWarningNotification,
@@ -584,8 +585,6 @@ open class MapView: UIView {
     open override func didMoveToWindow() {
         super.didMoveToWindow()
 
-        unsubscribeFromLifecycleNotifications()
-
         displayLink?.invalidate()
         displayLink = nil
 
@@ -595,7 +594,6 @@ open class MapView: UIView {
             return
         }
 
-        subscribeToLifecycleNotifications()
 
         displayLink = dependencyProvider.makeDisplayLink(
             window: window,
--- a/Sources/MapboxMaps/Foundation/MapView.swift
+++ b/Sources/MapboxMaps/Foundation/MapView.swift
@@ -479,6 +479,7 @@ open class MapView: UIView {
 
     @objc private func appDidEnterBackground() {
         displayLink?.isPaused = true
+        mapboxMap.reduceMemoryUse()
     }
 
     @available(iOS 13.0, *)
@@ -494,6 +495,7 @@ open class MapView: UIView {
         guard notification.object as? UIScene == window?.parentScene else { return }
 
         displayLink?.isPaused = true
+        mapboxMap.reduceMemoryUse()
     }
 
     @objc private func didReceiveMemoryWarning() {
--- a/Sources/MapboxMaps/Foundation/MapView.swift
+++ b/Sources/MapboxMaps/Foundation/MapView.swift
@@ -594,7 +594,6 @@ open class MapView: UIView {
             return
         }
 
-
         displayLink = dependencyProvider.makeDisplayLink(
             window: window,
             target: ForwardingDisplayLinkTarget { [weak self] in
--- a/Sources/MapboxMaps/Foundation/MapView.swift
+++ b/Sources/MapboxMaps/Foundation/MapView.swift
@@ -595,6 +595,8 @@ open class MapView: UIView {
             return
         }
 
+        subscribeToLifecycleNotifications()
+
         displayLink = dependencyProvider.makeDisplayLink(
             window: window,
             target: ForwardingDisplayLinkTarget { [weak self] in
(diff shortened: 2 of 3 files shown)


evil159 (approved):

Sources/MapboxMaps/Foundation/MapView.swift
```diff
@@ -494,6 +495,7 @@ open class MapView: UIView {
         guard notification.object as? UIScene == window?.parentScene else { return }
 
         displayLink?.isPaused = true
+        mapboxMap.reduceMemoryUse()
     }
```
macdrevx, line 499: I think we only subscribe to these notifications when we have a non-nil display link. Do we need to change that in light of this addition?

Sources/MapboxMaps/Foundation/MapView.swift
```diff
@@ -593,6 +595,8 @@ open class MapView: UIView {
             return
         }
 
+        subscribeToLifecycleNotifications()
+
```
macdrevx, line 599: Now that we're using these lifecycle notifications in a more general-purpose way, would it be simpler to just subscribe at the end of commonInit and stay subscribed all the time?

macdrevx: Being subscribed all the time doesn't seem like it would be significant in terms of performance since the handlers do very little work and are called infrequently. Reducing memory use would be valuable regardless of whether the MapView belongs to a window. Besides performance, are there any other things we should be concerned about?

maios: good point, I will remove the unsubscribe line and make MapView subscribe to these notifications always

Sources/MapboxMaps/Foundation/MapView.swift
```diff
@@ -582,8 +585,6 @@ open class MapView: UIView {
     open override func didMoveToWindow() {
         super.didMoveToWindow()
 
-        unsubscribeFromLifecycleNotifications()
```
macdrevx, line 585: Was this the only place this method was called? If so, let's remove the method.

macdrevx (commented):
1 additional comment

Sources/MapboxMaps/Foundation/MapView.swift
maios: done, I also removed test that related to this

maios: done, I also removed test that related to this

Tests/MapboxMapsTests/Foundation/MapViewTests.swift
```diff
@@ -261,91 +261,6 @@ final class MapViewTests: XCTestCase {
         XCTAssertEqual(runner.updateStub.invocations.count, 1)
     }
 
-    func testAppLifecycleNotificationSubscribedWhenDidMoveToNewWindow() {
```
maios, line 264: move this to another test class so I can manage Mock Bundle with mock scene support better.

maios, line 264: this test case is no longer valid

Sources/MapboxMaps/Foundation/MapView.swift
macdrevx: Looks like the change to remove the method was not pushed.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ Mapbox welcomes participation and contributions from everyone.
 ## main
 
 * Expose API to check whether an image exists in `Style`. ([#1297](https://github.com/mapbox/mapbox-maps-ios/pull/1297))
+* Call `MapboxMap.reduceMemoryUse` when application goes to background. ([#1301](https://github.com/mapbox/mapbox-maps-ios/pull/1301))
```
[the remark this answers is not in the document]
macdrevx: Doing it each time seems like a good starting point to me. If we run into trouble with that approach, we could invest in additional complexity to add some delay.

Tests/MapboxMapsTests/Foundation/MapViewTests.swift
```diff
@@ -410,3 +301,97 @@ final class MapViewTests: XCTestCase {
...
+            urlOpener: attributionURLOpener)
+
+        mapView.didMoveToWindow()
+
+        notificationCenter.post(name: UIScene.willEnterForegroundNotification, object: window.parentScene)
+
+        XCTAssertEqual(displayLink.$isPaused.setStub.invocations.map(\.parameters), [false])
+    }
+
+    @available(iOS 13, *)
```
macdrevx, line 381: This doesn't work to exclude a unit test. You need a runtime check as shown in some of the deleted tests.

Tests/MapboxMapsTests/Foundation/MapViewTests.swift
```diff
@@ -410,3 +301,97 @@ final class MapViewTests: XCTestCase {
...
+        displayLink = nil
+        cameraAnimatorsRunnerEnablable = nil
+        bundle = nil
+        notificationCenter = nil
+        super.tearDown()
+    }
+
+    @available(iOS 13, *)
+    func testDisplayLinkResumedWhenSceneMovingToForeground() throws {
+        bundle.infoDictionaryStub.defaultReturnValue = ["UIApplicationSceneManifest": []]
```
macdrevx, line 366: Seems like this is already happening in setUp

Tests/MapboxMapsTests/Foundation/MapViewTests.swift
```diff
@@ -410,3 +301,97 @@ final class MapViewTests: XCTestCase {
...
+        mapView = MapView(
+            frame: CGRect(origin: .zero, size: CGSize(width: 100, height: 100)),
+            mapInitOptions: MapInitOptions(),
+            dependencyProvider: dependencyProvider,
+            orientationProvider: orientationProvider,
+            urlOpener: attributionURLOpener)
+
+        mapView.didMoveToWindow()
```
macdrevx, line 374: This also seems to already be in setUp

Tests/MapboxMapsTests/Foundation/MapViewTests.swift
```diff
@@ -410,3 +301,97 @@ final class MapViewTests: XCTestCase {
...
+    func testDisplayLinkPausedWhenSceneMovingToBackground() throws {
+        bundle.infoDictionaryStub.defaultReturnValue = ["UIApplicationSceneManifest": []]
+
+        mapView = MapView(
+            frame: CGRect(origin: .zero, size: CGSize(width: 100, height: 100)),
+            mapInitOptions: MapInitOptions(),
+            dependencyProvider: dependencyProvider,
+            orientationProvider: orientationProvider,
+            urlOpener: attributionURLOpener)
+        mapView.didMoveToWindow()
```
macdrevx, line 391: Same here

macdrevx (commented):
Tests need some updates

Tests/MapboxMapsTests/Foundation/MapViewTests.swift
```diff
@@ -261,110 +261,112 @@ final class MapViewTests: XCTestCase {
...
     }
 
     func testDisplayLinkResumedWhenSceneMovingToForeground() throws {
         guard #available(iOS 13.0, *) else {
             throw XCTSkip("Test requires iOS 13 or higher.")
         }
-        bundle.infoDictionaryStub.defaultReturnValue = ["UIApplicationSceneManifest": []]
-        mapView.didMoveToWindow()
 
+        mapView.didMoveToWindow()
```
macdrevx, line 369: Does adding the view to the window in setUp call this for you?

macdrevx (commented):
Another question

Tests/MapboxMapsTests/Foundation/MapViewTests.swift
maios: yes it should, it is better to have this call in setup.

Tests/MapboxMapsTests/Foundation/MapViewTests.swift
```diff
@@ -261,109 +261,112 @@ final class MapViewTests: XCTestCase {
...
-    }
-
-    func testDisplayLinkPausedWhenAppMovingToBackground() {
-        notificationCenter.post(name: UIApplication.didEnterBackgroundNotification, object: nil)
+        metalView = try XCTUnwrap(dependencyProvider.makeMetalViewStub.invocations.first?.returnValue)
+        // reset is required here to ignore the setNeedsDisplay() invocation during initialization
+        metalView.setNeedsDisplayStub.reset()
 
-        XCTAssertEqual(displayLink.$isPaused.setStub.invocations.map(\.parameters), [true])
+        mapView.didMoveToWindow()
```
macdrevx, line 348:
```suggestion
```

What I was trying to say is that you don't need this line at all. `window.addSubview(mapView)` calls this method for you.

maios: that's true, good catch. I have removed it :-D

macdrevx (approved):

OdNairy (approved):

PR #1306, opened 2022-04-29
Feature/mock style manager

maios:
Update unit tests for Style with better chance to mock StyleManager.
NOTE: Unit tests for style add image sources will be done separately

## Pull request checklist:


Tests/MapboxMapsTests/Foundation/Mocks/MockStyleManager.swift
```diff
@@ -0,0 +1,331 @@
...
+        fatalError()
+    }
+
+    static func getStyleLayerPropertyDefaultValue(
+        forLayerType layerType: String,
+        property: String)
+    -> MapboxCoreMaps.StylePropertyValue {
+
+        fatalError()
+    }
```
OdNairy, line 72: Would we like to keep this implementation failable or we are going to fix it later?

OdNairy (approved):

Tests/MapboxMapsTests/Foundation/Mocks/MockStyleManager.swift
maios: we are going to fix it later in this PR :-D

Tests/MapboxMapsTests/Foundation/StyleTests.swift
```diff
@@ -38,4 +40,394 @@ final class StyleTests: XCTestCase {
         style.styleManager.setStyleProjectionForProperties(["name": "globe"])
         XCTAssertEqual(style.projection.name, .globe)
     }
+
+    func testStyleIsLoaded() {
+        let mockStyleManager = MockStyleManager()
+        let sut = Style(with: mockStyleManager)
+
+        var isStyleLoaded: Bool = .random()
```
macdrevx, line 48:
```suggestion
        let isStyleLoaded: Bool = .random()
```

macdrevx:
Seeing some warnings like "/Users/andrewhershberger/Developer/mapbox-maps-ios/Tests/MapboxMapsTests/Foundation/Extensions/Core/DownloadStatusTests.swift:3:8: 'MapboxCommon_Private' inconsistently imported as implementation-only"

Tests/MapboxMapsTests/Foundation/Mocks/MockStyleManager.swift
```diff
@@ -0,0 +1,343 @@
...
+
+    @Stubbed var stubStyleURI: String!
+    func getStyleURI() -> String {
+        stubStyleURI ?? ""
+    }
+
+    func setStyleURIForUri(_ uri: String) {
+        stubStyleURI = uri
+    }
```
macdrevx, line 18: Interesting approach to treat the getter and setter like a property. A couple thoughts…

* Rather than using nil-coalescing, I would just make `stubStyleURI` non-optional an initialize it to `""`
* This new approach has some merit, but it is also inconsistent with our other mocks. We should think carefully about what introducing this new technique would mean for the maintainability of the project overall. Would we want to update other mocks to use the same technique? If not, what would the impact be over time for other developers who come into the project?

Tests/MapboxMapsTests/Foundation/Mocks/MockStyleManager.swift
```diff
@@ -0,0 +1,343 @@
...
+
+    @Stubbed var stubStyleJSON: String!
```
macdrevx, line 20:
```suggestion
    @Stubbed var stubStyleJSON: String!
```

In other mocks, we follow a consistent naming convention for stub properties: `styleJSONStub` would be more consistent.

Tests/MapboxMapsTests/Foundation/Mocks/MockStyleManager.swift
```diff
@@ -0,0 +1,343 @@
...
+    // MARK: Style Layers
+
+    @Stubbed var stubStyleLayers: [MapboxCoreMaps.StyleObjectInfo] = .random(withLength: 3) {
+        MapboxCoreMaps.StyleObjectInfo(
+            id: .randomAlphanumeric(withLength: 12),
+            type: MapboxMaps.LayerType.random().rawValue)
+    }
+    func styleLayerExists(forLayerId layerId: String) -> Bool {
+        stubStyleLayers.contains(where: { $0.id == layerId })
+    }
```
macdrevx, line 52: This doesn't record parameters like most stubs. A standalone Stub would be better here.

Also, the default return value seems very specific. I would recommend moving it into the test that needs it.

Tests/MapboxMapsTests/Foundation/Mocks/MockStyleManager.swift
```diff
@@ -0,0 +1,343 @@
...
+        forLayerId layerId: String,
+        property: String
+    ) -> MapboxCoreMaps.StylePropertyValue {
+
+        if let stub = mockery.stub(of: MockStyleManager.getStyleLayerProperty(forLayerId:property:)) {
+            return stub.call(with: self)(layerId, property)
+        } else {
+            return MapboxCoreMaps.StylePropertyValue(value: "foo", kind: .undefined)
+        }
+    }
```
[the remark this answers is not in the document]
maios: Which part do you mean? The mockery part or the part where we use default mock value?

maios:
> Seeing some warnings like "/Users/andrewhershberger/Developer/mapbox-maps-ios/Tests/MapboxMapsTests/Foundation/Extensions/Core/DownloadStatusTests.swift:3:8: 'MapboxCommon_Private' inconsistently imported as implementation-only"

yeah interesting to me, that we are doing `@_implementationOnly import MapboxCoreMaps_Private` but then when importing `MapboxCommon_Private` we are not using `implementationOnly`, why is there a difference when importing these 2 modules in unit tests target?

Tests/MapboxMapsTests/Foundation/Mocks/MockStyleManager.swift
```diff
@@ -0,0 +1,462 @@
...
+
+        addStyleCustomLayerStub.call(
+            with: AddStyleCustomLayerParameters(layerId: layerId, layerHost: layerHost, layerPosition: layerPosition)
+        )
+    }
+
+    let addPersistentStyleLayerStub = Stub<AddStyleLayerParameters, Expected<NSNull, NSString>>(
+        defaultReturnValue: .init(value: NSNull())
+    )
+    @objc func addPersistentStyleLayer(
```
macdrevx, line 177: is the `@objc` attribute necessary?

maios: good catch :-D it is not needed, I added it to try on some swizzle and forgot to remove it

macdrevx (approved):

PR #1325, opened 2022-05-10
Add integration CI jobs for Turf

macdrevx:
Adds two CI jobs:

* The PR workflow now includes a job that runs the unit tests with the min supported Turf version.
* A new workflow is added to enable triggering unit tests against a specified Turf revision.
* Updates depsvalidator for improved compatibility with future Swift Package manifest definitions.

## Pull request checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,6 +1,11 @@
 ---
 version: 2.1
 
+parameters:
+  turf-revision:
+    type: string
+    default: ""
+
 orbs:
   # https://circleci.com/orbs/registry/orb/circleci/slack
   slack: circleci/slack@4.5
@@ -75,6 +80,7 @@ workflows:
           xcode: "13.0.0"
 
   optional-tests:
+    unless: << pipeline.parameters.turf-revision >>
     jobs:
       - run-tests?:
           type: approval
@@ -115,6 +121,7 @@ workflows:
             - trigger-compatibility-verification?
 
   steve:
+    unless: << pipeline.parameters.turf-revision >>
     jobs:
       - swiftlint
       - depsvalidator:
@@ -149,6 +156,11 @@ workflows:
       - unit-test-sdk:
           name: "Run Unit tests"
           xcode: "13.0.0"
+
+      - unit-test-sdk:
+          name: "Run Unit Tests with min Turf"
+          xcode: "13.0.0"
+          turf-requirement: --version 2.0.0
 
       # Build all tests with host application
       # Run on main by default
@@ -236,6 +248,14 @@ workflows:
             - update-examples
             - validate-integrations
 
+  turf-integration:
+    when: << pipeline.parameters.turf-revision >>
+    jobs:
+      - unit-test-sdk:
+          name: "Run Unit Tests"
+          xcode: "13.0.0"
+          turf-requirement: --revision << pipeline.parameters.turf-revision >>
+
 # ==============================================================================
 
 jobs:
@@ -263,6 +283,9 @@ jobs:
       app-name:
         type: string
         default: "MapboxTestHost"
+      turf-requirement:
+        type: string
+        default: ""
     macos:
       xcode: << parameters.xcode >>
     environment:
@@ -435,6 +458,12 @@ jobs:
     steps:
       - checkout
       - configure-environment
+      - when:
+          condition: << parameters.turf-requirement >>
+          steps:
+            - run:
+                name: Resolve Turf
+                command: swift package resolve && swift package resolve << parameters.turf-requirement >> turf-swift
       # Building and testing are split into 2, with the aim that we'll be able to reuse
       # the build product and test on multiple simulators
       - run:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -251,7 +251,7 @@ workflows:
   turf-integration:
     when: << pipeline.parameters.turf-revision >>
     jobs:
-      - unit-test-sdk:
+      - turf-integration:
           name: "Run Unit Tests"
           xcode: "13.0.0"
           turf-requirement: --revision << pipeline.parameters.turf-revision >>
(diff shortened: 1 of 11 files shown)


macdrevx:
Turf's Bitrise workflow will need to be updated to point to the main branch as soon as this is merged.

macdrevx:
Looks like there's a bug in my approach to resolving the turf version. While `swift package resolve …` works, the subsequent `xcodebuild` reverts the changes. There's an xcodebuild flag `-onlyUsePackageVersionsFromResolvedFile` that should prevent this, but [it seems to be broken](https://forums.swift.org/t/error-during-package-resolution-on-ci-machine-whenever-updating-package-dependencies/56472/2). The bug should be fixed in Swift 5.7, but for now, I'm going to implement this in a different way.

I plan to write a script that modifies Package.swift to use the exact turf commit. While I'm at it, I'll update it to use the 5.5 package description format since we're requiring Xcode 13+ now.

.circleci/config.yml
```diff
@@ -75,6 +80,7 @@ workflows:
           xcode: "13.0.0"
 
   optional-tests:
+    unless: << pipeline.parameters.turf-revision >>
```
OdNairy, line 83: Does it work for string types? Does it mean we will skip basic workflows if anything is set in `turf-revision`?

.circleci/config.yml
```diff
@@ -435,6 +458,12 @@ jobs:
     steps:
       - checkout
       - configure-environment
+      - when:
+          condition: << parameters.turf-requirement >>
+          steps:
+            - run:
+                name: Resolve Turf
+                command: swift package resolve && swift package resolve << parameters.turf-requirement >> turf-swift
```
OdNairy, line 466: Nit:
```suggestion
                command: |
                    swift package resolve
                    swift package resolve << parameters.turf-requirement >> turf-swift
```

.circleci/config.yml
```diff
@@ -844,6 +873,42 @@ jobs:
             timeout 1800 bash -c 'while [[ "$(curl -n -s -o /dev/null -w ''%{http_code}'' https://api.mapbox.com/downloads/v2/mobile-maps-ios/releases/ios/$VERSION/MapboxMaps.zip)" != "200" ]]; do sleep 5; done' || false
           no_output_timeout: 30m
 
+  turf-integration:
+    <<: *base-job
```
OdNairy, line 877: Can we drop the `base-job` injection here? I'd like to drop it someday from anywhere in CircleCI configuration if favor of shared executors.

Package.swift
```diff
@@ -13,26 +13,34 @@ let package = Package(
...
-        .package(name: "MapboxCoreMaps", url: "https://github.com/mapbox/mapbox-core-maps-ios.git", .exact("10.5.1")),
-        .package(name: "MapboxCommon", url: "https://github.com/mapbox/mapbox-common-ios.git", .exact("21.3.0")),
-        .package(name: "MapboxMobileEvents", url: "https://github.com/mapbox/mapbox-events-ios.git", .exact("1.0.7")),
-        .package(name: "Turf", url: "https://github.com/mapbox/turf-swift.git", from: "2.0.0"),
-        .package(name: "CocoaImageHashing", url: "https://github.com/ameingast/cocoaimagehashing", .exact("1.9.0"))
+        .package(url: "https://github.com/mapbox/mapbox-core-maps-ios.git", revision: "v10.5.1"),
+        .package(url: "https://github.com/mapbox/mapbox-common-ios.git", revision: "v21.3.0"),
+        .package(url: "https://github.com/mapbox/mapbox-events-ios.git", revision: "v1.0.7"),
+        .package(url: "https://github.com/mapbox/turf-swift.git", from: "2.0.0"),
+        .package(url: "https://github.com/ameingast/cocoaimagehashing", from: "1.9.0"),
```
OdNairy, line 20: Is it a required change for swift-tools 5.5? 
The previous version looks a bit better for me as we name the product just one in dependencies rather than several times in different places. I suppose we don't have a multi-product packages?

.circleci/config.yml
```diff
@@ -844,6 +873,42 @@ jobs:
...
+          command: swift package resolve && swift package resolve << parameters.turf-requirement >> turf-swift
+      # Building and testing are split into 2, with the aim that we'll be able to reuse
+      # the build product and test on multiple simulators
+      - run:
+          name: Building MapboxMaps for simulator
+          command: make build-sdk-for-testing-simulator
+          no_output_timeout: 5m
+      - run:
+          name: Testing MapboxMaps with simulator
+          command: make test-sdk-without-building-simulator
```
OdNairy, line 895: So we are not reusing because it would also trigger a lot of extra jobs, right?

OdNairy: We also have to call additional GitHub status checks.

.circleci/config.yml
macdrevx: Correct

macdrevx: When this is set, I only want the turf-integration workflow to run

Package.swift
macdrevx: The old methods are now deprecated and will be removed in a future version.

.circleci/config.yml
macdrevx: Yep

Package.swift
```diff
@@ -12,27 +12,37 @@ let package = Package(
...
+    // replace `.package(name:url:_: .exact())` with `.package(url:exact:)`
     dependencies: [
-        .package(name: "MapboxCoreMaps", url: "https://github.com/mapbox/mapbox-core-maps-ios.git", .exact("10.5.1")),
-        .package(name: "MapboxCommon", url: "https://github.com/mapbox/mapbox-common-ios.git", .exact("21.3.0")),
-        .package(name: "MapboxMobileEvents", url: "https://github.com/mapbox/mapbox-events-ios.git", .exact("1.0.7")),
-        .package(name: "Turf", url: "https://github.com/mapbox/turf-swift.git", from: "2.0.0"),
-        .package(name: "CocoaImageHashing", url: "https://github.com/ameingast/cocoaimagehashing", .exact("1.9.0"))
+        .package(name: "mapbox-core-maps-ios", url: "https://github.com/mapbox/mapbox-core-maps-ios.git", .exact("10.5.1")),
+        .package(name: "mapbox-common-ios", url: "https://github.com/mapbox/mapbox-common-ios.git", .exact("21.3.0")),
+        .package(name: "mapbox-events-ios", url: "https://github.com/mapbox/mapbox-events-ios.git", .exact("1.0.7")),
```
macdrevx, line 20: I reverted back to the deprecated methods here since the replacement for `.exact` isn't available until Swift 5.6. Initially I used revision and the tag name instead, but that doesn't play nicely with depsvalidator.

Package.swift
```diff
@@ -12,27 +12,37 @@ let package = Package(
...
+        .package(url: "https://github.com/mapbox/turf-swift.git", from: "2.0.0"),
+        .package(url: "https://github.com/ameingast/cocoaimagehashing", from: "1.9.0"),
```
macdrevx, line 22: cocoaimagehashing is just for testing, so no need for exact pinning.

scripts/depsvalidator/Package.swift
```diff
@@ -1,9 +1,9 @@
-// swift-tools-version:5.4
+// swift-tools-version:5.5
 import PackageDescription
 
 let package = Package(
     name: "DepsValidator",
-    platforms: [.macOS(.v10_13)],
+    platforms: [.macOS(.v10_15)],
```
macdrevx, line 6: Upgraded this because I was trying out a dependency that required it. I removed the dependency, but left this change since it seems unnecessary to be so aggressive with supporting old macOS versions.

scripts/depsvalidator/Sources/DepsValidatorLibrary/Models/Process+Shell.swift
```diff
@@ -8,6 +8,7 @@ extension Process {
         if let environment = environment {
             process.environment = process.environment ?? [:]
             process.environment?.merge(environment, uniquingKeysWith: { (_, rhs) in rhs })
+            process.environment?.merge(ProcessInfo.processInfo.environment, uniquingKeysWith: { (_, rhs) in rhs })
```
macdrevx, line 11: Not sure why this wasn't necessary before, but without this, custom rules were unable to inherit the PATH

OdNairy: Should we run this merge before `process.environment?.merge(environment`? Wouldn't the ProcessInfo environment override variables?

OdNairy (approved):

scripts/depsvalidator/Sources/DepsValidatorLibrary/Models/Process+Shell.swift
macdrevx: Yes, good catch

PR #1328, opened 2022-05-11
Example for navigation simulation

maios:
MAPSIOS-167 Create a simulation for a navigation with vanishing routes effect.

https://user-images.githubusercontent.com/30261043/173351246-1baaf8ef-ab60-48de-8510-d5c45aeb94b9.mp4

Please note that there is something in progress calculation of Turf so it seems that the progress goes ahead of the movement, this also happens on Android. We will be looking into this separately.
## Pull request checklist:


Apps/Examples/Examples/All Examples/NavigationSimulatorExample.swift
```diff
@@ -0,0 +1,274 @@
...
+        CLLocationCoordinate2D(latitude: 45.49893, longitude: -122.70774)
+    ])
+}
+
+extension NavigationSimulatorExample: LocationConsumer {
+
+    func locationUpdate(newLocation: Location) {
+        let progress = navigationSimulator.distanceTravelled / navigationSimulator.routeLength
+        routeLine.lineTrimOffset = .constant([0, progress])
+        casingLine.lineTrimOffset = .constant([0, progress])
```
pengdev, line 272: I think the issue is that the routeLine and casingLine in iOS is a struct and once the property is updated, you need to set the line back to the style again.

One option is to use the low level style API, to set the lineTrimOffset to the style directly, with the desired layer id.

Apps/Examples/Examples/All Examples/NavigationSimulator.swift
```diff
@@ -0,0 +1,98 @@
...
+
+    func startUpdatingLocation() {
+        let location = CLLocation(
+            coordinate: currentLocation,
+            altitude: 0,
+            horizontalAccuracy: kCLLocationAccuracyBestForNavigation,
+            verticalAccuracy: kCLLocationAccuracyBestForNavigation,
+            // Turf calculates bearing in decimal degrees within -180 to 180,
+            // while Apple's course requires value in decimal degrees from 0 - 359.9
+            course: direction < 0 ? 360 + direction : direction,
```
OdNairy, line 84: I do not quite understand why we don't just make `direction + 180` here?

OdNairy (approved):

Apps/Examples/Examples/All Examples/NavigationSimulatorExample.swift
```diff
@@ -0,0 +1,211 @@
...
+        mapView = MapView(frame: view.bounds)
+        configureMap()
+
+        view.addSubview(mapView)
+        mapView.translatesAutoresizingMaskIntoConstraints = false
+        NSLayoutConstraint.activate([
+            mapView.topAnchor.constraint(equalTo: view.safeAreaLayoutGuide.topAnchor),
+            mapView.leftAnchor.constraint(equalTo: view.safeAreaLayoutGuide.leftAnchor),
+            mapView.bottomAnchor.constraint(equalTo: view.safeAreaLayoutGuide.bottomAnchor),
+            mapView.rightAnchor.constraint(equalTo: view.safeAreaLayoutGuide.rightAnchor),
```
evil159, line 34: I wonder if we should use `leading/trailing` instead `left/right` nowadays 🤔 
```suggestion
            mapView.leadingAnchor.constraint(equalTo: view.safeAreaLayoutGuide.leadingAnchor),
            mapView.bottomAnchor.constraint(equalTo: view.safeAreaLayoutGuide.bottomAnchor),
            mapView.trailingAnchor.constraint(equalTo: view.safeAreaLayoutGuide.trailingAnchor),
```


Though e.g. `OrnamentsManager` doesn't seem to support right-to-left locals automatically(ornament view are positioned with absolute constraints).

Apps/Examples/Examples/All Examples/NavigationSimulatorExample.swift
```diff
@@ -0,0 +1,211 @@
...
+            Exp(.interpolate) {
+                Exp(.exponential) {
+                    1.5
+                }
+                Exp(.zoom)
+                10.0
+                Exp(.product) {
+                    7.0
+                    1.0
+                }
```
evil159, line 143: I'm not quite sure if I understand this correctly, but shouldn't it translate to something like `7.0 * 1.0` - given that `.product` - "Returns the product of the inputs"?

Apps/Examples/Examples/Models/Examples.swift
```diff
@@ -181,6 +181,9 @@ struct Examples {
         Example(title: "Add a custom location provider",
                 description: "Display the location puck at a custom location.",
                 type: CustomLocationProviderExample.self),
+        Example(title: "Simulate navigation",
+                description: "Simulate a navigation route with pre-defined route (from LA to San Francisco) with location puck",
```
evil159, line 185:
```suggestion
                description: "Simulate a driving trip from LA to San Francisco along a pre-defined route",
```

Apps/Examples/Examples/All Examples/NavigationSimulator.swift
maios: both scales have 0 indicating North, so just `direction + 180` won't work here

Apps/Examples/Examples/All Examples/NavigationSimulatorExample.swift
maios: yes, I think explicitly putting one here is for the sake of readiblity in its string representation :-D we could instead do `Exp(.literal) { value }` but I think we should keep this to make it consistent with Android.

Apps/Examples/Examples/All Examples/NavigationSimulatorExample.swift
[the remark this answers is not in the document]
maios: interesting, we should actually constraint or bind the map to the safe area and not the whole view controller, since it is configured for its contents to be extended under the navigation bar -> we can't see the top of the map as Andrew pointed out here https://github.com/mapbox/mapbox-maps-ios/pull/1225#discussion_r838886458
I think the way navigation bar looks like that when view is constrained to navigation bar is because of the way we configure `UINavigationBarAppearance` for iOS 13+

this snippet fixed the problem though, wdyt?
```
        if #available(iOS 13.0, *) {
            let navigationBarAppearance = UINavigationBarAppearance()
            navigationBarAppearance.configureWithOpaqueBackground()

            appearance.scrollEdgeAppearance = navigationBarAppearance
            appearance.compactAppearance = navigationBarAppearance
            appearance.standardAppearance = navigationBarAppearance
        }
```

Apps/Examples/Examples/All Examples/NavigationSimulatorExample.swift
```diff
@@ -0,0 +1,202 @@
...
+        mapView.translatesAutoresizingMaskIntoConstraints = false
+        NSLayoutConstraint.activate([
+            mapView.topAnchor.constraint(equalTo: view.safeAreaLayoutGuide.topAnchor),
+            mapView.leftAnchor.constraint(equalTo: view.safeAreaLayoutGuide.leftAnchor),
+            mapView.bottomAnchor.constraint(equalTo: view.safeAreaLayoutGuide.bottomAnchor),
+            mapView.rightAnchor.constraint(equalTo: view.safeAreaLayoutGuide.rightAnchor),
+        ])
+
+        // The below line is used for internal testing purposes only.
+        finish()
```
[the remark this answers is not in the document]
maios: I have seen `finish()` called in multiple places i.e viewDidLoad, viewDidAppear and in `mapLoaded` event. Correct me if I'm wrong but it seems to be that the purpose of this is to send a notification that is observed by our tests, so it should be fine here.

Apps/Examples/Examples/All Examples/NavigationSimulatorExample.swift
evil159: I think it makes sense for Arabic, Hebrew, Persian, Urdu and number of other RTL languages

Apps/Examples/Examples/All Examples/NavigationSimulatorExample.swift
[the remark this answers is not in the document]
OdNairy: Not quite sure why do they do it, but let's keep it in sync 😄

Apps/Examples/Examples/All Examples/NavigationSimulatorExample.swift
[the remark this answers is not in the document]
evil159: I think extending the map under the bars and making it fill the whole screen makes it looks like it belongs on iOS(the map is visible under the translucent bars). I checked how apple handles this in their Maps app, they use the globe view on lower zoom levels that eliminates this issue. Google Maps seems to both use opaque bars **and** have a part of the map behind them(obstructing a portion of the map behind them, of course).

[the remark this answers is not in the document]
maios: for the sake of example I think it makes more sense to display the full map without any part that is hidden, since we don't really have anything to show/hide navigation bar so it looks weird if a part of the map is not visible.

| Full | Safe-area |
| --- | --- |
| Simulator Screen Shot - iPhone 11 - 2022-06-16 at 12 10 38 | Simulator Screen Shot - iPhone 11 - 2022-06-16 at 12 10 57 |

Apps/Examples/Examples/All Examples/NavigationSimulatorExample.swift
maios: I too think the syntax of iOS `Expression` could be improved. At the moment it can take up too many lines of code and provide little readability :-D would be nice to have a team discussion on this and propose new or addition way for Expression.

PR #1331, opened 2022-05-12
Metal view explicit drawing v2

evil159:
This is a [second](https://github.com/mapbox/mapbox-maps-ios/pull/1157) attempt at enabling the "explicit drawing" mode for our internal `MTKView`. 
Previously it was [reverted](https://github.com/mapbox/mapbox-maps-ios/pull/1216) due to the crashes it caused.

The reason for crashing(metal view having zero size) has already been addressed [here](https://github.com/mapbox/mapbox-maps-ios-internal/issues/998). However, the metal view still has zero size before the first layout pass that resizes it to the minimum size.

This PR makes sure that the metal view is initialised with minimal width and height.

## Pull request checklist:

### DIFF
--- a/Sources/MapboxMaps/Foundation/MapView.swift
+++ b/Sources/MapboxMaps/Foundation/MapView.swift
@@ -626,6 +626,10 @@ open class MapView: UIView {
 
 extension MapView: DelegatingMapClientDelegate {
     internal func scheduleRepaint() {
+        guard let metalView = metalView, !metalView.bounds.isEmpty else {
+            return
+        }
+
         needsDisplayRefresh = true
     }
 
--- a/Sources/MapboxMaps/Foundation/MapView.swift
+++ b/Sources/MapboxMaps/Foundation/MapView.swift
@@ -638,12 +638,8 @@ extension MapView: DelegatingMapClientDelegate {
     }
 
     internal func getMetalView(for metalDevice: MTLDevice?) -> MTKView? {
-        let minSize = CGSize(width: 1, height: 1)
-        let finalSize = CGSize(
-            width: max(minSize.width, bounds.width),
-            height: max(minSize.height, bounds.height))
-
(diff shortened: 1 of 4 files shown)


Sources/MapboxMaps/Foundation/MapView.swift
```diff
@@ -627,6 +626,10 @@ open class MapView: UIView {
 
 extension MapView: DelegatingMapClientDelegate {
     internal func scheduleRepaint() {
+        guard let metalView = metalView, !metalView.bounds.isEmpty else {
+            return
+        }
+
```
evil159, line 632: This might be unnecessary since the metal view is not supposed to have zero size. On the other hand this shouldn't not cause any problems so it might be an extra layer of protections against the crashes related to zero size.

Sources/MapboxMaps/Foundation/MapView.swift
```diff
@@ -635,10 +638,12 @@ extension MapView: DelegatingMapClientDelegate {
...
-        let metalView = dependencyProvider.makeMetalView(frame: bounds, device: metalDevice)
-        displayCallback = {
-            metalView.setNeedsDisplay()
-        }
+        let minSize = CGSize(width: 1, height: 1)
+        let finalSize = CGSize(
+            width: max(minSize.width, bounds.width),
+            height: max(minSize.height, bounds.height))
+
+        let metalView = dependencyProvider.makeMetalView(frame: CGRect(origin: .zero, size: finalSize), device: metalDevice)
```
macdrevx, line 646: Maybe cgrectunion could be of use here

macdrevx (approved):

OdNairy (approved):

Sources/MapboxMaps/Foundation/MapView.swift
evil159: Much better this way 👍

PR #1347, opened 2022-05-23
Feature/add puck mercator scale

maios:
Add mercator scale to 3D puck. 

| Before | After |
| --- | --- |
| <video src="https://user-images.githubusercontent.com/30261043/174052695-44f7b750-9480-45b3-8238-94f9088ad028.mp4"> | <video src="https://user-images.githubusercontent.com/30261043/174052780-2fc049f0-0b75-4882-925f-ab13730c418a.mp4"> |
| <video src="https://user-images.githubusercontent.com/30261043/174052856-7afc16bf-d51f-4725-97fb-c2914b17b07c.mp4"> | <video src="https://user-images.githubusercontent.com/30261043/174053250-7498cac2-e713-4c01-a617-d9bca80dcc5f.mp4"> |

## Pull request checklist:

MAPSIOS-168

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ Mapbox welcomes participation and contributions from everyone.
 ## main
 
 * Introduce ModelLayer experimental API to render 3D models on the map. ([#1348](https://github.com/mapbox/mapbox-maps-ios/pull/1348))
+* Add mercator scale factor to 3D puck, so that the 3D puck size won't increase as latitude increases. ([#1347](https://github.com/mapbox/mapbox-maps-ios/pull/1347))
 
 ## 10.6.0-beta.1 - May 20, 2022
 


evil159 (commented):
When covering large distances the 3D location puck seems to grow bigger when approaching both poles, just before setting into the final size. The location puck's size appears to be normal when the puck is stationary(even around the poles). 

https://user-images.githubusercontent.com/1478430/169855550-f2a84dc4-e820-4229-bc6a-5f105ac4e7d9.mp4

The issue is less pronounced but still visible when the location puck moves by a smaller magnitude:

https://user-images.githubusercontent.com/1478430/169856061-cb269a62-343e-4a49-bc5a-281edca5063f.mp4

Could it be a caused by too big of a threshold on the latitude change?

Tests/MapboxMapsTests/Location/Puck/Puck3DTests.swift
```diff
@@ -207,10 +206,33 @@ final class Puck3DTests: XCTestCase {
...
+        let modelLayer = try XCTUnwrap(style.addPersistentLayerStub.invocations.first?.parameters.layer as? ModelLayer)
+        let modelScaleString = try XCTUnwrap(try modelLayer.paint?.modelScale?.jsonString())
+
+        let modelScalePattern =
+            #"^\["interpolate","# +
+            #"\["exponential",0.5\],"# +
+            #"\["zoom"\],"# +
+            #"0.5,\["literal",\[(?:[0-9]+(?:\.[0-9]+)*,?){3}\]\],"# +
+            #"22,\["literal",\[(?:[0-9]+(?:\.[0-9]+)*,?){3}\]\]\]$"#
+        XCTAssertNotNil(modelScaleString.range(of: modelScalePattern, options: .regularExpression))
```
pengdev, line 233: this looks cleaner than the Android implementation 😄

Sources/MapboxMaps/Location/Puck/Puck3D.swift
```diff
@@ -89,15 +95,69 @@ internal final class Puck3D: Puck {
...
+        // Mercator scale
+        maybeUpdateMercatorScale(at: location.coordinate.latitude)
+
         // create the layer if needed
         if !style.layerExists(withId: Self.layerID) {
             var modelLayer = ModelLayer(id: Self.layerID)
             modelLayer.source = Self.sourceID
             assert(modelLayer.paint != nil)
             modelLayer.paint?.modelLayerType = .constant(.locationIndicator)
-            modelLayer.paint?.modelScale = configuration.modelScale
```
pengdev, line 98: so previously there's no default scale expression for the 3D puck?

Sources/MapboxMaps/Location/Puck/Puck3D.swift
```diff
@@ -89,15 +95,69 @@ internal final class Puck3D: Puck {
...
             modelLayer.paint?.modelRotation = configuration.modelRotation
+            updateModelScale(layer: &modelLayer)
             try! style.addPersistentLayer(modelLayer, layerPosition: nil)
         }
     }
+
+    private func maybeUpdateMercatorScale(at latitude: Double) {
+        let validLatitudeRange = -85.051128779806604...85.051128779806604
```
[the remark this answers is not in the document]
pengdev: are these `latitudeMin` `latitudeMax` actually defined as public property inside `Projection` on iOS?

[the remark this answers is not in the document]
evil159: Yes, they are 
https://github.com/mapbox/mapbox-maps-ios/blob/c61ccd41febfb2df2ed35e8155a566f37b8163b6/Sources/MapboxMaps/Foundation/Projection.swift#L3-L5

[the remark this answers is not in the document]
pengdev: ok, seems like this is a misalignment, seems it is exposed to enable better gestures UX? cc @macdrevx

Sources/MapboxMaps/Location/Puck/Puck3D.swift
```diff
@@ -89,15 +95,69 @@ internal final class Puck3D: Puck {
...
+                            scales.map { $0 * mercatorScale }
+                        }
+                    }
+                )
+
+            default: return configuration.modelScale
+            }
+        }()
+
+        layer.paint?.modelScale = modelScale
```
pengdev, line 161: this is not applied immediately, right? you need to set the whole layer(including other properties) to the style afterwards?

Sources/MapboxMaps/Location/Puck/Puck3D.swift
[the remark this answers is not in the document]
evil159: Looks like it got exposed along with the `ResourceOptionManager` https://github.com/mapbox/mapbox-maps-ios/pull/396

Sources/MapboxMaps/Location/Puck/Puck3D.swift
pengdev: I think this might be the reason that the model is not scaled as rapidly as Android, you can set the model scale property separately from the rest of the model layer properties, using `style.setStyleLayerProperty(layerId, "model-scale", scaleExpression)`. This way, the UI probably should look smoother.

Sources/MapboxMaps/Location/Puck/Puck3D.swift
maios: afaik, we don't have default scale expression for 3D puck previously, but @macdrevx can correct me :-D

pengdev: so the puck will always be the real-world size previously?

macdrevx: @maios is correct. modelScale defaulted to `nil`.

Sources/MapboxMaps/Location/Puck/Puck3D.swift
```diff
@@ -89,15 +95,69 @@ internal final class Puck3D: Puck {
             try! style.addSource(source, id: Self.sourceID)
         }
 
+        // Mercator scale
+        maybeUpdateMercatorScale(at: location.coordinate.latitude)
+
         // create the layer if needed
         if !style.layerExists(withId: Self.layerID) {
```
macdrevx, line 102: There seems to be a problem here — the first time this code runs, there could be an attempt to update the layer before it is created.

Sources/MapboxMaps/Location/Puck/Puck3D.swift
```diff
@@ -89,15 +95,69 @@ internal final class Puck3D: Puck {
...
             assert(modelLayer.paint != nil)
             modelLayer.paint?.modelLayerType = .constant(.locationIndicator)
-            modelLayer.paint?.modelScale = configuration.modelScale
             modelLayer.paint?.modelRotation = configuration.modelRotation
+            updateModelScale(layer: &modelLayer)
             try! style.addPersistentLayer(modelLayer, layerPosition: nil)
         }
     }
+
+    private func maybeUpdateMercatorScale(at latitude: Double) {
```
macdrevx, line 113: The `IfNeeded` naming convention would be a good fit here: `updateModelScaleIfNeeded`

Sources/MapboxMaps/Location/Puck/Puck3D.swift
```diff
@@ -89,15 +95,69 @@ internal final class Puck3D: Puck {
...
+                        maxZoom
+                        Exp(.literal) {
+                            scales.map { $0 * mercatorScale }
+                        }
+                    }
+                )
+
+            default: return configuration.modelScale
+            }
+        }()
```
macdrevx, line 159: No need for a closure here…

```
        let modelScale: Value<[Double]>?
        switch configuration.modelScale {
        case .constant(let scale):
            let maxZoom = 22.0
            let minZoom = 0.5
            // To make the 3D puck's size constant across different zoom levels, the 3D puck's size (real world object size)
            // should be exponential to the zoom level.
            // The base of the exponential expression is decided by how the tile pyramid works:
            // at zoom level n, we have 2^(n+1) tiles to cover the earth.
            let exponentialBase = 0.5
            modelScale = .expression(
                Exp(.interpolate) {
                    Exp(.exponential) { exponentialBase }
                    Exp(.zoom)
                    minZoom
                    Exp(.literal) {
                        scale.map { scale -> Double in
                            let modelScale = pow(2.0, maxZoom - minZoom)
                            return modelScale * scale * mercatorScale
                        }
                    }
                    maxZoom
                    Exp(.literal) {
                        scale.map { $0 * mercatorScale }
                    }
                }
            )
        default:
            modelScale = configuration.modelScale
        }
```

Sources/MapboxMaps/Location/Puck/Puck3D.swift
macdrevx: > this is not applied immediately, right?

Whether true or not, I think the fact that it's not clear suggests that some refactoring is in order to make the flow easier to follow.

> you can set the model scale property separately from the rest of the model layer properties, using style.setStyleLayerProperty(layerId, "model-scale", scaleExpression)

+1. I'd like to see a "create or update" pattern similar to how the source is managed (see lines 85–90 of original)

macdrevx (approved):

Sources/MapboxMaps/Location/Puck/Puck3D.swift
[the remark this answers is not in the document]
pengdev: ok, I think we can keep it as is, as these are public constants from long time ago.

Sources/MapboxMaps/Location/Puck/Puck3D.swift
pengdev: ok, I guess it's good to have the alignment now 😄

Sources/MapboxMaps/Location/Puck/Puck3D.swift
```diff
@@ -59,14 +60,18 @@ internal final class Puck3D: Puck {
...
-                Log.warning(
-                    forMessage: "Puck3DConfiguration.model.orientation?.count must be 3 or nil. Actual orientation is \(invalidOrientation). Resetting it to [0, 0, 0].",
-                    category: "Puck")
-            }
-            model.orientation = [0, 0, 0]
-        }
+
+        model.orientation = model.orientation
+            .flatMap { orientation -> [Double]? in
+                guard orientation.count == 3 else {
```
pengdev, line 66: interesting to read fancy swift code 👍

Sources/MapboxMaps/Style/Style.swift
```diff
@@ -9,6 +9,7 @@ internal protocol StyleProtocol: AnyObject {
     func removeLayer(withId id: String) throws
     func layerExists(withId id: String) -> Bool
     func setLayerProperties(for layerId: String, properties: [String: Any]) throws
+    func setLayerProperty(for layerId: String, property: String, value: Any) throws
```
pengdev, line 12: so you didn't expose this method previously for iOS? This would be a great way to update single style layer property. Would be good to mention in the change log.

BTW, how did you track the public API changes on iOS SDK?

Sources/MapboxMaps/Location/Puck/Puck3D.swift
```diff
@@ -89,14 +94,72 @@ internal final class Puck3D: Puck {
...
+            modelLayer.modelScale = modelScale
             modelLayer.modelType = .constant(.locationIndicator)
-            modelLayer.modelScale = configuration.modelScale
             modelLayer.modelRotation = configuration.modelRotation
             try! style.addPersistentLayer(modelLayer, layerPosition: nil)
+        } else if needsUpdateModelScale {
+            try? style.setLayerProperty(
+                for: Self.layerID,
+                property: "model-scale",
+                value: modelScale?.toJSON() as Any)
```
pengdev, line 112: probably you should also update the model rotation as well, the model rotation is driven by the location updates as well.

Sources/MapboxMaps/Style/Style.swift
pengdev: emm.. ok, so this API is available in the public-facing Style class previously?

pengdev: ok, thanks for the clarification.

Sources/MapboxMaps/Location/Puck/Puck3D.swift
pengdev: Resolved as the bearing is updated through model source property, which is more performant.

pengdev (approved):

PR #1348, opened 2022-05-23
ModelLayer support

evil159:
This PR adds support for ModelLayer. 
New API:
```swift
@_spi(Experimental) ModelLayer
@_spi(Experimental) Style.addModel(id:modelURI:) throws
@_spi(Experimental) Style.removeModel(id:) throws
@_spi(Experimental) Style.modelExists(id:) -> Bool
ModelLayerType -> ModelType
```

Adds `ModelLayerExample`:
Simulator Screen Shot - iPhone 13 - 2022-05-23 at 14 48 35

## Pull request checklist:


Sources/MapboxMaps/Style/ModelLayer.swift
```diff
@@ -1,96 +0,0 @@
...
-
-        enum CodingKeys: String, CodingKey {
-            case visibility
-        }
-    }
-
-    /// Changes to a paint property are cheap and happen synchronously.
-    internal var paint: ModelLayer.Paint?
-
-    internal struct Paint: Codable {
```
pengdev, line 40: seems the layer properties are not grouped into paint and layout properties any more, I guess this is generated the same way with other layers?

Apps/Examples/Examples/All Examples/ModelLayerExample.swift
```diff
@@ -0,0 +1,68 @@
...
+
+        try! style.addSource(source, id: Constants.sourceId)
+
+        var layer = ModelLayer(id: "model-layer-id")
+        layer.source = Constants.sourceId
+        layer.modelId = .expression(Exp(.get) { Constants.modelIdKey })
+        layer.modelType = .constant(.common3d)
+        layer.modelScale = .constant([100, 100, 100])
+
+        try! style.addLayer(layer)
```
pengdev, line 60: Is there a `getLayer` methods in iOS? do we need to expose `ModelLayer` for that API?

Sources/MapboxMaps/Style/ModelLayer.swift
evil159: Yeah, generated layer properties are unrolled into a flat list

Apps/Examples/Examples/All Examples/ModelLayerExample.swift
evil159: There is `layer(withId:)` and `layer(withId:type:)`, the latter one supports any layer type as long as it conforms to the `Layer` protocol
https://github.com/mapbox/mapbox-maps-ios/blob/c61ccd41febfb2df2ed35e8155a566f37b8163b6/Sources/MapboxMaps/Style/Style.swift#L103-L106

pengdev (approved):

Sources/MapboxMaps/Style/Style.swift
```diff
@@ -989,6 +989,42 @@ public final class Style: StyleProtocol {
...
+
+    /// Checks whether a given style model exists.
+    ///
+    /// - Parameter id: Style model identifier.
+    ///
+    /// - Returns: `true` if the given model exists, `false` otherwise.
+    @_spi(Experimental) public func modelExists(withId id: String) -> Bool {
+        styleManager.hasStyleModel(forModelId: id)
+    }
+
```
macdrevx, line 1027: Do these replace model source?

macdrevx (approved):

Sources/MapboxMaps/Style/Style.swift
astojilj: Good point.

> Do these replace model source?
Gl-native won't deprecate model source API and it will stay as is (_private_ API between mobile and native SDKs) since the API is designed around performant property update. At least for now.

> (and should Puck3D be updated to use these?)
These APIs allow instancing, by assigning 3D models to vector source's point geometry features. While feature state API  or updating geojson source allows updating individual instances, it is still expected to cost more that calling model source methods directly. It is experimental API and underlying implementation is expected to improve so we should revisit this and potential model source code removal.

hegelstad:
Maybe a stupid question, but can Mapbox SDK for IOS render 3D objects in place of map markers? Is animation possible?

joshkopecek:
@hegelstad I was looking to do the exact same thing. They introduced the feature experimentally in May, only to remove it later. You can still add a 3D 'puck' as the user location, but looks like the ModelLayer was internalised, and is not listed in the 'supported layer types'. It seems there's no way of doing this.

PR #1362, opened 2022-05-27
Well-formed map event types

evil159:
This PR adds a way to get a type-safe data for a subscribed event.

**Before**
```swift
mapView.mapboxMap.onNext(.styleDataLoaded) { event in
    guard let data = event.data as? [String: Any],
          let response = data["response"] as? [String: Any],
          let source = data["source"] as? String else {
        return
    }

    if source == "tile-store" {
        ...
    }
}
```

**After**
```swift
mapView.mapboxMap.onTypedNext(.resourceRequest) { event in
    event.payload.response.source == .tileStore {
        ...
    }
}
```

Public API additions:
* `MapboxMap.onTypedNext(_:handler:)`
* `MapboxMap.onTypedEvery(_:handler:)`
* `MapEvents.Event`(with event types defined as immutable static variables)
* `TypedEvent` for wrapping `MBMEvent` and carrying type-safe payload

Considerations:
* `payload` property carrying the type-safe event data is a `lazy` property to avoid performance hit of encoding the untyped event data into binary data and then back into typed payload
* unit tests for typed events are not that straightforward

This PR also deprecates the old `onNext/Every` methods, as well as `MapEvents.EventKind` with the intention to remove them in the next major release.

## Pull request checklist:


Sources/MapboxMaps/Foundation/Extensions/Core/MapEvents+Deprecated.swift
```diff
@@ -0,0 +1,270 @@
+import Foundation
+
+public extension MapEvents {
+    /// Events emitted by the SDK
+    @available(*, deprecated, message: "Use 'Event' instead")
+    enum EventKind: RawRepresentable, CaseIterable {
```
OdNairy, line 6: Does it generate a deprecation warning in the type declaration line of code OR in the line we use the concrete case (like `.mapLoaded`)?

Sources/MapboxMaps/Foundation/Extensions/Core/MapEvents.swift
```diff
@@ -68,274 +68,98 @@ import Foundation
...
+    /// Typed events emitted by the SDK
+    struct Event<Payload: Decodable> {
+        /// The style has been fully loaded, and the map has rendered all visible tiles.
+        public static var mapLoaded: Event<NoPayload> { .init(name: MapEvents.mapLoaded) }
+
+        /// Describes an error that has occurred while loading the Map. The `type` property defines what resource could
+        /// not be loaded and the `message` property will contain a descriptive error message.
+        /// In case of `source` or `tile` loading errors, `source-id` will contain the id of the source failing.
+        /// In case of `tile` loading errors, `tile-id` will contain the id of the tile.
+        public static var mapLoadingError: Event<MapLoadingErrorPayload> { .init(name: MapEvents.mapLoadingError) }
```
OdNairy, line 83: As we already reimplementing event types, should we consider unionization of these events? 
It's now possible to have a single event with `Result<Void, MapLoadingErrorPayload>` in payload type, right?

Sources/MapboxMaps/Foundation/Events/Payload/NoPayload.swift
```diff
@@ -0,0 +1,5 @@
+import Foundation
+
+public struct NoPayload: Decodable {
```
OdNairy, line 3: My recommendation is to use empty-case `enum` over `struct` as it generates less memory fingerprint. Little kind of phantom type.

Sources/MapboxMaps/Foundation/Extensions/Core/MapEvents.swift
```diff
@@ -68,274 +68,98 @@ import Foundation
...
-                return MapEvents.renderFrameStarted
-            case .renderFrameFinished:
-                return MapEvents.renderFrameFinished
-            case .cameraChanged:
-                return MapEvents.cameraChanged
-            case .resourceRequest:
-                return MapEvents.resourceRequest
-            }
+    /// Typed events emitted by the SDK
+    struct Event<Payload: Decodable> {
```
OdNairy, line 75: TBD:
I'm not quite sure that we must have a `Decodable` requirement in that place. It's not a part of the requirement for the Event, it's more implementation detail.

Sources/MapboxMaps/Foundation/MapboxMap.swift
```diff
@@ -78,21 +78,15 @@ public final class MapboxMap: MapboxMapProtocol {
...
-                fatalError("Unexpected event type")
+        onTypedNext(.styleLoaded) { _ in
+            if !self.style.isLoaded {
+                Log.warning(forMessage: "style.isLoaded == false, was this an empty style?", category: "Style")
             }
+            completion(.success(self.style))
+        }
+
+        onTypedNext(.mapLoadingError) { event in
+            completion(.failure(event.payload.error))
```
OdNairy, line 89: Not quite related to this PR, but I assume we have to save cancellables for this calls.

Sources/MapboxMaps/Foundation/MapboxMap.swift
```diff
@@ -860,9 +854,25 @@ extension MapboxMap {
...
+    ///   - handler: The closure to execute when the event occurs.
+    ///
+    /// - Returns: A `Cancelable` object that you can use to stop listening for
+    ///     the event. This is especially important if you have a retain cycle in
+    ///     the handler.
     @discardableResult
-    private func onNext(eventTypes: [MapEvents.EventKind], handler: @escaping (Event) -> Void) -> Cancelable {
-        return observable.onNext(eventTypes, handler: handler)
+    public func onTypedNext<Payload: Decodable>(_ eventType: MapEvents.Event<Payload>, handler: @escaping (TypedEvent<Payload>) -> Void) -> Cancelable {
+        return observable.onTypedNext(eventType, handler: handler)
```
OdNairy, line 875: Can we discuss the API design here?
I think the API would be better if we preserve the same `onNext` prefix. As it might be type-checker issues, we can make it `onNext<…>(event: MapsEvents.Event<…>)`
What do you think?

OdNairy (changes requested):
Great job! The new Event API looks amazing. I'm confident that a lot of people would be happy with that.
Until we merge this PR and publish this API, I'd like to discuss some new opportunities.

Sources/MapboxMaps/Foundation/Events/Payload/NoPayload.swift
maios: Instead of doing this, what do you think if we have a marker protocol called `MapEventType`, which has 2 conformers
- `HasPayloadMapEvent` represents an event with some payload.
- `EmptyPayloadMapEvent` represents an event with zero payload.

```
public protocol MapEventType {}

public struct HasPayloadMapEvent<Payload: Decodable>: MapEventType {
    private let event: MapboxCoreMaps.Event
    public var eventName: String { event.type }

    public lazy var payload: Payload! = try? event.typedPayload()

    internal init(event: MapboxCoreMaps.Event) {
        self.event = event
    }
}

public struct EmptyPayloadMapEvent:MapEventType {
    private let event: MapboxCoreMaps.Event
    public var eventName: String { event.type }

    internal init(event: MapboxCoreMaps.Event) {
        self.event = event
    }
}
```
then we would constraint `MapEvents.Event` to this `MapEventType`, and the subscription will have a benefit of knowing which type of event it is and whether it has payload or not.

Sources/MapboxMaps/Style/StyleErrors.swift
```diff
@@ -33,7 +33,7 @@ public enum TypeConversionError: Error {
 ///
 /// The associated message (which is returned by `errorDescription`) contains
 /// a descriptive error message.
-public enum MapLoadingError: LocalizedError {
+public enum MapLoadingError: LocalizedError, Decodable {
```
maios, line 36: Is this necessary? I don't see where `MapLoadingError` is usable as Decodable.

maios (commented):
Great job! I only have a couple of comments and questions, 
I wish all the event payload model can be generated :-D

Sources/MapboxMaps/Foundation/Events/Payload/NoPayload.swift
OdNairy: Sounds interesting!
But I do not quite understand how would the subscription know what type of payload it has?

Sources/MapboxMaps/Foundation/Extensions/Core/MapEvents.swift
evil159: Interesting idea, I guess we could, indeed, have a single event, however it would be inconsistent with other events and a bit messy to implement(this combined event maps to two `MapsEvents`).

Sources/MapboxMaps/Foundation/Events/Payload/NoPayload.swift
maios:
```
public extension MapEvents {
  /// Typed events emitted by the SDK
  struct Event<EventType: MapEventType> {
    /// The style has been fully loaded, and the map has rendered all visible tiles.
    public static var mapLoaded: Event<EmptyPayloadMapEvent> { .init(name: MapEvents.mapLoaded) }

    /// Describes an error that has occurred while loading the Map. The `type` property defines what resource could
    /// not be loaded and the `message` property will contain a descriptive error message.
    /// In case of `source` or `tile` loading errors, `source-id` will contain the id of the source failing.
    /// In case of `tile` loading errors, `tile-id` will contain the id of the tile.
    public static var mapLoadingError: Event<HasPayloadMapEvent<MapLoadingError>> { .init(name: MapEvents.mapLoadingError) }
}

@discardableResult
public func onNext<EventType>(event: MapEvents.Event<EventType>, handler: @escaping (EventType) -> Void) -> Cancelable {}

mapboxMap.onNext(event: .mapLoaded) { event in
  // event.eventName
  // event.payload <-- Value of type 'EmptyPayloadMapEvent' has no member 'payload'
}

mapboxMap.onNext(event: .mapLoadingError) { event in 
  // event.eventName
  // event.payload <-- MapLoadingError
}
```
It was something like this, do you think it makes more sense 🤔

OdNairy: > It was something like this, do you think it makes more sense 🤔

Nice idea, I like it!
Should we discuss the naming?

Sources/MapboxMaps/Foundation/Extensions/Core/MapEvents.swift
OdNairy: As we discussed internally, there is no way to avoid this requirement. The only alternative is to write down a static list of each event and manual conversion plan which has no benefits over the `Decodable` requirement.

Sources/MapboxMaps/Foundation/Extensions/Core/MapEvents.swift
OdNairy: One of my points was that migration to the async/await functions would be more native if these events were combined.
However, I just understand that the function `onNext(event:)` should be `throws` by default as it is not a thing for most of the events.
Probably, we have to keep the layout of the existing events until we find some better way to utilize that.

OdNairy: I also assume that `Result` does not confirm `Decodable` so we cannot use it as a payload type.

Sources/MapboxMaps/Foundation/Events/Payload/NoPayload.swift
evil159: As we confirmed during an in-person discussion - changes to public protocols often result in a breaking change, so public protocols should only be introduced in situations where there's a reason for developers to implement the protocol. 
While this makes for a better API, doing this will require us to do expose a public protocol that's not supposed to be conformed to.

Sources/MapboxMaps/Foundation/Extensions/Core/MapEvents.swift
evil159: Apparently, the compiler doesn't complain if this requirement is lifed. Payload is still restricted to be Decodable in `onNext/Every` methods so that is enough.

Sources/MapboxMaps/Foundation/MapboxMap.swift
evil159: Nice idea, I renamed `onTypedNext/onTypedEvery` to `onNext(event:.../onEvery(event:...`

Sources/MapboxMaps/Foundation/Extensions/Core/MapEvents.swift
OdNairy: Sounds like we should remove `: Decodable` requirement in `onNext/onEvery` and left a single on Event itself

evil159: Good point, this seems to work as well :)

OdNairy:
@evil159 Can you please fix the mobile-metrics repo as well? 
image

Sources/MapboxMaps/Foundation/Events/Payload/ResourceRequestPayload.swift
```diff
@@ -0,0 +1,76 @@
...
+
+    public struct Error: Swift.Error, Decodable {
+        // swiftlint:disable:next nesting
+        public enum Reason: String, Decodable {
+            case success
+            case notFound = "not-found"
+            case server
+            case connection
+            case rateLimit = "rate-limit"
+            case inOfflineMode = "in-offline-mode"
```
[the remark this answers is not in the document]
evil159: Yeah, not sure about an external library but we can think about implementing one, probably outside of the scope of this PR

Sources/MapboxMaps/Foundation/MapboxMap.swift
evil159: I guess we could cancel the previous subscriptions when trying to load a new style before a previous one is loaded, e.g:
```
mapboxMap.loadStyle(.light) { _ in }
mapboxMap.loadStyle(.satellite) { _ in } <- this call would cancel the `styleLoaded` and `mapLoadingError` observations from the previous line
```

Sources/MapboxMaps/Foundation/Events/Payload/MapLoadingErrorPayload.swift
```diff
@@ -0,0 +1,51 @@
...
+import MapboxCoreMaps
+
+public struct MapLoadingErrorPayload {
+    public let error: MapLoadingError
+    public let sourceId: String?
+    public var tileId: CanonicalTileID? {
+        return decodedTileId?.tileID
+    }
+
+    internal let decodedTileId: DecodableCanonicalTileID?
```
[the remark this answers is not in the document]
evil159: I think shadowing `CanonicalTileID` would be a binary breaking change(though a source compatible one), we'll need to discuss if that's acceptable 
cc @OdNairy

evil159:
> @evil159 Can you please fix the mobile-metrics repo as well?
@OdNairy Not sure if somebody fixed it already, but the failure went away magically 🪄

OdNairy:
> > @evil159 Can you please fix the mobile-metrics repo as well?
> > @OdNairy Not sure if somebody fixed it already, but the failure went away magically 🪄

It's quite interesting that the actual issue was that the new binary size of SDK was too big. We have some pinned value and a 10% tolerance. Now you are under the limit.

OdNairy:
Oh, actually I cannot find any successful binary-size job. They are still in a pending state.

Sources/MapboxMaps/Foundation/Events/Payload/ResourceRequestPayload.swift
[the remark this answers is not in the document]
OdNairy: I believe it would be better to implement it as a part of this PR. There are tons of new `CodingKeys` that can be eliminated with appropriate `KeyDecodingStrategy`. It's also safer to run the "experimental" decoding strategy over some scope that you are very familiar with and have a total control.

The implementation also looks pretty straightforward. Something like:

```swift
decoder.keyDecodingStrategy = .custom { keys in
    func camelCased(_ string: String) -> String {
        return string.lowercased()
            .split(separator: "-")
            .enumerated()
            .map { $0.offset > 0 ? $0.element.capitalized : $0.element.lowercased() }
            .joined()
    }

    let lastKey = keys.last!.stringValue

    return AnyKey(stringValue: camelCased(lastKey))!
}
```

We also have to introduce `AnyKey` (intValue is required by CodingKey):

```swift
struct AnyKey: CodingKey {
    var stringValue: String
    var intValue: Int?

    init?(stringValue: String) {
        self.stringValue = stringValue
        self.intValue = nil
    }

    init?(intValue: Int) {
        self.stringValue = String(intValue)
        self.intValue = intValue
    }
}
```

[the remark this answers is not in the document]
OdNairy: I can confirm that this implementation works 😎 
image

Sources/MapboxMaps/Foundation/MapboxObservable.swift
```diff
@@ -3,8 +3,12 @@ import MapboxCoreMaps
 internal protocol MapboxObservableProtocol: AnyObject {
     func subscribe(_ observer: Observer, events: [String])
     func unsubscribe(_ observer: Observer, events: [String])
+    @available(*, deprecated)
     func onNext(_ eventTypes: [MapEvents.EventKind], handler: @escaping (Event) -> Void) -> Cancelable
+    @available(*, deprecated)
```
OdNairy, line 8: Can we add `renamed:` part here as well?

Sources/MapboxMaps/Foundation/MapboxObservable.swift
```diff
@@ -92,6 +96,27 @@ internal final class MapboxObservable: MapboxObservableProtocol {
...
+        // avoids a retain cycle. MapboxObservable holds a strong reference to observer, which has a
+        // strong reference to cancelable, which has a strong reference to BlockCancelable, which only
+        // has weak references back to MapboxObservable and observer. If MapboxObservable is deinited,
+        // observer will be released.
+        cancelable.add(BlockCancelable { [weak self, weak observer] in
+            if let self = self, let observer = observer {
+                self.unsubscribe(observer, events: [])
+            }
+        })
+        return cancelable
```
OdNairy, line 116: As we are going to support deprecated methods for a while, can we refactor it to reduce duplicated code?
Third function with `[String]` as an argument looks nice.

Sources/MapboxMaps/Foundation/MapboxObservable.swift
```diff
@@ -102,6 +127,18 @@ internal final class MapboxObservable: MapboxObservableProtocol {
...
+        let observer = BlockObserver {
+            handler(MapEvent(event: $0))
+        }
+        subscribe(observer, events: [event.name])
+        return BlockCancelable { [weak self, weak observer] in
+            if let self = self, let observer = observer {
+                self.unsubscribe(observer, events: [])
+            }
+        }
+    }
```
OdNairy, line 140: Same here

Sources/MapboxMaps/Snapshot/Snapshotter.swift
```diff
@@ -305,11 +305,33 @@ extension Snapshotter: MapEventsObservable {
...
+    /// - Parameters:
+    ///   - eventType: The event type to listen to.
+    ///   - handler: The closure to execute when the event occurs.
+    ///
+    /// - Returns: A `Cancelable` object that you can use to stop listening for
+    ///     the event. This is especially important if you have a retain cycle in
+    ///     the handler.
+    @discardableResult
+    public func onNext<Payload>(event: MapEvents.Event<Payload>, handler: @escaping (MapEvent<Payload>) -> Void) -> Cancelable {
+        return observable.onNext(event: event, handler: handler)
```
OdNairy, line 332: I don't think that we actually need a refactor in these methods as they are just proxies.

Sources/MapboxMaps/Foundation/Events/Payload/ResourceRequestPayload.swift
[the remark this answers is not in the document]
OdNairy: Can we still implement it as a part of this PR?

Sources/MapboxMaps/Foundation/MapboxObservable.swift
evil159: Nice idea! Refactored.

Sources/MapboxMaps/Foundation/Extensions/KebabCaseKeyDecodingStrategy.swift
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    init?(intValue: Int) {
+        self.stringValue = String(intValue)
+        self.intValue = intValue
+    }
+}
+
+private extension String {
+    var camelCased: String {
```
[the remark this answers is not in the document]
OdNairy: I think we have to change it to the function to utilize separator argument.

Sources/MapboxMaps/Foundation/MapboxObservable.swift
```diff
@@ -71,14 +75,24 @@ internal final class MapboxObservable: MapboxObservableProtocol {
             }
         }
     }
-
+    @available(*, deprecated)
```
OdNairy, line 78: If we already applied deprecation on Protocol, should we duplicate it in implementation? 🤔 
And if we do, should we also copy-paste `renamed` part?

Tests/MapboxMapsTests/Foundation/Extensions/KebabCaseKeyDecodingStrategyTests.swift
```diff
@@ -0,0 +1,59 @@
...
+
+        XCTAssertEqual(converter([TestingKey(intValue: 0)!]).stringValue, "0")
+        XCTAssertEqual(converter([TestingKey("")]).stringValue, "")
+        XCTAssertEqual(converter([TestingKey("key")]).stringValue, "key")
+        XCTAssertEqual(converter([TestingKey("-key")]).stringValue, "key")
+        XCTAssertEqual(converter([TestingKey("key-")]).stringValue, "key")
+        XCTAssertEqual(converter([TestingKey("-key-")]).stringValue, "key")
+        XCTAssertEqual(converter([TestingKey("a-key")]).stringValue, "aKey")
+        XCTAssertEqual(converter([TestingKey("a-long-key")]).stringValue, "aLongKey")
+        XCTAssertEqual(converter([TestingKey("a-very-long-key")]).stringValue, "aVeryLongKey")
```
OdNairy, line 39: Can we also add few more keys:
1. `12-3-a-key`
2. `a-key-1-23`

Tests/MapboxMapsTests/Foundation/Extensions/KebabCaseKeyDecodingStrategyTests.swift
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    func testIntConversion() throws {
+        guard case let JSONDecoder.KeyDecodingStrategy.custom(converter) = .convertFromKebabCase else {
+            XCTFail("Kebab case strategy is a not custom decoding strategy.")
+            return
+        }
+
+        XCTAssertEqual(converter([TestingKey(intValue: 10325346)!]).stringValue, "10325346")
```
OdNairy, line 48: Does it duplicates `testStringConversion` first assert?
`XCTAssertEqual(converter([TestingKey(intValue: 0)!]).stringValue, "0")`

Tests/MapboxMapsTests/Foundation/MapboxObservableTests.swift
```diff
@@ -270,6 +469,56 @@ final class MapboxObservableTests: XCTestCase {
...
+            { try verifyInvocation(eventType: .styleLoaded) },
+            { try verifyInvocation(eventType: .styleImageMissing) },
+            { try verifyInvocation(eventType: .styleImageRemoveUnused) },
+            { try verifyInvocation(eventType: .sourceDataLoaded) },
+            { try verifyInvocation(eventType: .sourceAdded) },
+            { try verifyInvocation(eventType: .sourceRemoved) },
+            { try verifyInvocation(eventType: .renderFrameStarted) },
+            { try verifyInvocation(eventType: .renderFrameFinished) },
+            { try verifyInvocation(eventType: .cameraChanged) },
+            { try verifyInvocation(eventType: .resourceRequest) }
```
OdNairy, line 514: Could we use `CaseIterable` to apply for each verification for each case?

OdNairy: Even if we cannot, we have to share the list of cases over duplication.

Tests/MapboxMapsTests/Foundation/MapboxObservableTests.swift
```diff
@@ -270,6 +469,56 @@ final class MapboxObservableTests: XCTestCase {
...
+        ]
+        // swiftlint:enable opening_brace
+
+        try eventInvocations.randomElement()!()
```
OdNairy, line 518: Why do we go for a random one instead of verifying each of them?

OdNairy (commented):
The new approach with custom key decoding strategy looks great 😍
I'd like to make a few changes before we merge:
1. Write each case on the new line. Avoid putting two, three or even more cases in single line
2. There are still a few places where we have associated rawValue. They looks unnessesary if we use kebabCase key strategy.

Sources/MapboxMaps/Foundation/Events/Payload/ResourceRequestPayload.swift
```diff
@@ -0,0 +1,60 @@
+import Foundation
+
+public struct ResourceRequestPayload: Decodable {
+    public let dataSource: ResourceDataSource
+    public let request: ResourceRequest
+    public let response: ResourceResponse?
+}
+
+public enum ResourceDataSource: String, Decodable {
+    case resourceLoader = "resource-loader", network, database, asset, fileSystem = "file-system"
```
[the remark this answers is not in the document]
evil159: This is a "model" enum, so it's cases don't get transformed in the same way as `CodingKey`-conformant enum cases.

Sources/MapboxMaps/Foundation/Events/Payload/ResourceRequestPayload.swift
```diff
@@ -0,0 +1,60 @@
...
+}
+
+public struct ResourceRequest: Decodable {
+    public enum Kind: String, Decodable {
+        case unknown, style, source, tile, glyphs, spriteImage = "sprite-image", spriteJSON = "sprite-json", image
```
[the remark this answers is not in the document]
evil159: Same reason as https://github.com/mapbox/mapbox-maps-ios/pull/1362#discussion_r897878636

Sources/MapboxMaps/Foundation/Events/Payload/ResourceRequestPayload.swift
```diff
@@ -0,0 +1,60 @@
...
+
+    let url: String
+    let kind: Kind
+    let priority: Priority
+    let loadingMethod: [LoadingMethod]
+}
+
+public struct ResourceResponse: Decodable {
+    public enum Source: String, Decodable {
+        case network, cache, tileStore = "tile-store", localFile = "local-file"
```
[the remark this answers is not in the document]
evil159: Same thing as with https://github.com/mapbox/mapbox-maps-ios/pull/1362#discussion_r897878636

Sources/MapboxMaps/Foundation/Events/Payload/ResourceRequestPayload.swift
```diff
@@ -0,0 +1,60 @@
...
+
+    public struct Error: Swift.Error, Decodable {
+        // swiftlint:disable:next nesting
+        public enum Reason: String, Decodable {
+            case success
+            case notFound = "not-found"
+            case server
+            case connection
+            case rateLimit = "rate-limit"
+            case inOfflineMode = "in-offline-mode"
```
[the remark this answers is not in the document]
evil159: Same as https://github.com/mapbox/mapbox-maps-ios/pull/1362#discussion_r897878636

Sources/MapboxMaps/Foundation/MapboxObservable.swift
evil159: Good point, I missed that, added them 👍

Sources/MapboxMaps/Foundation/Events/Payload/ResourceRequestPayload.swift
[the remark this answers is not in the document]
evil159: Do you think this kind of notation doesn't quite work for enums with explicit raw values? This was intentional as the list of cases is small enough to fit on a single, 120 character wide line. I'm fine with either way for this case, just curious about your reasoning.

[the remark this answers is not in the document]
OdNairy: Easier to read, possible to add documents or comment on per-case approach

Sources/MapboxMaps/Foundation/Extensions/KebabCaseKeyDecodingStrategy.swift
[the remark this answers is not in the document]
evil159: I like the explicit version 👍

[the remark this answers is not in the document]
evil159: I think currently this is used only with dash separator and it's private to this file so future modifications are not a concern. I mean I'd still keep it as a property and turn into a method when it is needed.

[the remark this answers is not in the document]
OdNairy: We would forgot about this method when time comes to re-use it 🙂 
I see no issue to convert it to the function with argument.

Sources/MapboxMaps/Foundation/Events/Payload/ResourceRequestPayload.swift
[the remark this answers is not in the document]
evil159: What do you think about simpler cases?
```swift
enum Foo {
    case bar, baz
}
```
I think in this case it is quite acceptable to put in a single line.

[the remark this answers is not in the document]
OdNairy: You still cannot provide documentation per each case if needed.

Tests/MapboxMapsTests/Foundation/MapboxObservableTests.swift
evil159: This is something that I'm not sure how to achieve - there don't seem to be a way to declare a list of events, hence this approach.

Sources/MapboxMaps/Foundation/Events/Payload/ResourceRequestPayload.swift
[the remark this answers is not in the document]
maios: quite late to the party but I have personal rule of no more than 3 cases per line, and if one comes with a custom raw value then it goes to a separated line. This is something we should discuss and agree to have it in convention.

Tests/MapboxMapsTests/Foundation/MapboxObservableTests.swift
evil159: 1. This is how the `EventKind` events are tested
2. It's matter of balance on the scale time<->test coverage and this provides a good middle ground

OdNairy: How long this test takes?

evil159: 0.1s - random element vs 0.36s for all. There are 11 tests like this.

Sources/MapboxMaps/Foundation/Extensions/KebabCaseKeyDecodingStrategy.swift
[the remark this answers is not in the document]
evil159: Fair point, converted it to a method

Sources/MapboxMaps/Foundation/Events/Payload/ResourceRequestPayload.swift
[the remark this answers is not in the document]
evil159: I split enum case declarations into separate lines, let's discuss separately.

Sources/MapboxMaps/Foundation/Events/Payload/ResourceRequestPayload.swift
```diff
@@ -0,0 +1,76 @@
...
+    let priority: Priority
+    let loadingMethod: [LoadingMethod]
+}
+
+public struct ResourceResponse: Decodable {
+    public enum Source: String, Decodable {
+        case network
+        case cache
+        case tileStore = "tile-store"
+        case localFile = "local-file"
```
OdNairy, line 48: Do we need rawValues here?

OdNairy (approved):
Well done! 🎉

Sources/MapboxMaps/Foundation/Events/Payload/ResourceRequestPayload.swift
evil159: I think it's still the same case as in https://github.com/mapbox/mapbox-maps-ios/pull/1362#discussion_r897878636 - these are not coding keys, but rather values so they won't get automatically converted by the custom `KeyDecodingStrategy`.

OdNairy: Missed the previous answer. Thanks!

PR #1364, opened 2022-05-30
Add internal MapRecorder API

OdNairy:
## Pull request checklist:

### DIFF
--- a/Sources/MapboxMaps/Foundation/MapRecorder.swift
+++ b/Sources/MapboxMaps/Foundation/MapRecorder.swift
@@ -1,7 +1,7 @@
 import Foundation
 @_implementationOnly import MapboxCoreMaps_Private
 
-internal class MapRecorder {
+final internal class MapRecorder {
     let recorder: MapboxCoreMaps_Private.MapRecorder
 
     internal init(mapView: Map) {
--- a/Sources/MapboxMaps/Foundation/MapRecorder.swift
+++ b/Sources/MapboxMaps/Foundation/MapRecorder.swift
@@ -1,0 +1,37 @@
+import Foundation
+@_implementationOnly import MapboxCoreMaps_Private
+
+internal class MapRecorder {
+    let recorder: MapboxCoreMaps_Private.MapRecorder
+
+    internal init(mapView: Map) {
+        recorder = MapboxCoreMaps_Private.MapRecorder(map: mapView)
+    }
+
+    // MARK: Recording
+
+    internal func start() {
+        recorder.startRecording()
+    }
+
+    internal func stop() {
+        recorder.stopRecording()
+    }
+
+    // MARK: Replay
+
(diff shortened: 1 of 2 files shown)


Sources/MapboxMaps/Foundation/MapRecorder.swift
```diff
@@ -0,0 +1,48 @@
...
+#if swift(>=5.5)
+    @available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
+    func replay(content: String, playbackCount: Int = 1, playbackSpeedMultiplier: Double = 1.0) async {
+        return await withCheckedContinuation { continuation in
+            replay(content: content, playbackCount: playbackCount, playbackSpeedMultiplier: playbackSpeedMultiplier) {
+                continuation.resume(returning: ())
+            }
+        }
+    }
+#endif
```
OdNairy, line 47: @maios @evil159 What do you think about including async implementation with compilation and OS version limitation?
We still gonna support compilation on the pre swift-5.5 version due to `#if swift` and support iOS 11+ as this function is just a wrapper.

OdNairy: Unfortunately, this approach doesn't work with Xcode 13.0.0 because it expects concurrency support to be iOS 15+.
It's still possible to add one more level of `#if swift` to avoid it but I'm not quite sure that it makes sense.
Probably we have to cut off the ticket for `async/await` support.

evil159: I support this, we could limit the support for async/await methods to Xcode 13.3(swift 5.6) which will enable us to make async/await methods available on iOS 13. Just like Realm did [here](https://github.com/realm/realm-swift/commit/8349198c6e7c160bf2e5bd76260af640bfe1c82d).

evil159: I support this, we could limit the support for async/await methods to Xcode 13.3(swift 5.6) which will enable us to make async/await methods available on iOS 13. Just like Realm did [here](https://github.com/realm/realm-swift/commit/8349198c6e7c160bf2e5bd76260af640bfe1c82d).

OdNairy: @maios What do you think about this? I like the idea to limit swift 5.6 and start providing `async`/`await` support

OdNairy: I afraid it would mean we cannot use `async`/`await` API on the CI test runners which are Xcode 13.0у

Sources/MapboxMaps/Foundation/MapRecorder.swift
```diff
@@ -0,0 +1,37 @@
+import Foundation
+@_implementationOnly import MapboxCoreMaps_Private
+
+internal class MapRecorder {
```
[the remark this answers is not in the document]
OdNairy: Done

PR #1406, opened 2022-06-23
Revert encoding nil layer values

evil159:
This reverts #1400 as `null` is not a valid value for some layer properties, which results in following error's when trying to update a layer:
`Could not update background layer due to error: StyleError(rawValue: "Cannot set layer property: maxzoom error: value must be a number")`

We'd need to find a way to address this error before we can move forward with this change.


maios (approved):
Interesting! It seems that resetting a property using `Style.updateLayer` doesn't work at all for me, but using `Style.setLayerProperty` and `Style.setLayerProperties` work, so it might be that JSON doesn't translate well to layer properties.
also I found this https://github.com/mapbox/mapbox-maps-ios/blob/main/Sources/MapboxMaps/Style/Style.swift#L442 which returns a default value of a property for a Layer, imo this can be useful.

PR #1446, opened 2022-07-12
Fix view annotation losing its feature association after update

evil159:
This PR addresses the issue when a view annotation loses its feature id association when updating other(non-`associatedFeatureId`) properties.

## Pull request checklist:

### DIFF
--- a/Sources/MapboxMaps/Annotations/ViewAnnotationManager.swift
+++ b/Sources/MapboxMaps/Annotations/ViewAnnotationManager.swift
@@ -193,7 +193,7 @@ public final class ViewAnnotationManager {
         let isHidden = !(options.visible ?? true)
         expectedHiddenByView[view] = isHidden
         viewsById[id]?.isHidden = isHidden
-        if let id = currentFeatureId, id != options.associatedFeatureId {
+        if let id = currentFeatureId, let updatedId = options.associatedFeatureId, id != updatedId {
             viewsByFeatureIds[id] = nil
         }
         if let featureId = options.associatedFeatureId {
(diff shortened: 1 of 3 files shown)


endanke (approved):

Sources/MapboxMaps/Annotations/ViewAnnotationManager.swift
```diff
@@ -193,7 +193,7 @@ public final class ViewAnnotationManager {
         let isHidden = !(options.visible ?? true)
         expectedHiddenByView[view] = isHidden
         viewsById[id]?.isHidden = isHidden
-        if let id = currentFeatureId, id != options.associatedFeatureId {
+        if let id = currentFeatureId, let updatedId = options.associatedFeatureId, id != updatedId {
             viewsByFeatureIds[id] = nil
```
OdNairy, line 197: I do not quite understand this change. It looks like the same lines to me. Can you explain to me, please?

OdNairy: Okay, now I see – `options.associatedFeatureId` has to be non-nil, right?

OdNairy (approved):

Sources/MapboxMaps/Annotations/ViewAnnotationManager.swift
evil159: By unwrapping the `options.associatedFeatureId` this makes sure that a view loses its feature id association only when the feature id got updated, previously it would loose the association on every update.

Some sample code to illustrate this:
```swift 
let view = UIView()
let id = UUID().uuidString
let options = ViewAnnotationOptions(associatedFeatureId: id)

try manager.add(view, options: options)
try manager.update(view, options: ViewAnnotationOptions(width: 100, height: 100))

// previosly
manager.view(forFeatureId: id) // nil 💥

// fixed
manager.view(forFeatureId: id) // returns the view ✅
```

Sources/MapboxMaps/Annotations/ViewAnnotationManager.swift
```diff
@@ -193,7 +193,7 @@ public final class ViewAnnotationManager {
         let isHidden = !(options.visible ?? true)
         expectedHiddenByView[view] = isHidden
         viewsById[id]?.isHidden = isHidden
-        if let id = currentFeatureId, id != options.associatedFeatureId {
+        if let id = currentFeatureId, let updatedId = options.associatedFeatureId, id != updatedId {
```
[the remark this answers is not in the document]
OdNairy: It's also probably make sense to add an explicit comment above

PR #1468, opened 2022-07-15
[compass] expose compass image property

tobrun:
This PR allows to provide a custom image for the map compass. 

```swift
mapView.ornaments.options.compass.image = UIImage(named: "foobar")
```

https://user-images.githubusercontent.com/2151639/179181387-ecf81e2c-f91f-4e4a-b1cf-e0ec9a327502.mp4

## Pull request checklist:

### DIFF
--- a/Sources/MapboxMaps/Ornaments/Compass/MapboxCompassOrnamentView.swift
+++ b/Sources/MapboxMaps/Ornaments/Compass/MapboxCompassOrnamentView.swift
@@ -50,18 +50,24 @@ internal class MapboxCompassOrnamentView: UIButton {
                                               value: "Rotates the map to face due north",
                                               comment: "Accessibility hint")
 
-        if let image = createCompassImage() {
-            containerView.image = image
-            NSLayoutConstraint.activate([
-                widthAnchor.constraint(equalToConstant: image.size.width),
-                heightAnchor.constraint(equalToConstant: image.size.height),
-                containerView.widthAnchor.constraint(equalToConstant: image.size.width),
-                containerView.heightAnchor.constraint(equalToConstant: image.size.height)
-            ])
-        }
         containerView.translatesAutoresizingMaskIntoConstraints = false
+        if let image = createCompassImage() {
+            updateImage(image: image)
+        }
         addSubview(containerView)
         addTarget(self, action: #selector(didTap), for: .touchUpInside)
+    }
+
+    func updateImage(image: UIImage?) {
+        guard let image = image else { return }
+
+        containerView.image = image
+        NSLayoutConstraint.activate([
+            widthAnchor.constraint(equalToConstant: image.size.width),
+            heightAnchor.constraint(equalToConstant: image.size.height),
+            containerView.widthAnchor.constraint(equalToConstant: image.size.width),
+            containerView.heightAnchor.constraint(equalToConstant: image.size.height)
+        ])
     }
 
     required internal init?(coder: NSCoder) {
--- a/Sources/MapboxMaps/Ornaments/Compass/MapboxCompassOrnamentView.swift
+++ b/Sources/MapboxMaps/Ornaments/Compass/MapboxCompassOrnamentView.swift
@@ -9,6 +9,7 @@ internal class MapboxCompassOrnamentView: UIButton {
     }
 
     internal var containerView = UIImageView()
+    internal var containerViewConstraints = [NSLayoutConstraint]()
     internal var visibility: OrnamentVisibility = .adaptive {
         didSet {
             animateVisibilityUpdate()
@@ -60,14 +61,15 @@ internal class MapboxCompassOrnamentView: UIButton {
 
     func updateImage(image: UIImage?) {
         guard let image = image else { return }
-
+        NSLayoutConstraint.deactivate(containerViewConstraints)
         containerView.image = image
-        NSLayoutConstraint.activate([
+        containerViewConstraints = [
             widthAnchor.constraint(equalToConstant: image.size.width),
             heightAnchor.constraint(equalToConstant: image.size.height),
             containerView.widthAnchor.constraint(equalToConstant: image.size.width),
             containerView.heightAnchor.constraint(equalToConstant: image.size.height)
-        ])
+        ]
+        NSLayoutConstraint.activate(containerViewConstraints)
     }
 
     required internal init?(coder: NSCoder) {
--- a/Sources/MapboxMaps/Ornaments/OrnamentsManager.swift
+++ b/Sources/MapboxMaps/Ornaments/OrnamentsManager.swift
@@ -162,6 +162,9 @@ public class OrnamentsManager: NSObject {
                                                        margins: options.attributionButton.margins)
         constraints.append(contentsOf: attributionButtonConstraints)
 
+        // Update the image of compass
+        _compassView.updateImage(image: options.compass.image)
+
         // Activate new constraints
         NSLayoutConstraint.activate(constraints)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 Mapbox welcomes participation and contributions from everyone.
 
 ## main
+* Expose image property for compass ornament. ([#1466](https://github.com/mapbox/mapbox-maps-ios/pull/1466))
 
 * Expand scale bar range up to 15000 km/10000 miles. ([#1455](https://github.com/mapbox/mapbox-maps-ios/pull/1455))
 * Add the ability to override scale bar units. ([#1473](https://github.com/mapbox/mapbox-maps-ios/pull/1473))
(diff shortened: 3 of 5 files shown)


Sources/MapboxMaps/Ornaments/Compass/MapboxCompassOrnamentView.swift
```diff
@@ -51,19 +51,25 @@ internal class MapboxCompassOrnamentView: UIButton {
...
+    func updateImage(image: UIImage?) {
+        if image != nil {
+            containerView.image = image
+            NSLayoutConstraint.activate([
+                widthAnchor.constraint(equalToConstant: image!.size.width),
+                heightAnchor.constraint(equalToConstant: image!.size.height),
+                containerView.widthAnchor.constraint(equalToConstant: image!.size.width),
+                containerView.heightAnchor.constraint(equalToConstant: image!.size.height)
+            ])
+        }
```
OdNairy, line 70: It is better to use `guard` statement to unwrap optional values:
```suggestion
        guard let image = image else { return }

        containerView.image = image

        NSLayoutConstraint.activate([
            widthAnchor.constraint(equalToConstant: image.size.width),
            heightAnchor.constraint(equalToConstant: image.size.height),
            containerView.widthAnchor.constraint(equalToConstant: image.size.width),
            containerView.heightAnchor.constraint(equalToConstant: image.size.height)
        ])
```

Sources/MapboxMaps/Ornaments/OrnamentsManager.swift
```diff
@@ -162,6 +162,9 @@ public class OrnamentsManager: NSObject {
                                                        margins: options.attributionButton.margins)
         constraints.append(contentsOf: attributionButtonConstraints)
 
+        // Update the image of compass
+        _compassView.updateImage(image: options.compass.image)
```
OdNairy, line 166: Wouldn't we nil compass image here if we have no custom one? 🤔

Sources/MapboxMaps/Ornaments/Compass/MapboxCompassOrnamentView.swift
```diff
@@ -51,19 +51,25 @@ internal class MapboxCompassOrnamentView: UIButton {
...
-            ])
+            updateImage(image: image)
         }
         containerView.translatesAutoresizingMaskIntoConstraints = false
         addSubview(containerView)
         addTarget(self, action: #selector(didTap), for: .touchUpInside)
     }
 
+    func updateImage(image: UIImage?) {
+        if image != nil {
```
evil159, line 62:
```suggestion
        if let image = image {
```
This way you won't need to force unwrap the image below

Sources/MapboxMaps/Ornaments/Compass/MapboxCompassOrnamentView.swift
```diff
@@ -51,19 +51,25 @@ internal class MapboxCompassOrnamentView: UIButton {
...
+            containerView.image = image
+            NSLayoutConstraint.activate([
```
evil159, line 64: Old constraints needs to be deactivated, as there will be conflicting constraints if image size changed.

Alternatively, the container view can be constrained to the edges of the compass ornament view. You could store references to the width/height constraints of the compass ornament view and update them when updating the image.

```swift
widthConstraint.constant = image.size.width
heightConstraint.constant = image.size.height
```

Sources/MapboxMaps/Ornaments/OrnamentsManager.swift
tobrun: this updateImage function ignores nil images, in that case it would be a no-op. This is the part that I'm not sure about what is the most common/expected way for swift developers. 

Should I nil check here and not invoke updateImage + make updateImage arguments not optional.
or what it does now?

Sources/MapboxMaps/Ornaments/Compass/MapboxCompassOrnamentView.swift
```diff
@@ -51,19 +51,25 @@ internal class MapboxCompassOrnamentView: UIButton {
...
+                widthAnchor.constraint(equalToConstant: image!.size.width),
```
evil159, line 65: Just want to make sure that this is intentional - previously the compass view were of size 40x40 pt, now we are giving the compass size up to the discretion of developers, is this how it behaves on Android? An alternative could be to keep the compass view size constant and scale the image up/down.

evil159 (changes requested):

Sources/MapboxMaps/Ornaments/Compass/MapboxCompassOrnamentView.swift
```diff
@@ -51,19 +51,25 @@ internal class MapboxCompassOrnamentView: UIButton {
...
 
         if let image = createCompassImage() {
-            containerView.image = image
-            NSLayoutConstraint.activate([
-                widthAnchor.constraint(equalToConstant: image.size.width),
-                heightAnchor.constraint(equalToConstant: image.size.height),
-                containerView.widthAnchor.constraint(equalToConstant: image.size.width),
-                containerView.heightAnchor.constraint(equalToConstant: image.size.height)
-            ])
+            updateImage(image: image)
```
maios, line 54: I believe [line 56](https://github.com/mapbox/mapbox-maps-ios/pull/1468/files#diff-b7e6600deb16d0a592c242ab3b13cdeb779b00639a5348dba82ab6452ef18903R56) should be placed before this `if` statement because it applies autolayout to the `containerView`

Sources/MapboxMaps/Ornaments/Compass/MapboxCompassOrnamentView.swift
maios: interesting, I thought if we constraint the subview to the edges of its superview and set some compression/hugging priority then the superview will also resize to always fit the subview?

Sources/MapboxMaps/Ornaments/Compass/MapboxCompassOrnamentView.swift
```diff
@@ -51,19 +51,25 @@ internal class MapboxCompassOrnamentView: UIButton {
...
         }
         containerView.translatesAutoresizingMaskIntoConstraints = false
         addSubview(containerView)
         addTarget(self, action: #selector(didTap), for: .touchUpInside)
     }
 
+    func updateImage(image: UIImage?) {
```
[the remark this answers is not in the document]
maios: oh yeah sorry didn't realize you need this to be internal for the unit tests.

Sources/MapboxMaps/Ornaments/Compass/MapboxCompassOrnamentView.swift
tobrun: >  now we are giving the compass size up to the discretion of developers, is this how it behaves on Android?

yes, to give context on how it works on Android, we expose:
 - gravity: allows to position an element (top-right by default for compass)
 - margins: allows to offset the view from the edge

>  An alternative could be to keep the compass view size constant and scale the image up/down.

That would limit the freedom of the developer to build any kind of UX. 
If we properly update the constraints, I'm expecting it should work for any image size.

Sources/MapboxMaps/Ornaments/Compass/MapboxCompassOrnamentView.swift
tobrun: interresting! Since this code was already this way :)
will patch

Sources/MapboxMaps/Ornaments/Compass/MapboxCompassOrnamentView.swift
tobrun: was able to validate it doesn't throw any warnings anymore about constraints!

Sources/MapboxMaps/Ornaments/OrnamentsManager.swift
maios: If we first set a non-nil image for a compass, then after that we want to reset this by passing a nil compass image, this won't work.
```
// 1. This should use the custom compass image
mapView.ornaments.options.compass.image = anImage
// later
// 2. I would expect this to use default compass image.
mapView.ornaments.options.compass.image = nil
```

evil159: I'd also expect the compass image would be reverted to default after resetting it to nil. 

On Android resetting the compass image to `null` seems to be ignored as well. [CompassViewPlugin.kt#L70](https://github.com/mapbox/mapbox-maps-android/blob/f8a72f9f5b49b28d4880dba86b2c9603b423b82d/plugin-compass/src/main/java/com/mapbox/maps/plugin/compass/CompassViewPlugin.kt#L70)

evil159: I think I'm fine with this as long as both platforms are aligned.

evil159 (approved):

OdNairy (approved):

maios (approved):

PR #1473, opened 2022-07-18
Add the ability to override scale bar units

evil159:
This PR adds a public configuration option to allow overriding scale bar distance units. The default value is taken from the current locale.

#### New public API:
`ScaleBarViewOptions.useMetricUnits`

## Pull request checklist:

### DIFF
--- a/Sources/MapboxMaps/Ornaments/DistanceFormatter.swift
+++ b/Sources/MapboxMaps/Ornaments/DistanceFormatter.swift
@@ -11,20 +11,27 @@ internal class DistanceFormatter: MeasurementFormatter {
     ///
     /// - parameter distance: The distance, measured in meters.
     /// - returns: A localized formatted distance string including units.
-    internal func string(fromDistance distance: CLLocationDistance) -> String {
+    internal func string(fromDistance distance: CLLocationDistance, useMetricSystem: Bool? = nil) -> String {
 
         numberFormatter.roundingIncrement = 0.25
 
         var measurement = Measurement(value: distance, unit: UnitLength.meters)
 
-        if !locale.usesMetricSystem {
+        let shouldUseMetricSystem: Bool
+        if let useMetricSystem = useMetricSystem {
+            shouldUseMetricSystem = useMetricSystem
+        } else {
+            shouldUseMetricSystem = locale.usesMetricSystem
+        }
+
+        if !shouldUseMetricSystem {
             unitOptions = .providedUnit
             measurement.convert(to: .miles)
             if measurement.value <= 0.2 {
                 measurement.convert(to: .feet)
             }
         } else {
-            unitOptions = .naturalScale
+            unitOptions = [.providedUnit, .naturalScale]
         }
         return string(from: measurement)
     }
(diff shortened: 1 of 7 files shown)


Sources/MapboxMaps/Ornaments/DistanceFormatter.swift
```diff
@@ -11,20 +11,27 @@ internal class DistanceFormatter: MeasurementFormatter {
...
 
         var measurement = Measurement(value: distance, unit: UnitLength.meters)
 
-        if !locale.usesMetricSystem {
+        let shouldUseMetricSystem: Bool
+        if let useMetricSystem = useMetricSystem {
+            shouldUseMetricSystem = useMetricSystem
+        } else {
+            shouldUseMetricSystem = locale.usesMetricSystem
+        }
```
OdNairy, line 25: Can be simplifed:
```suggestion
        let shouldUseMetricSystem = useMetricSystem ?? locale.usesMetricSystem
```

Sources/MapboxMaps/Ornaments/DistanceFormatter.swift
```diff
@@ -11,20 +11,27 @@ internal class DistanceFormatter: MeasurementFormatter {
...
+
+        if !shouldUseMetricSystem {
```
OdNairy, line 27: Nit: My personal preference but can we revert the logic here? It's a bit easier to read the `if statement` than `if NOT statement` for me.
What do you think?

Sources/MapboxMaps/Ornaments/OrnamentOptions.swift
```diff
@@ -50,6 +50,9 @@ public struct ScaleBarViewOptions: OrnamentOptionsProtocol, Equatable {
     public var margins: CGPoint = defaultOrnamentsMargin
     /// The default value for this property is `.adaptive`.
     public var visibility: OrnamentVisibility = .adaptive
+    /// Specifies the whether the scale bar uses the metric system.
+    /// True if the scale bar is using metric units, false if the scale bar is using imperial units.
+    public var useMetricUnits: Bool = Locale.current.usesMetricSystem
```
OdNairy, line 55: Discussion: While the Bool flag works fine here and even Apple uses the same Bool in `Locale`, I would like to suggest using `enum` with concrete cases. It would increase readability as well as make some room for extensions and, potential and theoretical, support of our measurement systems (if there are any exist).

OdNairy: For example, we have a pattern of access to a table of constants like `useMetricUnits ? Constants.metricTable : Constants.imperialTable`. 
We could make a function/variable in custom enum to make access even easier.

Sources/MapboxMaps/Ornaments/DistanceFormatter.swift
```diff
@@ -11,20 +11,27 @@ internal class DistanceFormatter: MeasurementFormatter {
     ///
     /// - parameter distance: The distance, measured in meters.
     /// - returns: A localized formatted distance string including units.
-    internal func string(fromDistance distance: CLLocationDistance) -> String {
+    internal func string(fromDistance distance: CLLocationDistance, useMetricSystem: Bool? = nil) -> String {
```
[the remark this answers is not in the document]
evil159: I made it optional here to be able to indicate that locale's measurement system should be used. 
Tried this but it doesn't work:
`internal func string(fromDistance distance: CLLocationDistance, useMetricSystem: Bool = locale.usesMetricSystem) -> String
`

OdNairy (approved):

PR #1476, opened 2022-07-19
MAPSIOS-118 Implement DictionaryEncoder for Layer properties.

maios:
- Encode Layer properties directly to a dictionary, instead of encoding to data using JSONEncoder and then serializing the data to a JSON dictionary.
- Support encoding null values based on encoding context.
- When updating a layer, merge the old properties to new properties, if a property is reset (old is non-nil and new is nil) then we will use the default value.

**NOTE**:  This will depend on the upstream PR https://github.com/mapbox/mapbox-gl-native-internal/pull/3736 in order to support passing null `StyleTransition`

## Pull request checklist:


Sources/MapboxMaps/Foundation/DictionaryCoding/DictionaryEncoder.swift
```diff
@@ -0,0 +1,399 @@
...
+            if DictionaryEncoder.isSupportedType(T.self) {
+                storage[key.stringValue] = .value(value)
+            } else {
+                let result = try Encoder(codingPath: codingPath.appending(key: key), userInfo: userInfo).encode(value)
+                storage[key.stringValue] = .value(result)
+            }
+        }
+
+        func encodeIfPresent<T: Encodable>(_ value: T?, forKey key: Key) throws {
+            let shouldEncodeNilValue = userInfo[.shouldEncodeNilValues] as? Bool ?? false
```
evil159, line 176: With `shouldEncodeNilValues` baked into the source it'll be more clear to expose it as a property on `DictionaryEncoder` rather than user info key.
Primary use case for `userInfo` is for the user of the API to pass some contextual info down the line.

Sources/MapboxMaps/Foundation/DictionaryCoding/DictionaryEncoder.swift
```diff
@@ -0,0 +1,399 @@
...
+        func encode<T: Encodable>(_ value: T, forKey key: Key) throws {
+            if DictionaryEncoder.isSupportedType(T.self) {
+                storage[key.stringValue] = .value(value)
+            } else {
+                let result = try Encoder(codingPath: codingPath.appending(key: key), userInfo: userInfo).encode(value)
+                storage[key.stringValue] = .value(result)
+            }
+        }
+
+        func encodeIfPresent<T: Encodable>(_ value: T?, forKey key: Key) throws {
```
evil159, line 175: It seems that something is wrong with this - this method does not seem to be called for primitive types(like Double - e.g. minzoom is not encoded).

Tests/MapboxMapsTests/Foundation/DictionaryEncoderTests.swift
```diff
@@ -0,0 +1,436 @@
...
+            [:]
+        )
+
+        // Int
+        XCTAssertEqual(
+            try sut.encode(DummyNilable(value: 10)) as? [String: Int?],
+            ["value": 10]
+        )
+        XCTAssertEqual(
+            try sut.encode(DummyNilable<Int>(value: nil)) as? [String: Int?],
```
evil159, line 330: There's some funky stuff going on - this works with a generic struct. But when using a similar specialized struct - the result is an empty dictionary.

```swift
    private struct DummyFloatEncodable: Encodable {
        let value: Float?
    }
```

Tests/MapboxMapsTests/Foundation/StyleTests.swift
```diff
@@ -344,4 +344,44 @@ final class StyleTests: XCTestCase {
             bounds: CoordinateBounds(southwest: .random(), northeast: .random()))
         )
     }
+
+    func testStyleCanUpdateLayer() throws {
```
evil159, line 348: Add tests that check that:
* nilling out unmodified properties works correctly
* updating a property value that already has a value works correctly

Tests/MapboxMapsTests/Foundation/DictionaryEncoderTests.swift
evil159: Maybe a sure way to test these would be to define a dedicated struct for every test 🤔

Sources/MapboxMaps/Foundation/DictionaryCoding/DictionaryEncoder.swift
```diff
@@ -0,0 +1,399 @@
...
+    enum Error: Swift.Error {
+        case unexpectedType
+        case incomplete(at: [CodingKey])
+    }
+
+    private static func isSupportedType<T: Encodable>(_ type: T.Type) -> Bool {
+        T.self == Data.self || T.self == NSData.self ||
+        T.self == Date.self || T.self == NSDate.self ||
+        T.self == Decimal.self || T.self == NSDecimalNumber.self ||
+        T.self == URL.self || T.self == NSURL.self
```
evil159, line 28: What does it mean that only these types are supported?

evil159 (changes requested):
Very interesting to see a custom implementation of Swift's encoders and decoders 👍

Tests/MapboxMapsTests/Foundation/StyleTests.swift
```diff
@@ -344,4 +344,44 @@ final class StyleTests: XCTestCase {
...
+        XCTAssertTrue(rootProperties.keys.contains("minzoom"), "minzoom is reset and should be presented")
+        XCTAssertTrue(rootProperties.keys.contains("maxzoom"), "maxzoom is reset and should be presented")
+
+        let paintProperties = try XCTUnwrap(rootProperties["paint"] as? [String: Any])
+        XCTAssertTrue(paintProperties.keys.contains("background-opacity-transition"), "background-opacity-transition is reset and should be presented")
+        XCTAssertFalse(paintProperties.keys.contains("background-opacity"), "background-opacity is not reset and should not be presented")
+
+        let layoutProperties = try XCTUnwrap(rootProperties["layout"] as? [String: Any])
+        XCTAssertEqual(layoutProperties["visibility"] as? String, "visible", "visibility is not reset and should keep old value")
+    }
```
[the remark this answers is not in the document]
maios: it does not work with the test setup like this, because we would need to get a dictionary of layer's properties then decode it to a Layer, so we could only verify that when we call `Style.updateLayer` it will invoke its `StyleManager` with the expected parameters, which is a Dictionary of merged old properties and new properties.

Tests/MapboxMapsTests/Foundation/DictionaryEncoderTests.swift
maios: yes, I am also thinking of using stylegen to generate the tests case for each Layer's struct :-D let's see!

Tests/MapboxMapsTests/Foundation/StyleTests.swift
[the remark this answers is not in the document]
evil159: Can we have an integration tests for this?

Sources/MapboxMaps/Foundation/DictionaryCoding/DictionaryEncoder.swift
maios: aye, the function name should be changed into something that makes more sense, but this is to check for the type and encode them properly (maybe later will be based on configuration i.e data will be encoded with base64, date will be encoded with ISO8601 etc)
I will rename the function.

Sources/MapboxMaps/Foundation/DictionaryCoding/DictionaryEncoder.swift
maios: good point, this is done now!

Sources/MapboxMaps/Foundation/DictionaryCoding/DictionaryEncoder.swift
maios: my bad, I didn't commit `encodeIfPresent` for primitive types :-D

Tests/MapboxMapsTests/Foundation/DictionaryEncoderTests.swift
maios: This test is expected the dictionary to be empty though? Did you mean to comment this in other places?

Tests/MapboxMapsTests/Foundation/StyleTests.swift
maios: done!

Sources/MapboxMaps/Foundation/DictionaryCoding/DictionaryEncoder.swift
```diff
@@ -164,19 +202,73 @@ private extension DictionaryEncoder {
...
-            let shouldEncodeNilValue = userInfo[.shouldEncodeNilValues] as? Bool ?? false
-            if value != nil || shouldEncodeNilValue {
-                try encode(value, forKey: key)
-            }
+            guard value != nil || shouldEncodeNilValues else { return }
+            try encode(value, forKey: key)
+        }
+
+        func encodeIfPresent(_ value: Bool?, forKey key: Key) throws {
+            try encodeIfPresent(NilEncodable(value), forKey: key)
```
evil159, line 219: What does wrapping the value into `NilEncodable` do here? For me it seems that it is the same as passing `value` directly, if the value is nil - nothing will be encoded, non-nil - it will be encoded.

maios: because these are overloading methods with same signature, so I wrap them in a struct so it will trigger `func encodeIfPresent<T: Encodable>(_ value: T, forKey key: Key)` instead, otherwise I would have to do the check for `shouldEncodeNilValues` in all the methods of `encodeIfPresent`
Another approach would be having a private method
```
private func encodeNilIfNeeded<T: Encodable>(_ value: T?, forKey key: Key) throws {
  guard value != nil || shouldEncodeNilValues else { return }
  try encode(value, forKey: key)
}
```
then have all the `encodeIfPresent` to point to this one, what do you think?

evil159 (approved):
Well done 👍

PR #1485, opened 2022-07-27
Fix latitude being NaN during flyTo

evil159:
In some specific cases when using paddings close to the screen size and high zoom flyTo could produce a native exception:

`latitude must not be NaN`

Fix this issue by treating infinite values in algorithm properly.

## Pull request checklist:


maios (approved):
can the values in test be random?

kiryldz (approved):

evil159:
> can the values in test be random?

This test checks for a specific case where the composition of source, destination camera states, camera bounds, as well as map size produce the value of total length of the camera flight path(`FlyToInterPolator.S`) to be not a number.

PR #1486, opened 2022-07-27
Remove experimental model layer and shadows API

evil159:
This PR removes docs from shadow-related APIs and removes shadows from examples


Sources/MapboxMaps/Style/Generated/Layers/ModelLayer.swift
```diff
@@ -19,7 +19,7 @@ import Foundation
     /// Model to render.
     public var modelId: Value<String>?
 
-    /// Enable/Disable shadow casting for this layer
+    ///
```
OdNairy, line 22: Should we left a note "Undocumented" or similar here?

OdNairy (approved):

Sources/MapboxMaps/Style/Generated/Enums/Enums.swift
```diff
@@ -403,19 +403,6 @@ public enum SkyType: String, Codable, CaseIterable {
...
-/// Defines rendering behavior of model in respect to other 3D scene objects.
-public enum ModelType: String, Codable, CaseIterable {
-
-    /// Integrated to 3D scene, using depth testing, along with terrain, fill-extrusions and custom layer.
-    case common3d = "common-3d"
-
-    /// Displayed over other 3D content, occluded by terrain.
-    case locationIndicator = "location-indicator"
-
-}
```
OdNairy, line 417: This change might be a breaking change :(

PR #1500, opened 2022-07-29
Update `OrnamentsManager` to dynamically respect language direction.

ZiZasaurus:
- Changed `leftAnchor` and `rightAnchor` to `leadingAnchor` and `trailingAnchor` so they would not be ignored by RTL configuration. 
- Updated `constraints(with: positions: margins:)` method to take device language direction into account when placing `scaleBar`, `logoView`, `attributionButton` and `compassView` ornaments. 

- Will need to add tests. 

| Before | After |
| --- | --- |
|Simulator Screen Shot - iPhone 12 - 2022-07-29 at 23 18 22| Simulator Screen Shot - iPhone 12 - 2022-08-01 at 12 45 02|

## Pull request checklist:

### DIFF
--- a/Sources/MapboxMaps/Ornaments/OrnamentsManager.swift
+++ b/Sources/MapboxMaps/Ornaments/OrnamentsManager.swift
@@ -186,7 +186,6 @@ public class OrnamentsManager: NSObject {
     }
 
     private func constraints(with view: UIView, position: OrnamentPosition, margins: CGPoint) -> [NSLayoutConstraint] {
-        
         let layoutGuide = view.superview!.safeAreaLayoutGuide
         switch position {
         case .topLeft:
--- a/Sources/MapboxMaps/Ornaments/OrnamentsManager.swift
+++ b/Sources/MapboxMaps/Ornaments/OrnamentsManager.swift
@@ -2,10 +2,19 @@ import UIKit
 
 public enum OrnamentPosition: String, Equatable {
     // Clockwise from top left
+    @available(*, deprecated, renamed: "topLeading")
     case topLeft
+    @available(*, deprecated, renamed: "topTrailing")
     case topRight
+    @available(*, deprecated, renamed: "bottomLeading")
     case bottomRight
+    @available(*, deprecated, renamed: "bottomTrailing")
     case bottomLeft
+
+    case topLeading
+    case topTrailing
+    case bottomLeading
+    case bottomTrailing
 }
 
 public enum OrnamentVisibility: String, Equatable {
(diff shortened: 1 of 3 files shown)


Sources/MapboxMaps/Ornaments/OrnamentsManager.swift
```diff
@@ -177,23 +177,33 @@ public class OrnamentsManager: NSObject {
...
+        // handle AutoLayout changes when device character direction changes
+        let currentDeviceLanguage: String! = Locale.current.languageCode
+        let rightToLeft = Locale.LanguageDirection.rightToLeft
+
+        if Locale.characterDirection(forLanguage: currentDeviceLanguage) == rightToLeft {
+            UIView.appearance().semanticContentAttribute = .forceRightToLeft
+            //0 is on the left side
+            _scaleBarView.semanticContentAttribute = .spatial
+        }
+
```
evil159, line 189: I think this can be skipped if `MapboxScaleBarOrnamentView` would use directional anchors(`leading`, `trailing`) and `UITraitCollection.layoutDirection` for checking the direction of layout.
E.g.:
```swift
    private func usesRightToLeftLayout() -> Bool {
        return traitCollection.layoutDirection == .rightToLeft
    }
```

evil159 (changes requested):

Sources/MapboxMaps/Ornaments/OrnamentsManager.swift
```diff
@@ -177,23 +177,33 @@ public class OrnamentsManager: NSObject {
     }
 
     private func constraints(with view: UIView, position: OrnamentPosition, margins: CGPoint) -> [NSLayoutConstraint] {
+        // handle AutoLayout changes when device character direction changes
+        let currentDeviceLanguage: String! = Locale.current.languageCode
+        let rightToLeft = Locale.LanguageDirection.rightToLeft
+
+        if Locale.characterDirection(forLanguage: currentDeviceLanguage) == rightToLeft {
+            UIView.appearance().semanticContentAttribute = .forceRightToLeft
```
maios, line 185: I don't think we should force `semanticContentAttribute` to appearance proxy as it affects other views too
```suggestion
            view.superview?.semanticContentAttribute = .forceRightToLeft
```

maios (changes requested):

Sources/MapboxMaps/Ornaments/OrnamentsManager.swift
```diff
@@ -177,23 +177,33 @@ public class OrnamentsManager: NSObject {
...
+            //0 is on the left side
+            _scaleBarView.semanticContentAttribute = .spatial
```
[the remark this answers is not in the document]
ZiZasaurus: I think that's what it would have had to be but i worked with @evil159 to change it so we don't even need to rely on this characterDirection check.

Sources/MapboxMaps/Ornaments/ScaleBar/MapboxScaleBarOrnamentView.swift
```diff
@@ -346,7 +346,7 @@ internal class MapboxScaleBarOrnamentView: UIView {
     // MARK: - Convenience Methods
 
     private func usesRightToLeftLayout() -> Bool {
-        return UIView.userInterfaceLayoutDirection(for: semanticContentAttribute) == .rightToLeft
```
maios, line 349: why did we change this? imo the former is better since we are exposing scale bar view under `OrnamentsManager`, so the client code can always enforce `semanticContentAttributed` on this view.
```
mapView.ornaments.scaleBarView.semanticContentAttribute = .forceLeftToRight
```
when this happens, the former will return false, but the new one will return true when the trait environment is in RTL locale. WDYT?

evil159: It was my suggestion, using `traitCollection` will allow the scale bar to respect trait collection being overridden upstream. After looking more closely, I found out that neither of two is the recommended way to check layout direction.
`UIView.effectiveUserInterfaceLayoutDirection` is. Probably should switch to using this instead.

evil159 (approved):

maios (approved):
🎉

Sources/MapboxMaps/Ornaments/OrnamentsManager.swift
```diff
@@ -2,10 +2,19 @@ import UIKit
 
 public enum OrnamentPosition: String, Equatable {
     // Clockwise from top left
+    @available(*, deprecated, renamed: "topLeading")
     case topLeft
+    @available(*, deprecated, renamed: "topTrailing")
     case topRight
+    @available(*, deprecated, renamed: "bottomLeading")
     case bottomRight
+    @available(*, deprecated, renamed: "bottomTrailing")
```
MaximAlien, line 11: @ZiZasaurus, we've noticed this change during latest Maps SDK integration. Just want to make sure that this is correct naming. Shouldn't it be `bottomLeading` for `bottomLeft` and `bottomTrailing` for `bottomRight`, similarly to top positions?

azarovalex: Yeah, can confirm it's a bug, bottomTrailing is actually shows on the left side of the screen😂

azarovalex: Imho we shouldn't deprecate left/right options, UIKit has both leftAnchor and leadingAnchor, not all people change the layout of the whole app to support RTL, so we should support all options.

ZiZasaurus: Thanks for this! Pushing this change for the next release but for now we will be keeping the deprecation as is.

OdNairy: @mapbox/maps-ios Should we revert it then?

evil159: Seems to be fixed in here https://github.com/mapbox/mapbox-maps-ios/pull/1563

OdNairy: @evil159 Sorry, forgot to mention this one:

> Imho we shouldn't deprecate left/right options, UIKit has both leftAnchor and leadingAnchor, not all people change the layout of the whole app to support RTL, so we should support all options.

pjleonard37: 👍  -- I like having both naming options

PR #1501, opened 2022-07-30
Animated wind custom layer on globe/terrain - custom layer draping demo

astojilj:
The Demo uses extended CustomLayerHost interface, the API that is planned to be released (so the API presented below is work in progress) with Mapbox Mobile SDK 10.8, both on iOS and Android. It is based on PoC https://github.com/mapbox/mapbox-gl-js/pull/11996.

Demo on iPad 11 Pro (2018):

https://user-images.githubusercontent.com/549216/181916446-8fddf871-20d1-4181-a6cc-9f3bb0e602e4.mov

Every 100ms (10 FPS video playback can be changed, it is a solution for the demo) a frame from video is uploaded to texture and rendered to tiles (draped)  of globe (works for terrain too).

TODO:
- [ ] timer to issue map re-render  since currently it is re-rendered on user interaction.

There is no API break nor change in behavior of existing custom layer APIs.

Draping = rendering to tile textures following style order of layer. Mapbox engine renders line, fill, background, raster, hillshade, some symbol layers, circle in future).

```
NS_SWIFT_NAME(CustomLayerHost)
@protocol MBMCustomLayerHost <NSObject>

/**
 * This method is called once per frame, before any other layer is rendered. It is to be used to
 * allow a layer to prepare resources or render into a texture and for layer to report back to
 * engine if draping (orthographic mode rendering to tile textures for globe and terrain) is supported.
 * The implementation should not make
 * any assumptions about the GL state (other than that the correct context is active). It may
 * make changes to the state, and is not required to reset values such as the depth mask, stencil
 * mask, and corresponding test flags to their original values.
 *
 * @param parameters The `custom layer render parameters`.
 *
 * @return CustomLayerRenderConfiguration that defines if e.g. the layer renders to tiles textures for globe or terrain.
 *
 */
- (nonnull MBMCustomLayerRenderConfiguration *)prerender:(nonnull MBMCustomLayerRenderParameters *)parameters
mtlCommandBuffer:(nonnull id<MTLCommandBuffer>)mtlCommandBuffer;

/**
 * Orthographic mode render of the the layer to the tile. This method is called once per tile
 * if globe or terrain are enabled and if there's need to re-render the tile texture. Tile textures
 * are cached and the need to re-render is evaluated based on data or styled properties change.
 * This method is called only if client code returns CustomLayerRenderParameters's isRenderToTileSupported
 * from prerender() method.
 * Implementation must not call endEncoding to provided mtlRenderCommandEncoder.
 * If there's a need for e.g. rendering to texture, implementation should use prerender() method.
 *
 * @param tileId Tile ID of terrain/globe tile that we are rendering to.
 *
 */
- (void)renderToTile:(nonnull MBMCanonicalTileID *)tileID
mtlRenderCommandEncoder:(nonnull id<MTLRenderCommandEncoder>)mtlRenderCommandEncoder;
```

Prerender method exists in gl-js and it is an important one for performance - gl-native clients are, for e.g. uploading and  rendering to texture, forced to switch from main render pass, bind other tergets or upload, and then switch back to main render pass. This introduces the same method and adds return struct so that implementatio can declare what's it doing.

Swift version (implemented in GlobeCustomLayerHost demo here:

func prerender(_ parameters: CustomLayerRenderParameters, mtlCommandBuffer: MTLCommandBuffer) -> CustomLayerRenderConfiguration
func render(toTile tileID: CanonicalTileID, mtlRenderCommandEncoder: MTLRenderCommandEncoder)

```
/**
 * CustomLayerHost declares, by returning this from CustomLayerHost's prerender, what kind of rendering it intends to do in the very next frame.
 * If there is need for rendering to tile textures, declare if all tiles need to be re-rendered because e.g. underlying data changed.
 */
NS_SWIFT_NAME(CustomLayerRenderConfiguration)
__attribute__((visibility ("default")))
@interface MBMCustomLayerRenderConfiguration : NSObject

- (nonnull instancetype)init;

- (nonnull instancetype)initWithIsRenderToTileSupported:(BOOL)isRenderToTileSupported
                                    shouldRerenderTiles:(BOOL)shouldRerenderTiles;

/**
 * If this value is true and there's need for draping (globe or terrain are enabled), then renderToTile will be called to enable
 * the rendering of the custom layer into the texture of a tile.
 * If this value is set, and terrain or globe are active, only render to tile render call for this layer is called,
 * and "immediate mode" render is not called. A custom layer is, at one frame, either rendered to tile or rendered
 * directly.
 */
@property (nonatomic, readonly, getter=isIsRenderToTileSupported) BOOL isRenderToTileSupported;

/**
 * Set to true only for frame when content has changed - when set to true,
 * all the terrain render cache would get invalidated and redrawn, causing a drop in performance if invoked on every frame.
 */
@property (nonatomic, readonly, getter=isShouldRerenderTiles) BOOL shouldRerenderTiles;
```

If this is not empty in prerender(), implementation is informed that there is need to render to tiles, either because globe or terrain (or both like in demo) are enabled. in such case, implementation decides to continue render in immediate mode (return false for isRenderToTileSupported::isRenderToTileSupported) and in this case renderToTile methods are not invoked by engine, or to render to tiles, by returning true for isRenderToTileSupported::isRenderToTileSupported.

If application needs to render both draped and immediate mode in the same time, two layers need to be added and one of the layers would skip rendering based on presence of CustomLayerRenderParameters::renderToTilesIDs (optional, supplied on terrain or globe both to prerender() and render() calls). If application needs to know if draping happens on globe or terrain, application would for now need to deduct it based on zoom level (globe -> mercator transition happens in zoom range [5..6] although it is expected to change soon, or get current projection value from Style) and if globe is enabled by application. Immediate mode custom layer API is under work and will further clarify this.

noaa-gmgsi-2k.mp4 globe wind video used in the demo is done by @rreusser from: https://registry.opendata.aws/noaa-gmgsi/

> License
Open Data. There are no restrictions on the use of this data.

## Pull request checklist:

### DIFF
--- a/Apps/DebugApp/DebugApp.xcodeproj/project.pbxproj
+++ b/Apps/DebugApp/DebugApp.xcodeproj/project.pbxproj
@@ -11,6 +11,7 @@
 		35F08DE02347987700768B9F /* DebugViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 35F08DDF2347987700768B9F /* DebugViewController.swift */; };
 		35F08DE52347987900768B9F /* Assets.xcassets in Resources */ = {isa = PBXBuildFile; fileRef = 35F08DE42347987900768B9F /* Assets.xcassets */; };
 		35F08DE82347987900768B9F /* LaunchScreen.storyboard in Resources */ = {isa = PBXBuildFile; fileRef = 35F08DE62347987900768B9F /* LaunchScreen.storyboard */; };
+		7473C3BA288995B40022279E /* DebugViewController.xib in Resources */ = {isa = PBXBuildFile; fileRef = 7473C3B9288995B40022279E /* DebugViewController.xib */; };
 		B569C20125FEBC2E008B0D92 /* MapboxMaps in Frameworks */ = {isa = PBXBuildFile; productRef = B569C20025FEBC2E008B0D92 /* MapboxMaps */; };
 /* End PBXBuildFile section */
 
@@ -34,6 +35,7 @@
 		35F08DE42347987900768B9F /* Assets.xcassets */ = {isa = PBXFileReference; lastKnownFileType = folder.assetcatalog; path = Assets.xcassets; sourceTree = "<group>"; };
 		35F08DE72347987900768B9F /* Base */ = {isa = PBXFileReference; lastKnownFileType = file.storyboard; name = Base; path = Base.lproj/LaunchScreen.storyboard; sourceTree = "<group>"; };
 		35F08DE92347987900768B9F /* Info.plist */ = {isa = PBXFileReference; lastKnownFileType = text.plist.xml; path = Info.plist; sourceTree = "<group>"; };
+		7473C3B9288995B40022279E /* DebugViewController.xib */ = {isa = PBXFileReference; lastKnownFileType = file.xib; path = DebugViewController.xib; sourceTree = "<group>"; };
 		CA56651F24479764006F392C /* base.xcconfig */ = {isa = PBXFileReference; lastKnownFileType = text.xcconfig; path = base.xcconfig; sourceTree = "<group>"; };
 /* End PBXFileReference section */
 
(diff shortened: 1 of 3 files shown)


astojilj:
@evil159, I've copied the gl-native Objective C demo here and quickly ported it to Swift. Thanks for support so far and on help in improving the swift code further.

evil159 (commented):
Thanks for adding a very nice example!

I wonder if `CADisplayLink` can be used to drive rendering of the layer continuously 🤔

PR #1509, opened 2022-08-03
Add support for heading calibration alert

evil159:
This PR adds a method to location manager's delegate protocol to provide the ability to display the heading calibration alert on top of the current window immediately.

## Public API changes
* `LocationPermissionsDelegate` deprecated
* `LocationPermissionsDelegate` renamed to `LocationManagerDelegate`
* `LocationManagerDelegate.locationManagerShouldDisplayHeadingCalibration(_:) -> Bool` added

## Pull request checklist:


evil159:
- [ ] Test this by making the heading calibration alert to appear

Sources/MapboxMaps/Location/LocationManager.swift
```diff
@@ -11,7 +11,7 @@ public final class LocationManager: NSObject {
     }
 
     /// The object that acts as the delegate of the location manager.
-    public weak var delegate: LocationPermissionsDelegate?
+    public weak var delegate: LocationManagerDelegate?
```
OdNairy, line 14: This is a breaking change.
I suppose your workaround with `LocationPermissionsDelegate: LocationManagerDelegate` is still a breaking change because the delegate property could be tested for being the concrete type `is LocationPermissionsDelegate`. Unfortunately, we have to keep the `LocationPermissionsDelegate` type here and cut the renaming ticket for the next release.

OdNairy (changes requested):

Sources/MapboxMaps/Location/LocationManager.swift
evil159: Indeed, this is technically a breaking change. Practically I figure this should be fine as this is non-source breaking change. Behavior-wise it should be work as well - if somebody is going to check if `delegate is LocationPermissionsDelegate` then they set it previously to `LocationPermissionsDelegate`(since they still use the old API).

Sources/MapboxMaps/Location/LocationProviderDelegate.swift
```diff
@@ -40,3 +40,7 @@ internal class EmptyLocationProviderDelegate: LocationProviderDelegate {
     func locationProvider(_ provider: LocationProvider, didUpdateLocations locations: [CLLocation]) {}
     func locationProviderDidChangeAuthorization(_ provider: LocationProvider) {}
 }
+
+internal protocol CalibratingLocationProviderDelegate: LocationProviderDelegate {
+    func locationProviderShouldDisplayHeadingCalibration(_ locationProvider: LocationProvider) -> Bool
```
maios, line 45: as discussed, we should have this as part of public `LocationProviderDelegate` so custom `LocationProvider` can have one possible channel of communicating back to LocationManager to consult whether or not should calibration alert be displayed.

Sources/MapboxMaps/Location/LocationManager.swift
maios: image
I agree with @OdNairy on this and cut the renaming for next major release. Even though it is silly to check against concrete `LocationPermissionsDelegate`, there is still a possibility and thus this change will be breaking changes for them.

Sources/MapboxMaps/Location/LocationProviderDelegate.swift
evil159: Adding another method to a public protocol would be a breaking change and we need to verify that methods with default implementation are dispatched correctly before using those.

maios (approved):

PR #1512, opened 2022-08-04
Support annotation view lookup by id.

maios:
- Allow adding annotation view with a custom id, that can be used later to look it up.

## Pull request checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Mapbox welcomes participation and contributions from everyone.
 * Fix NaN latitude crash rarely happening in `CameraAnimationsManager.fly(to:duration:completion)`. ([#1485](https://github.com/mapbox/mapbox-maps-ios/pull/1485))
 * Fix `Style.updateLayer(withId:type:update)` so resetting a layer's properties should work. ([#1476](https://github.com/mapbox/mapbox-maps-ios/pull/1476))
 * Add support for sonar-like pulsing animation around 2D puck. ([#1513](https://github.com/mapbox/mapbox-maps-ios/pull/1513))
+* Support view annotation lookup by an identifier. ([#1512](https://github.com/mapbox/mapbox-maps-ios/pull/1512))
 
 ## 10.7.0 - July 28, 2022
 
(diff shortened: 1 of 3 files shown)


evil159 (commented):
I wonder if we should add a convenience method to update(remove to a lesser extent as well) a view annotation with id now when we allow passing custom id when adding a view annotation.
This change is requested in the context of making it easier to change a view annotation, so this addition, while being very cheap to us, might make this use case a bit more convenient.
E.g.:
```swift
// it is possible to get a view and update options for it currently
if let view = view(forId: id) {
    manager.update(view, options)
}

// with a convenience method this could be a single invocation
manager.update(forId: id, options)

// the convenience method
public func update(forId id: String, options: ViewAnnotationOptions) throws {
    guard let view = view(forId: id) else {
          throw ViewAnnotationManagerError.annotationNotFound
    }

    update(view, options: options)
}
```

maios:
> I wonder if we should add a convenience method to update(remove to a lesser extent as well) a view annotation with id now when we allow passing custom id when adding a view annotation. This change is requested in the context of making it easier to change a view annotation, so this addition, while being very cheap to us, might make this use case a bit more convenient. E.g.:
> 
> ```swift
> // it is possible to get a view and update options for it currently
> if let view = view(forId: id) {
>     manager.update(view, options)
> }
> 
> // with a convenience method this could be a single invocation
> manager.update(forId: id, options)
> 
> // the convenience method
> public func update(forId id: String, options: ViewAnnotationOptions) throws {
>     guard let view = view(forId: id) else {
>           throw ViewAnnotationManagerError.annotationNotFound
>     }
> 
>     update(view, options: options)
> }
> ```

I wanted to do the same, but I want to clarify some design intention behind the current `ViewAnnotationsManager/update(_:options:)` to accept new `options` with same `associatedFeatureId` (currently we don't accept it, so new `associatedFeatureId` has to be either nil or different value), after this we might want to change the current implementation so I want to cut it to another ticket.

evil159 (approved):

MICHELLGRONER88 (commented):
#581 @sandeep-mapbox

MICHELLGRONER88 (commented):
[snapcodes.zip](https://github.com/mapbox/mapbox-maps-ios/files/9330195/snapcodes.zip)

PR #1544, opened 2022-08-24
MapView metrics reporter

OdNairy:
## Pull request checklist:
    
### DIFF
--- a/Sources/MapboxMaps/Foundation/MapView.swift
+++ b/Sources/MapboxMaps/Foundation/MapView.swift
@@ -559,7 +559,11 @@ open class MapView: UIView {
         mapboxMap.size = bounds.size
     }
 
+    @_spi(Metrics) public var metricsReporter: MapViewMetricsReporter?
+
     private func updateFromDisplayLink(displayLink: CADisplayLink) {
+        metricsReporter?.beforeDisplayLinkCallback(displayLink: displayLink)
+        defer { metricsReporter?.afterDisplayLinkCallback(displayLink: displayLink) }
         if window == nil {
             return
         }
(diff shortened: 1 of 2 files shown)


Sources/MapboxMaps/Foundation/MapViewMetricsReporter.swift
```diff
@@ -0,0 +1,10 @@
+import Foundation
+import MetalKit
+
+@_spi(Metrics) public protocol MapViewMetricsReporter {
```
evil159, line 4:
```suggestion
@_spi(Metrics) public protocol MapViewMetricsReporter: AnyObject {
```
Performance 🚀

Sources/MapboxMaps/Foundation/MapView.swift
```diff
@@ -574,7 +578,9 @@ open class MapView: UIView {
 
         if needsDisplayRefresh {
             needsDisplayRefresh = false
+            metricsReporter?.beforeMetalViewDrawCallback(metalView: metalView)
             metalView?.draw()
+            metricsReporter?.afterMetalViewDrawCallback(metalView: metalView)
```
evil159, line 583: I wonder if it would make sense to use compilation conditions for that, like
```swift
#if METRICS        
        metricsReporter?.beforeDisplayLinkCallback(displayLink: displayLink)
         defer { metricsReporter?.afterDisplayLinkCallback(displayLink: displayLink) }
#endif
```
Maybe I’m being paranoid but this way we would be sure that measurement code won’t have any effect on regular usage.

OdNairy: It was a great idea, but unfortunately, I cannot make it work  :( 
image

Sources/MapboxMaps/Foundation/MapViewMetricsReporter.swift
OdNairy: I'm not quite following. Why should we require the object to be a class?

PR #1576, opened 2022-09-13
Parse GeoJSON data on a background queue

evil159:
This PR offloads GeoJSON data conversion to a background queue, this addresses an issue with main thread being blocked during the conversion phase.
Currently GeoJSON data is converted into a dictionary on the main thread and then passed it as a parameter to `StyleManager.addStyleSource(forSourceId:properties)` to be passed to core and parsed on the main thread as well.

This change adds an empty GeoJSONSource first, then converts GeoJSON data into a string(to avoid ObjC -> C++ conversion penalty) on a background queue with hight priority, finally passes the GeoJSON data string to core(on the main thread) where it is parsed on a background thread. GeoJSON data is added separately from its source to leverage async parsing from the core.

This results in main thread being unoccupied during pretty much the whole time it takes to load GeoJSON data.

|||
| ----- | ----- |
| **Before** |  |
| **After** |   |

## Pull request checklist:
    

Sources/MapboxMaps/Camera/BasicCameraAnimatorImpl.swift
```diff
@@ -43,7 +43,7 @@ internal final class BasicCameraAnimatorImpl: BasicCameraAnimatorProtocol {
 
     private let mapboxMap: MapboxMapProtocol
 
-    private let mainQueue: MainQueueProtocol
+    private let mainQueue: DispatchQueueProtocol
```
kiryldz, line 46: should `mainQueue` be renamed (here and below) as well?

evil159: The instance here is still a main queue, the name of the protocol got updated to better reflect the built-in type it modeled after(`DispatchQueue`).

Sources/MapboxMaps/Style/StyleSourceManager.swift
```diff
@@ -0,0 +1,180 @@
...
+            return styleManager.addStyleSource(forSourceId: id, properties: properties)
+        }
+    }
+
+    internal func removeSource(withId id: String) throws {
+        try handleExpected {
+            return styleManager.removeStyleSource(forSourceId: id)
+        }
+
+        workItems.removeValue(forKey: id)?.cancel()
```
kiryldz, line 110: that use-case is not covered in Android, I'll create the follow-up task

Sources/MapboxMaps/Style/StyleSourceManager.swift
```diff
@@ -0,0 +1,180 @@
...
+
+    private func applyGeoJSONData(data: GeoJSONSourceData, sourceId id: String) {
+        workItems.removeValue(forKey: id)?.cancel()
+
+        let item = DispatchWorkItem { [weak self] in
+            if self == nil { return }
+
+            let json = try! data.toString()
+
+            self?.mainQueue.async { [weak self] in
```
kiryldz, line 168: what if we did schedule task in the `mainQueue` but before closure gets executed `removeSource` for given geojson get called? then `self` inside closure is still there (as the map object is alive) but the source is already removed at that point

evil159: Exactly! In this case there will be an attempt to set "data" for an unexisting source - we'll catch the error and log it.

kiryldz (approved):

maios (commented):
- Should we also cancel all unfinished work items in deinit?
- What do you think about using OperationQueue?

Sources/MapboxMaps/Style/StyleSourceManager.swift
```diff
@@ -0,0 +1,180 @@
...
+        try addSourceInternal(emptySource, id: id)
+
+        applyGeoJSONData(data: data, sourceId: id)
+    }
+
+    private func applyGeoJSONData(data: GeoJSONSourceData, sourceId id: String) {
+        workItems.removeValue(forKey: id)?.cancel()
+
+        let item = DispatchWorkItem { [weak self] in
+            if self == nil { return }
```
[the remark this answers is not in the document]
evil159: In order not to capture self needlessly(`toString` just below can take some time during which `self` might be deallocated).

Sources/MapboxMaps/Style/StyleSourceManager.swift
```diff
@@ -0,0 +1,180 @@
...
+        var emptySource = source
+        emptySource.data = .empty
+
+        try addSourceInternal(emptySource, id: id)
+
+        applyGeoJSONData(data: data, sourceId: id)
+    }
+
+    private func applyGeoJSONData(data: GeoJSONSourceData, sourceId id: String) {
+        workItems.removeValue(forKey: id)?.cancel()
```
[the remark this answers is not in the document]
evil159: This implementation favors the first submitted task and the last, this was a conscious decision.

Sources/MapboxMaps/Style/StyleSourceManager.swift
```diff
@@ -0,0 +1,180 @@
...
+    private typealias SourceId = String
+
+    internal static func sourcePropertyDefaultValue(for sourceType: String, property: String) -> StylePropertyValue {
+        return StyleManager.getStyleSourcePropertyDefaultValue(forSourceType: sourceType, property: property)
+    }
+
+    private let styleManager: StyleManagerProtocol
+    private let mainQueue: DispatchQueueProtocol
+    private let backgroundQueue: DispatchQueueProtocol
+    private var workItems = [SourceId: DispatchWorkItem]()
```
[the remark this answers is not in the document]
evil159: Like this?

```suggestion
    private var workItems = [SourceId: BlockCancellable]()
```

Sources/MapboxMaps/Style/StyleSourceManager.swift
```diff
@@ -0,0 +1,184 @@
...
+
+        if case GeoJSONSourceData.empty = data {
+            try addSourceInternal(source, id: id)
+            return
+        }
+
+        var emptySource = source
+        emptySource.data = .empty
+
+        try addSourceInternal(emptySource, id: id)
```
maios, line 159: nit: in both 3 cases we are calling to add internal source with either nil or empty data, can we simply it like this?
```
var source = source
let sourceData: GeoJSONSourceData! = source.data

source.data = sourceData.map { _ in GeoJSONSourceData.empty }
try addSourceInternal(source, id: id)

if sourceData != nil {
    applyGeoJSONData(data: sourceData, sourceId: id)
}
```

maios (approved):
🎉

Sources/MapboxMaps/Foundation/RetroactiveConformances/DispatchQueueProtocol.swift
```diff
@@ -0,0 +1,27 @@
...
+
+extension DispatchQueueProtocol {
+    func async(
+        group: DispatchGroup? = nil,
+        qos: DispatchQoS = .unspecified,
+        flags: DispatchWorkItemFlags = [],
+        execute work: @escaping @convention(block) () -> Void
+    ) {
+        async(group: group, qos: qos, flags: flags, execute: work)
+    }
```
OdNairy, line 23: I have some concerns about this one.
Methods implementation in protocol extensions has static dispatching if the method was not included in the protocol. 
IIRC function with default values would be represented by method without args.
I think we need to discuss this approach and do some extra testing.

Sources/MapboxMaps/Foundation/Extensions/Encodable+ToJSON.swift
```diff
@@ -16,4 +16,13 @@ internal extension Encodable {
...
     }
+
+    /// Given an Encodable object return the JSON representation as a string
+    /// - Throws: Errors occurring during conversion.
+    /// - Returns: A string with JSON representing the object.
+    func toString() throws -> String {
+        let data = try JSONEncoder().encode(self)
+
+        return String(decoding: data, as: UTF8.self)
+    }
```
OdNairy, line 27: I personally don't like the idea to have a method like that where JSONEncoder will be instantiated. 
The code is trivial and has no options to manipulate JSONEncoder properties.
However, I'm not going to insist on it.

Sources/MapboxMaps/Foundation/Extensions/Encodable+ToJSON.swift
```diff
@@ -16,4 +16,13 @@ internal extension Encodable {
...
         return array.first!
     }
+
+    /// Given an Encodable object return the JSON representation as a string
+    /// - Throws: Errors occurring during conversion.
+    /// - Returns: A string with JSON representing the object.
+    func toString() throws -> String {
+        let data = try JSONEncoder().encode(self)
+
+        return String(decoding: data, as: UTF8.self)
```
OdNairy, line 26: What would happen if the data is not UTF-8 representable? Method does not throw and do not return optional string. Would it be a runtime NSExpection then?
Would it be better to use safe APIs instead?

OdNairy (commented):
I have a concern about `mapLoaded`/`styleLoaded` events. Would the background GeoJSON processing impact them? Does it mean that customers would get `styleLoaded` event earlier than expected?
Another main concern is related to the method implementation in the `DispatchQueueProtocol` and method dispatching.

Sources/MapboxMaps/Style/Types/GeoJSONSourceData.swift
```diff
@@ -36,6 +37,11 @@ public enum GeoJSONSourceData: Codable {
             return
         }
 
+        if let decodedString = try? container.decode(String.self), decodedString.isEmpty {
+            self = .empty
+            return
+        }
```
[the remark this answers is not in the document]
evil159: This is needed to be able to decode `GeoJSONSourceData.empty` - I think it is a bug that it is possible to encode an empty data, but not to decode it.

Sources/MapboxMaps/Viewport/Helpers/MainQueue.swift
```diff
@@ -1,11 +0,0 @@
...
-// allow mocking the main queue in tests which avoids the need for waits
-internal protocol MainQueueProtocol: AnyObject {
-    func async(execute work: @escaping () -> Void)
-}
-
-internal final class MainQueue: MainQueueProtocol {
-    internal func async(execute work: @escaping () -> Void) {
-        DispatchQueue.main.async(execute: work)
-    }
-}
```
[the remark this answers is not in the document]
evil159: Not sure if enforcing particular instances with protocols brings much benefits - `DispatchQueue.main` is an instance of `DispatchQueue` after all. Even this approach with `MainQueueProtocol` never enforced anything - nothing prevented somebody to define another conformant(of the `MainQueueProtocol`) with whichever dispatch queue underneath and pass it to a place that accepts `MainQueueProtocol`.
I think things like this could be checked in unit tests.

Sources/MapboxMaps/Foundation/Extensions/Encodable+ToJSON.swift
evil159: Couldn't find the answer to your question - converted it to a failable initializer :)

Sources/MapboxMaps/Foundation/Extensions/Encodable+ToJSON.swift
evil159: Any alternatives in mind? There's no particular significance in using `JSONEncoder` indeed - it just was the easiest way to obtain a string from an object.

Sources/MapboxMaps/Style/StyleSourceManager.swift
```diff
@@ -0,0 +1,184 @@
...
+        backgroundQueue: DispatchQueueProtocol = DispatchQueue(label: "GeoJSON parsing queue", qos: .userInitiated)
+    ) {
+        self.styleManager = styleManager
+        self.mainQueue = mainQueue
+        self.backgroundQueue = backgroundQueue
+    }
+
+    // MARK: - Typed API
+
+    internal func source<T>(withId id: String, type: T.Type) throws -> T where T: Source {
```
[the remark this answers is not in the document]
evil159: Yeah, I applied your suggestion in other commit

PR #1596, opened 2022-09-23
Update update-version.sh to have recommended format for CFBundleShortVersionString

maios:
> The required format is three period-separated integers, such as 10.14.1. The string can only contain numeric characters (0-9) and periods.

[CFBundleShortVersionString](https://developer.apple.com/documentation/bundleresources/information_property_list/cfbundleshortversionstring)

As discussed yesterday with @evil159 we found out that the short version does not match the recommended/required format by Apply if our semantic version is as following `x.y.z-beta.1-something`

### DIFF
--- a/scripts/release/update-version.sh
+++ b/scripts/release/update-version.sh
@@ -20,7 +20,7 @@ fi
 
 SEM_VERSION=$1
 SEM_VERSION=${SEM_VERSION/#v}
-SHORT_VERSION=${SEM_VERSION%-*}
+SHORT_VERSION=${SEM_VERSION%%-*}
 
 brew_install_if_needed jq
 


OdNairy (commented):
I still have some concerns about dropping all the info. I suppose we use this version somewhere in the telemetry. So it would be impossible to detect public/private release via logs

maios:
> I still have some concerns about dropping all the info. I suppose we use this version somewhere in the telemetry. So it would be impossible to detect public/private release via logs

@OdNairy  The full semantic version is updated in MapboxMaps.json and MapboxMaps.podspec, can we detect from these?

OdNairy (approved):

OdNairy:
> > I still have some concerns about dropping all the info. I suppose we use this version somewhere in the telemetry. So it would be impossible to detect public/private release via logs
> 
> @OdNairy The full semantic version is updated in MapboxMaps.json and MapboxMaps.podspec, can we detect from these?

I don't think that MME takes the User-Agent fragment from versions.json.

PR #1604, opened 2022-09-27
Resolve initial view annotation placement issue

ZiZasaurus:
Removed the addition of view annotation to the containerView, which caused the view annotation to initially load in the top left corner.

### DIFF
--- a/Sources/MapboxMaps/Annotations/ViewAnnotationManager.swift
+++ b/Sources/MapboxMaps/Annotations/ViewAnnotationManager.swift
@@ -149,15 +149,17 @@ public final class ViewAnnotationManager {
         }
 
         view.translatesAutoresizingMaskIntoConstraints = false
+        view.isHidden = true 
 
         let id = id ?? UUID().uuidString
         try mapboxMap.addViewAnnotation(withId: id, options: creationOptions)
         viewsById[id] = view
         idsByView[view] = id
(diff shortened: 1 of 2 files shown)


maios (approved):

Sources/MapboxMaps/Annotations/ViewAnnotationManager.swift
```diff
@@ -158,7 +158,6 @@ public final class ViewAnnotationManager {
         if let featureId = creationOptions.associatedFeatureId {
             viewsByFeatureIds[featureId] = view
         }
-        containerView.addSubview(view)
```
evil159, line 161: Would the removal of this line generate the "Superview changed for annotation view. Use `ViewAnnotationManager.remove(_ view: UIView)` instead to remove it from the layout calculation." warning?

ZiZasaurus: Yes, this warning is generated when this line is removed.

maios:
Some tests are failing, could you check that?
Also with this new implementation, the observers will be notified about annotation view's visibility changed when the view is first added, could we add a test case for this one and maybe a changelog entry also?

Deloww (commented):
https://github.com/mapbox/mapbox-maps-ios/pull/1604/files

evil159 (approved):

PR #1613, opened 2022-10-03
[scripts] Update E2E trigger checker

jmalanen:


OdNairy (commented):
This fix would resolve the initial issue since it was raised before the script was executed. 
The issue in the line 490.
https://github.com/mapbox/mapbox-maps-ios/blob/25ab646a3a00b1ab33d9963037c90772d132fe53/.circleci/config.yml#L490-L499

jmalanen:
@OdNairy On second thought I think it's better to handle it in circle.yml like in your commit: https://github.com/mapbox/mapbox-maps-ios/commit/775f3ac9527d2f29ae4ea18547301f0d9f002920

Please merge it to `main`.

PR #1621, opened 2022-10-04
Expose list of added view annotations

maios:
## Pull request checklist:

### DIFF
--- a/Sources/MapboxMaps/Annotations/ViewAnnotationManager.swift
+++ b/Sources/MapboxMaps/Annotations/ViewAnnotationManager.swift
@@ -60,6 +60,9 @@ public final class ViewAnnotationManager {
     /// The default value is true, and setting this value to false will disable the validation.
     public var validatesViews = true
 
+    /// The complete list of annotations associated with the receiver.
+    public var annotations: [UIView] { Array(viewsById.values) }
+
     internal init(containerView: UIView, mapboxMap: MapboxMapProtocol) {
         self.containerView = containerView
         self.mapboxMap = mapboxMap
--- a/Sources/MapboxMaps/Annotations/ViewAnnotationManager.swift
+++ b/Sources/MapboxMaps/Annotations/ViewAnnotationManager.swift
@@ -9,19 +9,6 @@ public enum ViewAnnotationManagerError: Error {
     case geometryFieldMissing
 }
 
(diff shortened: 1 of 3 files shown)


Sources/MapboxMaps/Annotations/ViewAnnotationManager.swift
```diff
@@ -60,6 +60,9 @@ public final class ViewAnnotationManager {
     /// The default value is true, and setting this value to false will disable the validation.
     public var validatesViews = true
 
+    /// The complete list of annotations associated with the receiver.
+    public var annotations: [UIView] { Array(viewsById.values) }
```
alexshalamov, line 64: Should we also return `ViewAnnotationOptions` associated with the view? Otherwise, engine (developer) would need to iterate twice:

```
for (view : manager.annotations) {
  coordinate = manager.options(view).geometry // lookup 
}
```

maios: that makes sense, I have amended the api to return list of added annotation view and its options, thank you!

Sources/MapboxMaps/Annotations/ViewAnnotationManager.swift
```diff
@@ -9,6 +9,19 @@ public enum ViewAnnotationManagerError: Error {
     case geometryFieldMissing
 }
 
+/// Represents an annotation view.
+public struct ViewAnnotation {
```
evil159, line 13: It seems to me that introducing a whole new entity is a bit overkill for this case, would `Dictionary< ViewAnnotationOptions, UIView>` do the trick here?

maios: > Clients need to wrap crud methods and keep the list on their side if they want to iterate over view annotations and perform some action.

If I understand the task correctly, this is to move this from client to our code base, keep the list of added view and its options. If we do a dictionary like suggested then again the client would need to keep a list of options on their side.

evil159: With `annotations` defined like this: `public var annotations: Dictionary<ViewAnnotationOptions, UIView>`, `annotations.keys` returns the list of options and `annotations.values` returns the list of views.

evil159 (approved):

PR #1627, opened 2022-10-06
Prevent view annotation being shown erroneously after options update

evil159:
Updating options for an off-screen annotation with `visibility` set to `true` unhides the corresponding view which may result in it being displayed on top of the map. This PR addresses this issue by preventing updated view annotation to be un-hidden during the update. 
With this change view annotation are only shown as a result of view annotation position update event.

## Pull request checklist:

### DIFF
--- a/Sources/MapboxMaps/Annotations/ViewAnnotationManager.swift
+++ b/Sources/MapboxMaps/Annotations/ViewAnnotationManager.swift
@@ -220,7 +220,7 @@ public final class ViewAnnotationManager {
         let currentFeatureId = try? mapboxMap.options(forViewAnnotationWithId: id).associatedFeatureId
         try mapboxMap.updateViewAnnotation(withId: id, options: options)
 
-        if let visible = options.visible, visible == false {
+        if options.visible == false {
             expectedHiddenByView[view] = true
             view.isHidden = true
         }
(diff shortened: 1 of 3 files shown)


Sources/MapboxMaps/Annotations/ViewAnnotationManager.swift
```diff
@@ -219,9 +219,12 @@ public final class ViewAnnotationManager {
         }
         let currentFeatureId = try? mapboxMap.options(forViewAnnotationWithId: id).associatedFeatureId
         try mapboxMap.updateViewAnnotation(withId: id, options: options)
-        let isHidden = !(options.visible ?? true)
-        expectedHiddenByView[view] = isHidden
-        viewsById[id]?.isHidden = isHidden
+
+        if let visible = options.visible, visible == false {
```
OdNairy, line 223: It is the same, isn't it?
```suggestion
        if visible == false {
```

OdNairy: And if it's not, can we add some tests to catch the corner case, please? This Optional<Bool> logic is really tricky 😅

evil159: Same thing indeed, not sure why I went with more verbose syntax

OdNairy (approved):

PR #1625, opened 2022-10-06
Smooth reduced accuracy ring radius interpolation

evil159:
This PR adjusts interpolation steps of `reducedAccuracy`'s authorization accuracy ring radius to eliminate sudden changes of the ring radius. Additionally on higher zoom levels(>=8) radius of the accuracy ring is set to reflect the accuracy received from location provider.

### Zoom out
| Before | After |
| ----- | ----- |
| <video src="https://user-images.githubusercontent.com/1478430/194303560-8c61c5de-5a0a-431c-a536-cf32903bd99a.MP4" width = 250/> | <video src="https://user-images.githubusercontent.com/1478430/194303595-7d31dd03-ea58-4a44-bc13-bf2f19a7605d.MP4" width = 250/> |

### Zoom in
| Before | After |
| ----- | ----- |
| <video src="https://user-images.githubusercontent.com/1478430/194303772-b734a454-7570-4b25-b36d-ff63b4c59b16.MP4" width = 250/> | <video src="https://user-images.githubusercontent.com/1478430/194303854-3960a6ff-caf7-4a84-a9d5-4a756947edc0.mp4" width = 250/> |

## Pull request checklist:
 - [x] Describe the changes in this PR, especially public API changes.
 - [x] Include before/after visuals or gifs if this PR includes visual changes.
 - [x] Write tests for all new functionality. If tests were not written, please explain why.
 - [x] Add a changelog entry to to bottom of the relevant section (typically the `## main` heading near the top).
 - [ ] If this PR is a `v10.[version]` release branch fix / enhancement, merge it to `main` first and then port to `v10.[version]` release branch.

PRs must be submitted under the terms of our Contributor License Agreement [CLA](https://github.com/mapbox/mapbox-maps-ios/blob/main/CONTRIBUTING.md#contributor-license-agreement).

PR #1644, opened 2022-10-13
Improve reduced accuracy authorization location indicator behavior while zooming

evil159:
This PR further improves accuracy ring radius interpolation started in https://github.com/mapbox/mapbox-maps-ios/pull/1625

## Pull request checklist:

### DIFF
--- a/Sources/MapboxMaps/Location/Puck/Puck2D.swift
+++ b/Sources/MapboxMaps/Location/Puck/Puck2D.swift
@@ -162,20 +162,39 @@ internal final class Puck2D: Puck {
         case .reducedAccuracy:
             fallthrough
         @unknown default:
+            let zoomCutoffRange: ClosedRange<Double> = 4.0...7.5
+            let accuracyRange: ClosedRange<CLLocationDistance> = 1000...20_000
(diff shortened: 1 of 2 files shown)


Sources/MapboxMaps/Location/Puck/Puck2D.swift
```diff
@@ -313,3 +346,9 @@ private extension Puck2DConfiguration.Pulsing.Radius {
         }
     }
 }
+
+private extension ClosedRange where Bound == Double {
```
maios, line 350: can we have this extension as `internal` so we don't have a same function in unit tests?

maios, line 350: nit: can we have a little more generic constraints and not just Double only
```suggestion
private extension ClosedRange where Bound: AdditiveArithmetic {
```

maios (approved):

Sources/MapboxMaps/Location/Puck/Puck2D.swift
evil159: Nice tip, I got lost in Swift's numerics :)

PR #1645, opened 2022-10-13
Increase timeout for view annotation animation example

evil159:


Apps/Examples/Examples/Models/Examples.swift
```diff
@@ -138,7 +138,7 @@ struct Examples {
                 type: ViewAnnotationMarkerExample.self),
         Example(title: "View annotations: animation",
                 description: "Animate a view annotation along a route",
-                testTimeout: 60,
+                testTimeout: 120,
```
OdNairy, line 141: I'm not quite sure that it's a good idea to have such a long animation in our ExampleTests

evil159: Fair :) I reduced the animation duration instead

OdNairy (approved):

PR #1650, opened 2022-10-17
Avoid triggering assertion for the 3D puck layer when returning `allLayerIdentifiers`

evil159:
3D puck uses a special layer, it trips code in `allLayerIdentifiers` getter into triggering an assertion when obtaining layer info. This PR makes the `allLayerIdentifiers` getter to ignore 3D puck layer altogether.

## Pull request checklist:

### DIFF
--- a/Tests/MapboxMapsTests/Foundation/Style/StyleTests.swift
+++ b/Tests/MapboxMapsTests/Foundation/Style/StyleTests.swift
@@ -120,6 +120,13 @@ final class StyleTests: XCTestCase {
         })
     }
 
+    func testGetAllLayerIdentifiersDoesNotTriggerAssertFor3DPuckLayer() {
+        styleManager.getStyleLayersStub.defaultReturnValue = [StyleObjectInfo(id: Puck3D.layerID, type: "model")]
+
+        // test fails if there is an assertion triggered here
+        _ = style.allLayerIdentifiers
+    }
+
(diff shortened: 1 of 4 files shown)


azarovalex (approved):

Tests/MapboxMapsTests/Foundation/Style/StyleTests.swift
```diff
@@ -120,6 +120,13 @@ final class StyleTests: XCTestCase {
         })
     }
 
+    func testGetAllLayerIdentifiersDoesNotTriggerAssertFor3DPuckLayer() {
+        styleManager.getStyleLayersStub.defaultReturnValue = [StyleObjectInfo(id: Puck3D.layerID, type: "model")]
+
+        // test fails if there is an assertion triggered here
+        _ = style.allLayerIdentifiers
```
OdNairy, line 127: Can we check that all identifiers just don't include the 3D puck's one? We might have tests running on the Release config someday.

OdNairy (approved):
It looks like a good workaround but should we consider some more stable fix there? 🤔

Tests/MapboxMapsTests/Foundation/Style/StyleTests.swift
evil159: Yes, makes sense, I updated the test

PR #1653, opened 2022-10-18
Fix view annotation animation example crashing on iOS 13

evil159:
The reason why this example is crashing on iOS 13 is the `.common` runloop mode that is being used to add display link to the runloop. Changing the mode to `.default` eliminates the issue.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -112,7 +112,7 @@ workflows:
           scheme: Examples
           requires:
             - Build Examples tests [non-main]
-          devices: --device model=iphone8,version=14.7
+          devices: --device model=iphone11,version=13.6
 
   steve:
     unless: << pipeline.parameters.turf-revision >>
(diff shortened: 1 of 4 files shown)


maios (approved):
Can you explain about new change about events observation in this example?

evil159:
> Can you explain about new change about events observation in this example?

I don't have much info beyond some basic understanding about this, I wasn't able to tinker with it too much since it reproduces only on iOS 13.
There are different runloop modes we can use to add a display link to the runloop, that's how they behave in relation to touch even handling
* `.default` - does not provide callbacks when touch(tracking) is happening
* `.common` - provides callbacks during when touch(tracking) is happening
* `.tracking` - provides callbacks only when touch(tracking) is happening

For the purpose of this example `.common` is the best fit, however it is crashing on iOS 13, and given there's virtually no difference with `.default` mode - I switched to that for this example.

PR #1665, opened 2022-10-26
Bugfix/mapsios 328 better handling set style result observers

maios:
Fixed MAPSIOS-328
Previously when call to `MapboxMap/loadStyleURI(_:completion:)` and `MapboxMap/loadStyleJSON(_:completion:)`, we will observer 2 events `styleLoaded`and `mapLoadingError`; so even when the given style is successfully loaded, but there is some other error i.e source cannot be loaded, then `mapLoadingError`will fire with error type set to source, we will, however, observe this event and because we have not cancelled our observations when the first event related to our style occured, the completion blocked will be invoked more than once.
This PR fixes this problem by cancelling both observation when first even is received.
## Pull request checklist:

### DIFF
--- a/Sources/MapboxMaps/Foundation/MapboxMap.swift
+++ b/Sources/MapboxMaps/Foundation/MapboxMap.swift
@@ -79,16 +79,22 @@ public final class MapboxMap: MapboxMapProtocol {
     // MARK: - Style loading
 
     private func observeStyleLoad(_ completion: @escaping (Result<Style, Error>) -> Void) {
-        onNext(event: .styleLoaded) { [style] _ in
+        let cancellable = CompositeCancelable()
+
+        cancellable.add(onNext(event: .styleLoaded) { [style] _ in
             if !style.isLoaded {
                 Log.warning(forMessage: "style.isLoaded == false, was this an empty style?", category: "Style")
             }
             completion(.success(style))
-        }
+            cancellable.cancel()
+        })
 
-        onNext(event: .mapLoadingError) { event in
+        cancellable.add(onEvery(event: .mapLoadingError) { event in
+            guard case .style = event.payload.error else { return }
+
             completion(.failure(event.payload.error))
-        }
+            cancellable.cancel()
+        })
     }
 
     /// Loads a `style` from a StyleURI, calling a completion closure when the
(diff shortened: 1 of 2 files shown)


Tests/MapboxMapsTests/Foundation/MapboxMapTests.swift
```diff
@@ -238,6 +238,41 @@ final class MapboxMapTests: XCTestCase {
...
+        let styleLoadEventOccurred = expectation(description: "style-loaded event occurred")
+        let mapLoadingErrorEventOccurred = expectation(description: "map-loading-error event occurred")
+        let completionIsCalledOnce = expectation(description: "loadStyle completion should be called once")
+        completionIsCalledOnce.assertForOverFulfill = true
+
+        let mapboxObservable = try XCTUnwrap(mapboxObservableProviderStub.invocations.first?.returnValue as? MockMapboxObservable)
+        mapboxObservable.onTypedNextStub.defaultSideEffect = { invocation in
+            guard invocation.parameters.eventName == "style-loaded" else { return }
+
+            DispatchQueue.main.asyncAfter(deadline: .now() + 0.5) {
```
OdNairy, line 251: Can we avoid `asyncAfter` in this place?

OdNairy: If not, can we please use lower values like 0.1 and 0.2

Sources/MapboxMaps/Foundation/MapboxMap.swift
```diff
@@ -79,16 +79,22 @@ public final class MapboxMap: MapboxMapProtocol {
...
+        cancellable.add(onNext(event: .styleLoaded) { [style] _ in
             if !style.isLoaded {
                 Log.warning(forMessage: "style.isLoaded == false, was this an empty style?", category: "Style")
             }
             completion(.success(style))
-        }
+            cancellable.cancel()
+        })
+
+        cancellable.add(onEvery(event: .mapLoadingError) { event in
```
OdNairy, line 92: Why did we change `onNext` to `onEvery`?

maios: there might be other error represented by `mapLoadingError` arrive before style load error arrives, so I want to make sure we are catching error event indicating the given style cannot be loaded here.

OdNairy: Makes sense, good catch 👍

OdNairy (approved):
LGTM

PR #1678, opened 2022-11-01
Draggable annotations review suggestions

evil159:


Sources/MapboxMaps/Gestures/GestureRecognizers/MapboxLongPressGestureRecognizer.swift
```diff
@@ -0,0 +1,50 @@
...
+    internal func setTranslation(_ translation: CGPoint, in view: UIView?) {
+        guard let newTranslation = self.view?.convert(translation, from: view) else {
+            return
+        }
+
+        let touchLocation = location(in: self.view)
+
+        self.translationReference = CGPoint(
+            x: touchLocation.x + newTranslation.x,
+            y: touchLocation.y + newTranslation.y
```
OdNairy, line 46: Should we enforce the check that `translationReference` has the same `view` all the time? 🤔

Sources/MapboxMaps/Gestures/GestureRecognizers/MapboxLongPressGestureRecognizer.swift
```diff
@@ -0,0 +1,50 @@
+import Foundation
+import UIKit
+
+internal final class MapboxLongPressGestureRecognizer: UILongPressGestureRecognizer {
```
OdNairy, line 4: Can you, please, add some documentation on how this custom gesture recognizer differs from the standard one.

Issue #652, opened 2021-09-07
Unable to localize labels into Chinese


1ec5:
`Style.localizeLabels(into:forLayerIds:)` is unable to localize the map’s labels into Simplified Chinese or, one of [the locales supported by Mapbox Streets source v8](https://docs.mapbox.com/vector-tiles/reference/mapbox-streets-v8/#name-text--name_lang-code-text). Attempting to localize the labels into Simplified Chinese results in a symbol layer that is either unlocalized or completely blank. This issue also affects Hong Kong Traditional Chinese.

## Diagnosis

As of #480, `Style.getLocaleValue(locale:)` hard-codes the locale codes supported by Mapbox Streets source v8:

https://github.com/mapbox/mapbox-maps-ios/blob/34ea7ada3cea926ea308d00dd003178a16d3164e/Sources/MapboxMaps/Style/Style%2BLocalization.swift#L38

However, it matches the passed-in Locale’s bare language code against that list:

https://github.com/mapbox/mapbox-maps-ios/blob/34ea7ada3cea926ea308d00dd003178a16d3164e/Sources/MapboxMaps/Style/Style%2BLocalization.swift#L40-L43

A phone may be set to a Locale with the identifier `zh-Hant-HK`, meaning Hong Kong Traditional Chinese. But this Locale’s language code is just `zh`, so despite `zh-Hant` being supported by the Streets source, the method ends up localizing into `zh`, which is just Chinese. The special logic to map `zh-Hant-TW` to `zh-Hant` does not apply to `zh-Hant-HK`.

Another common identifier is `zh-Hans-CN`, which also falls through the cracks and gets mapped to just `zh`.

`convertExpressionForLocalization(symbolLayer:localeValue:)` ends up replacing each `name_en` get expression with a `name_zh` get expression. But the Streets source no longer supports the generic `zh` locale as of v8; instead, only `name_zh-Hans` and `name_zh-Hant` fields are provided. Depending on the style’s fallback rules, the label will either appear in the local language rather than Chinese, or it will be blank.

## Suggestion

`supportedLocaleIdentifiers` should be a list of locale codes that appear in the names of `name_*` fields in the Streets source – nothing less, nothing more.

When matching against this array, check both `Locale.languageCode` and `Locale.regionCode`.

/cc @mapbox/maps-ios @mapbox/navigation-ios

1ec5:
[...]

PR #1687, opened 2022-11-03
Fix language support and label localization 

pjleonard37:
This PR addresses #652, properly localizing labels into the preferred script based on Locale. To do so, this PR follows -- with some modifications to avoid breaking changes -- the approach from the Navigation SDK (see [here](https://github.com/mapbox/mapbox-navigation-ios/blob/28d13ff313cb44c47f69c6e386ec6f3dd855e63e/Sources/MapboxNavigation/VectorSource.swift#L27)). This approach provides more robust support for a Locale's preferred language and script option, by checking the preference against the set of language + scripts supported by Mapbox Streets [v8](https://docs.mapbox.com/data/tilesets/reference/mapbox-streets-v8/#common-fields) or [v7](https://docs.mapbox.com/data/tilesets/reference/legacy/mapbox-streets-v7/#name-fields). Additionally, test coverage is expanded to cover more Locales and the localization example modified to include a "Device Locale" option. 

https://user-images.githubusercontent.com/7976026/199850097-fc746fba-6d91-4d4a-a23e-2eca6fec3b11.mp4

## Pull request checklist:
    
### DIFF
--- a/Apps/Examples/Examples/All Examples/LocalizationExample.swift
+++ b/Apps/Examples/Examples/All Examples/LocalizationExample.swift
@@ -47,10 +47,14 @@ public class LocalizationExample: UIViewController, ExampleProtocol {
                                       message: "Please select a language to localize to.",
                                       preferredStyle: .actionSheet)
 
-        alert.addAction(UIAlertAction(title: "Device Settings", style: .default, handler: { [weak self] _ in
-            try! self?.mapView.mapboxMap.style.localizeLabels(into: Locale(identifier: Locale.preferredLanguages[0]))
+        alert.addAction(UIAlertAction(title: "Device Locale", style: .default, handler: { [weak self] _ in
+            do {
+                try self?.mapView.mapboxMap.style.localizeLabels(into: Locale(identifier: Locale.preferredLanguages[0]))
+            } catch {
+                print(error)
+            }
         }))
-        
+
         alert.addAction(UIAlertAction(title: "Spanish", style: .default, handler: { [weak self] _ in
             try! self?.mapView.mapboxMap.style.localizeLabels(into: Locale(identifier: "es"))
         }))
--- a/Apps/Examples/Examples/All Examples/LocalizationExample.swift
+++ b/Apps/Examples/Examples/All Examples/LocalizationExample.swift
@@ -49,7 +49,7 @@ public class LocalizationExample: UIViewController, ExampleProtocol {
 
         alert.addAction(UIAlertAction(title: "Device Locale", style: .default, handler: { [weak self] _ in
             do {
-                try self?.mapView.mapboxMap.style.localizeLabels(into: Locale(identifier: Locale.preferredLanguages[0]))
+                try self?.mapView.mapboxMap.style.localizeLabels(into: Locale.current)
             } catch {
                 print(error)
             }
--- a/Apps/Examples/Examples/All Examples/LocalizationExample.swift
+++ b/Apps/Examples/Examples/All Examples/LocalizationExample.swift
@@ -47,6 +47,10 @@ public class LocalizationExample: UIViewController, ExampleProtocol {
                                       message: "Please select a language to localize to.",
                                       preferredStyle: .actionSheet)
 
+        alert.addAction(UIAlertAction(title: "Device Settings", style: .default, handler: { [weak self] _ in
+            try! self?.mapView.mapboxMap.style.localizeLabels(into: Locale(identifier: Locale.preferredLanguages[0]))
+        }))
+        
         alert.addAction(UIAlertAction(title: "Spanish", style: .default, handler: { [weak self] _ in
             try! self?.mapView.mapboxMap.style.localizeLabels(into: Locale(identifier: "es"))
         }))
--- a/Sources/MapboxMaps/Style/Style+Localization.swift
+++ b/Sources/MapboxMaps/Style/Style+Localization.swift
@@ -31,40 +31,36 @@ extension Style {
         }
     }
 
-    /**
-     Returns the BCP 47 language tag supported by Mapbox Streets source v8 that is most preferred according to the given preferences.
-     */
+    /// Returns the BCP 47 language tag supported by Mapbox Streets source v8 that is most preferred according to the given preferences.
+    /// Docs for language, region, and script codes: https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPInternational/LanguageandLocaleIDs/LanguageandLocaleIDs.html
     internal func preferredMapboxStreetsLocalization(among preferences: [String]) -> String? {
         let supportedLocaleIdentifiersv8 = ["ar", "en", "es", "fr", "de", "it", "pt", "ru", "zh-Hans", "zh-Hant", "ja", "ko", "vi"].map(Locale.init(identifier:))
-        
+
         let preferredLocales = preferences.map(Locale.init(identifier:))
+
         let acceptsEnglish = preferredLocales.contains { $0.languageCode == "en" }
         var availableLocales = supportedLocaleIdentifiersv8
         if !acceptsEnglish {
             availableLocales.removeAll { $0.languageCode == "en" }
         }
-        
+
         let mostSpecificLanguage = Bundle.preferredLocalizations(from: availableLocales.map { $0.identifier },
                                                                  forPreferences: preferences)
             .max { $0.count > $1.count }
-        
+
         // `Bundle.preferredLocalizations(from:forPreferences:)` is just returning the first localization it could find.
         if let mostSpecificLanguage = mostSpecificLanguage, !preferredLocales.contains(where: { $0.languageCode == Locale(identifier: mostSpecificLanguage).languageCode }) {
             return nil
         }
-        
         return mostSpecificLanguage
     }
-    
-    /// Filters through source to determine supported locale styles.
-    /// This is needed for v7 support
+
+    /// Returns the shortened language identifier string representing a supported Mapbox Streets Localization
+    /// List of supported language identifiers: https://docs.mapbox.com/data/tilesets/reference/mapbox-streets-v8/#common-fields
     internal func getLocaleValue(locale: Locale) -> String? {
-        let preferences: [String]
-        preferences = [locale.identifier]
+        let preferences = [locale.identifier]
         var localeValue: String?
 
-        // Docs for language, region, and script codes  https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPInternational/LanguageandLocaleIDs/LanguageandLocaleIDs.html
-        // List of supported language identifiers: https://docs.mapbox.com/data/tilesets/reference/mapbox-streets-v8/#common-fields
         // Lists those supported by either v7 or v8
         let supportedLocaleIdentifiers = ["ar", "en", "es", "fr", "de", "it", "pt", "ru", "zh-Hans", "zh-Hant", "ja", "ko", "vi", "zh"]
 
@@ -95,7 +91,7 @@ extension Style {
                 }
             }
         }
-        
+
         localeValue = preferredMapboxStreetsLocalization(among: preferences) ?? nil
 
         return localeValue
--- a/Sources/MapboxMaps/Style/Style+Localization.swift
+++ b/Sources/MapboxMaps/Style/Style+Localization.swift
@@ -31,13 +31,41 @@ extension Style {
         }
     }
 
+    /**
+     Returns the BCP 47 language tag supported by Mapbox Streets source v8 that is most preferred according to the given preferences.
+     */
+    internal func preferredMapboxStreetsLocalization(among preferences: [String]) -> String? {
+        let supportedLocaleIdentifiersv8 = ["ar", "en", "es", "fr", "de", "it", "pt", "ru", "zh-Hans", "zh-Hant", "ja", "ko", "vi"].map(Locale.init(identifier:))
+        
+        let preferredLocales = preferences.map(Locale.init(identifier:))
+        let acceptsEnglish = preferredLocales.contains { $0.languageCode == "en" }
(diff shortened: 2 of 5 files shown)


Sources/MapboxMaps/Style/Style+Localization.swift
```diff
@@ -31,41 +31,70 @@ extension Style {
...
+    /// Returns the BCP 47 language tag supported by Mapbox Streets source v8 that is most preferred according to the given preferences.
+    /// Docs for language, region, and script codes: https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPInternational/LanguageandLocaleIDs/LanguageandLocaleIDs.html
+    internal func preferredMapboxStreetsLocalization(among preferences: [String]) -> String? {
+        let supportedLocaleIdentifiersv8 = ["ar", "en", "es", "fr", "de", "it", "pt", "ru", "zh-Hans", "zh-Hant", "ja", "ko", "vi"].map(Locale.init(identifier:))
+
+        let preferredLocales = preferences.map(Locale.init(identifier:))
+
+        let acceptsEnglish = preferredLocales.contains { $0.languageCode == "en" }
+        var availableLocales = supportedLocaleIdentifiersv8
+        if !acceptsEnglish {
```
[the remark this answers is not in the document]
pjleonard37: I took this bit directly from the navigation SDK implementation [here](https://github.com/mapbox/mapbox-navigation-ios/blob/28d13ff313cb44c47f69c6e386ec6f3dd855e63e/Sources/MapboxNavigation/VectorSource.swift#L27). However, I think it only makes sense because they fall back to the device Locale [here](https://github.com/mapbox/mapbox-navigation-ios/blob/28d13ff313cb44c47f69c6e386ec6f3dd855e63e/Sources/MapboxNavigation/VectorSource.swift#L62), which would have information about English preference. I can remove it on the Maps side. 

cc: @1ec5

Sources/MapboxMaps/Style/Style+Localization.swift
```diff
@@ -31,41 +31,64 @@ extension Style {
...
+    /// Returns the BCP 47 language tag supported by Mapbox Streets source v8 that is most preferred according to the given preferences.
+    /// Docs for language, region, and script codes: https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPInternational/LanguageandLocaleIDs/LanguageandLocaleIDs.html
+    internal func preferredMapboxStreetsLocalization(among preferences: [String]) -> String? {
+        let supportedLocaleIdentifiersv8 = ["ar", "en", "es", "fr", "de", "it", "pt", "ru", "zh-Hans", "zh-Hant", "ja", "ko", "vi"].map(Locale.init(identifier:))
+
+        let preferredLocales = preferences.map(Locale.init(identifier:))
+
+        let mostSpecificLanguage = Bundle.preferredLocalizations(from: supportedLocaleIdentifiersv8.map { $0.identifier },
+                                                                 forPreferences: preferences)
+            .max { $0.count > $1.count }
```
maios, line 43: I don't know if getting the most specific/prefered language by comparing the length is the good idea, would it make more sense if we get the first value of the returned `preferrredLocalizations(from:forPreferences)`, consider this will be an ordered list based on user's preferences?

> Return Value
> An array of locale identifiers, ordered according to user preference.
>
> If none of the user-preferred localizations are available, this method returns one of the values in localizationsArray.

evil159 (approved):

Apps/Examples/Examples/All Examples/LocalizationExample.swift
```diff
@@ -47,6 +47,14 @@ public class LocalizationExample: UIViewController, ExampleProtocol {
                                       message: "Please select a language to localize to.",
                                       preferredStyle: .actionSheet)
 
+        alert.addAction(UIAlertAction(title: "Device Locale", style: .default, handler: { [weak self] _ in
+            do {
+                try self?.mapView.mapboxMap.style.localizeLabels(into: Locale(identifier: Locale.preferredLanguages[0]))
```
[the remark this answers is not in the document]
pjleonard37: Updated in latest commit 👍

Sources/MapboxMaps/Style/Style+Localization.swift
```diff
@@ -31,41 +31,64 @@ extension Style {
...
+
+    /// Returns the shortened language identifier string representing a supported Mapbox Streets Localization
+    /// List of supported language identifiers: https://docs.mapbox.com/data/tilesets/reference/mapbox-streets-v8/#common-fields
     internal func getLocaleValue(locale: Locale) -> String? {
-        // Docs for language, region, and script codes  https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPInternational/LanguageandLocaleIDs/LanguageandLocaleIDs.html
-        let supportedLocaleIdentifiers = ["ar", "de", "en", "es", "fr", "it", "ja", "ko", "pt", "ru", "vi", "zh", "zh-Hans", "zh-Hant", "zh-Hant-TW"]
+        var localeValue: String?
+
+        // Lists those supported by either v7 or v8
+        let supportedLocaleIdentifiers = ["ar", "en", "es", "fr", "de", "it", "pt", "ru", "zh-Hans", "zh-Hant", "ja", "ko", "vi", "zh"]
```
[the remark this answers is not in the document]
pjleonard37: They are slightly different to reflect the differences in support on v7 and v8

Sources/MapboxMaps/Style/Style+Localization.swift
pjleonard37: Updated in latest commit 👍

Sources/MapboxMaps/Style/Style+Localization.swift
```diff
@@ -31,41 +31,64 @@ extension Style {
...
 
         // Do nothing if we do not support the locale
-        if !supportedLocaleIdentifiers.contains(locale.languageCode!) {
+        guard supportedLocaleIdentifiers.contains(locale.languageCode!) else {
```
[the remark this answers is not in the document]
pjleonard37: Updated in latest commit 👍 . The language code itself is now returned

pjleonard37:
Thanks for the reviews @maios and @evil159. Based on @maios 's [comment](https://github.com/mapbox/mapbox-maps-ios/pull/1687#discussion_r1014001047) it became clear that the initially proposed method did not allow for passing `Locale.current` or `Locale.autoupdatingCurrent`. Instead, the original approach expected a dev to pass a newly created Locale. I've updated this implementation to allow a dev to pass the current Locale, and then use the set of `Locale.preferredLanguages` (similar to what nav does [here](https://github.com/mapbox/mapbox-navigation-ios/blob/28d13ff313cb44c47f69c6e386ec6f3dd855e63e/Sources/MapboxNavigation/VectorSource.swift#L59)). In either case, preferences are then passed to `preferredMapboxStreetsLocalization(among:)` which returns the most preferred language available.

Sources/MapboxMaps/Style/Style+Localization.swift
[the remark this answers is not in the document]
1ec5: The navigation SDK code is more or less a straight Swift port of [this map SDK code](https://github.com/mapbox/mapbox-gl-native-ios/blob/ddf8e35970d1a8a46171bd798bcfa8f22d2760e0/platform/darwin/src/MGLVectorTileSource.mm#L120-L121). If I remember correctly, the rationale for this English check is that many styles, including the core Mapbox styles, hard-code English by default instead of the local language.

Sources/MapboxMaps/Style/Style+Localization.swift
1ec5: The idea here was to prefer, say, `zh-Hant` over `zh`, all things considered. Be sure to test whether `preferrredLocalizations(from:forPreferences)` is actually ordered by user preference. I remember being surprised about something like that in the past.

Sources/MapboxMaps/Style/Style+Localization.swift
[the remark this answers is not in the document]
1ec5: Note that a language code is only part of a locale identifier.

Sources/MapboxMaps/Style/Style+Localization.swift
```diff
@@ -31,41 +31,61 @@ extension Style {
...
+    /// Returns the BCP 47 language tag supported by Mapbox Streets source v8 that is most preferred according to the given preferences.
+    /// Docs for language, region, and script codes: https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPInternational/LanguageandLocaleIDs/LanguageandLocaleIDs.html
+    internal func preferredMapboxStreetsLocalization(among preferences: [String]) -> String? {
+        let supportedLanguageCodesv8 = ["ar", "en", "es", "fr", "de", "it", "pt", "ru", "zh-Hans", "zh-Hant", "ja", "ko", "vi"].map(Locale.init(identifier:))
+
+        let preferredLocales = preferences.map(Locale.init(identifier:))
+
+        let mostSpecificLanguage = Bundle.preferredLocalizations(from: supportedLanguageCodesv8.map(\.identifier),
+            forPreferences: preferences)
+            .first
```
OdNairy, line 43: Formatting looks weird as for me. Can you apply Xcode indentation and move `.first` to the same line as `preferences`?
```suggestion
        let mostSpecificLanguage = Bundle.preferredLocalizations(from: supportedLanguageCodesv8.map(\.identifier),
            forPreferences: preferences)
            .first
```

Sources/MapboxMaps/Style/Style+Localization.swift
```diff
@@ -31,41 +31,61 @@ extension Style {
...
 
-        // Do nothing if we do not support the locale
-        if !supportedLocaleIdentifiers.contains(locale.languageCode!) {
+        // `Bundle.preferredLocalizations(from:forPreferences:)` is just returning the first localization it could find.
+        if let mostSpecificLanguage = mostSpecificLanguage,
+            !preferredLocales.contains(where: { $0.languageCode == Locale(identifier: mostSpecificLanguage).languageCode }) {
```
OdNairy, line 47: I don't like an idea to instantiate Locale with `mostSpecificLanguage` the `N` times (as per `preferredLocales`).
Can you please extract the `Locale(identifier: mostSpecificLanguage).languageCode` part in front of `if` statement.

P.S. Feel free to use [`Optional.map`](https://developer.apple.com/documentation/swift/optional/map(_:)-7txtz) to simplify `Locale` instantiation.

Sources/MapboxMaps/Style/Style+Localization.swift
```diff
@@ -31,41 +31,61 @@ extension Style {
...
+        // Check if the passed Locale is the system or a created Locale
+        if locale == Locale.autoupdatingCurrent || locale == Locale.current {
+            preferences = Locale.preferredLanguages
+        } else {
+            preferences = [locale.identifier]
         }
 
-        for sourceInfo in vectorSources where locale.identifier.starts(with: "zh") {
+        // Lists language codes supported by Mapbox Streets v7
+        let supportedLanguageCodesv7 = ["ar", "en", "es", "fr", "de", "pt", "ru", "ja", "ko", "zh", "zh-Hans"]
```
OdNairy, line 65: I'm confused by having v7 list in the `getLocaleValue(locale: Locale)` method and v8 list in the `preferredMapboxStreetsLocalization`.
Can we have reorganise it, please?

OdNairy: We also need some space for future streets versions.

OdNairy: By the way, why there is only streets style localisation codes? What is about other core/custom styles?

maios (approved):

Sources/MapboxMaps/Style/Style+Localization.swift
```diff
@@ -31,41 +31,70 @@ extension Style {
...
     }
 
-    /// Filters through source to determine supported locale styles.
-    /// This is needed for v7 support
+    /// Returns the BCP 47 language tag supported by Mapbox Streets source v8 that is most preferred according to the given preferences.
+    /// Docs for language, region, and script codes: https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPInternational/LanguageandLocaleIDs/LanguageandLocaleIDs.html
+    internal func preferredMapboxStreetsLocalization(among preferences: [String]) -> String? {
+        let supportedLocaleIdentifiersv8 = ["ar", "en", "es", "fr", "de", "it", "pt", "ru", "zh-Hans", "zh-Hant", "ja", "ko", "vi"].map(Locale.init(identifier:))
+
+        let preferredLocales = preferences.map(Locale.init(identifier:))
```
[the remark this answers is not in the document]
pjleonard37: 👍  -- added in a check

Sources/MapboxMaps/Style/Style+Localization.swift
pjleonard37: 👍  -- modified the implementation to use `Optional.map`

Sources/MapboxMaps/Style/Style+Localization.swift
pjleonard37: These two lists reflect different support at the tileset/source level for various languages by [Streets v7](https://docs.mapbox.com/data/tilesets/reference/legacy/mapbox-streets-v7/#name-fields) and [Streets v8](https://docs.mapbox.com/data/tilesets/reference/mapbox-streets-v8/#common-fields), which would include all Mapbox core styles. In the most recent commit I added an explanation in `getLocaleValue(locale:)` and required passing the list directly to `preferredMapboxStreetsLocalization(among:, from:)` so this is more clear. 

One additional modification we could do is allow a developer to pass their own list of supported language codes (representing the set of language codes supported by their tileset). This may be outside the scope of this PR, but I can create a ticket for it.

evil159 (approved):
Great work investigating this, this worked flawlessly in my tests 🎉

PR #1733, opened 2022-11-16
Fix examples crash on iPad when presenting an action sheet

NiwakaDev:
Before this commit, LayerPositionExample crashes on iPad with the following error .

```
Thread 1: "Your application has presented a UIAlertController (<UIAlertController: 0x10851ca00>) of style UIAlertControllerStyleActionSheet from UINavigationController (<UINavigationController: 0x10881e800>). 
The modalPresentationStyle of a UIAlertController with this style is UIModalPresentationPopover. 
You must provide location information for this popover through the alert controller's popoverPresentationController. 
You must provide either a sourceView and sourceRect or a barButtonItem.  
If this information is not known when you present the alert controller, you may provide it in the UIPopoverPresentationControllerDelegate method -prepareForPopoverPresentation."
```

To fix it, provide a sourceView and sourceRect .


evil159 (approved):
Thank you @NiwakaDev for your contribution! I noticed there are other examples crashing when trying to present an action sheet on iPad and appended them to this PR.

PR #1742, opened 2022-11-18
Update gem dependencies

evil159:
This PR updates fastlane lock to `2.211.0` which is compatible with Xcode 14.


pjleonard37 (approved):
👍

PR #1747, opened 2022-11-21
Throttle location and heading updates

evil159:
This PR adds throttling to the streams of location and heading updates coming from a location provider to the location producer. This doesn't matter much for location updates as they are coming in with ~1sec interval usually. Heading updates on the other hand may come in much more frequently - especially for handheld devices, we don't make a good use of these updates as our location update animation duration is a bit over 1 second.

As part of supporting infrastructure this PR also introduces an observable entity called `Throttle` - it works as a wrapper on top of `ObservableValue`, allowing the user to specify a time window during which all value updates will be ignored except the very last one. 

## Pull request checklist:

### DIFF
--- a/Sources/MapboxMaps/Foundation/ObservableValue.swift
+++ b/Sources/MapboxMaps/Foundation/ObservableValue.swift
@@ -1,9 +1,97 @@
-internal final class ObservableValue<Value> where Value: Equatable {
-    private var observers = [Observer]() {
+import Foundation
+
+/// Wrapper for ``ObservableValue`` to throttle its value updates.
+/// Republishes only the latest event received during the time window interval.
+internal final class Throttle<Value> where Value: Equatable {
+    private var subscriptions = [BlockSubscription<Value>]() {
         didSet {
-            if !observers.isEmpty, oldValue.isEmpty {
+            if !subscriptions.isEmpty, oldValue.isEmpty {
+                cancelToken = observableValue.observe { [weak self] newValue in
+                    if self?.value == nil { // first update
+                        self?.notifyImmediately(with: newValue)
+                    } else {
+                        self?.onValueUpdated(newValue: newValue)
+                    }
+                    return true
+                }
+            } else if subscriptions.isEmpty, !oldValue.isEmpty {
+                cancelToken = nil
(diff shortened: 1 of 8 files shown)


Sources/MapboxMaps/Foundation/ObservableValue.swift
```diff
@@ -1,9 +1,99 @@
...
+    ) {
+        self.observableValue = value
+        self.dispatchQueue = dispatchQueue
+        self.windowDuration = windowDuration
+
+        cancelToken = observableValue.observe { [weak self] newValue in
+            if self?.value == nil { // first update
+                self?.notifyImmediately(with: newValue)
+            } else {
+                self?.onValueUpdated(newValue: newValue)
```
maios, line 34: what would happen if the 2nd value is emitted after a duration that is bigger than `windowDuration` since 1st value, do we emit 2nd value immediately or we wait for extra `windowDuration`to pass?

diagram

Sources/MapboxMaps/Foundation/ObservableValue.swift
```diff
@@ -1,9 +1,99 @@
...
+            }
+            self.item = nil
+        }
+
+        self.item = item
+        dispatchQueue.asyncAfter(deadline: .now() + windowDuration, execute: item)
+    }
+}
+
 internal final class ObservableValue<Value> where Value: Equatable {
```
maios, line 91: could we also have a convenience function to convert an `ObservableValue` to `Throttle`?
```
let observable = ObservableValue<Int>()

observable
  .throttle(forDuration: 3) // Throttle<Int>
  .observe { _ in } // Cancellable
```

persidskiy (changes requested):
lgtm, but it would be nice to:
1. solve problems with synchronization
2. clean up API of Throttle a little by reusing ObservableValue for throttled updates.

evil159:
After investigating this thoroughly I found out that it doesn’t particularly matter how often location/heading updates come, location interpolation will anyway propagate those changes on every frame. So addressing excessive updates of location/heading will require rethink interpolation first, and address incoming location/heading update frequency second

PR #1758, opened 2022-11-28
Update PointAnnotation.swift

Knapiii:
Make PointAnnotation open, so that we can extend it to a custom Point Annotation class

### DIFF
--- a/Sources/MapboxMaps/Annotations/Generated/PointAnnotation.swift
+++ b/Sources/MapboxMaps/Annotations/Generated/PointAnnotation.swift
@@ -1,7 +1,7 @@
 // This file is generated.
 import Foundation
 
-public struct PointAnnotation: Annotation {
+open struct PointAnnotation: Annotation {
 
     /// Identifier for this annotation
     public let id: String
@@ -52,7 +52,8 @@ public struct PointAnnotation: Annotation {
     ///   - coordinate: Coordinate where this annotation should be rendered
     public init(id: String = UUID().uuidString, coordinate: CLLocationCoordinate2D) {
         let point = Point(coordinate)
-        self.init(id: id, point: point)
+        self.id = id
+        self.point = point
     }
 
     // MARK: - Style Properties -


Sources/MapboxMaps/Annotations/Generated/PointAnnotation.swift
```diff
@@ -1,7 +1,7 @@
 // This file is generated.
 import Foundation
 
-public struct PointAnnotation: Annotation {
+open struct PointAnnotation: Annotation {
```
evil159, line 4: @Knapiii Could you elaborate one the motivation behind this change? It's not entirely clear what are you trying to achieve changing the access modifier to `open`.

Knapiii: I can not create my own Custom Annotation class and inherit `PointAnnotation`, if `PointAnnotation` is public.
I'm getting the error `Cannot inherit from non-open class 'PointAnnotation' outside of its defining module` if I try.

I want to have my own variables and states on the `PointAnnotation` class

evil159: I don't think that changing the access modifier to `open` will help to inherit from `PointAnnotation`, in Swift only classes can be subclassed. 

If you want to maintain an extra state for each annotation my suggestion is to have a separate model for it and link it with the corresponding annotation through the `id` property.
e.g.:
```swift
final class MyViewController: UIViewController {
    func viewDidLoad() {
        super.viewDidLoad()
        
        let myModels = [Store(id: "23432-234324", name: "...", location: ..., myCustomProperty: ...)]
        let annotations = myModels.map { model in
            var annotation = PointAnnotation(id: model.id, coordinate: model.location)
            annotation.textField = model.name
            return annotation
        }
    annotationManager.annotations = annotations
    }
}
```

evil159:
@Knapiii closing this, let me know if my suggestion works for you.

Sources/MapboxMaps/Annotations/Generated/PointAnnotation.swift
Knapiii: @evil159 When I tested and made `PointAnnotation` open, it actually did help me inherit `PointAnnotation`

Knapiii:
> @Knapiii closing this, let me know if my suggestion works for you.

@evil159 When I tested and made PointAnnotation open, it actually did help me inherit PointAnnotation

PR #1767, opened 2022-11-30
Update LICENSE.md and add job to check that we have the latest LICENSE

OdNairy:
This PR does two things:
1. Add a new CircleCI job to check that content of LICENSE.md includes data from dependencies
2. Updates LICENSE.md to include the latest dependencies content

## Pull request checklist:
    
MAPSIOS-462

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -128,6 +128,7 @@ workflows:
   steve:
     unless: << pipeline.parameters.turf-revision >>
     jobs:
+      - sanity-check
       - swiftlint
       - depsvalidator:
           name: Check dependencies versions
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -128,6 +128,7 @@ workflows:
   steve:
     unless: << pipeline.parameters.turf-revision >>
     jobs:
+      - sanity-check
```
evil159, line 131: Should this be a part of the release workflow(or maybe not the release workflow per se but part of release branch checks)? Since the license is taken from GL Native - it should stay the same as long as we are using the same GL Native version.

OdNairy: Right but I think we are going to add more checks to this job that make sense on each build run

evil159: It just if I understand correctly - this check fails when contents of GL Native's `LICENSE-iOS.md` differs from `LICENSE.md` and this can happen regardless of the contents of the PR, it would be quite confusing when this happens. 
I wonder if it is possible to run this job only when dependencies are changed - i.e. when Package.swift and/or MapboxMaps.podspec is changed.

OdNairy: It is already related to the dependency update PR in implicit way. We download `LICENSE-iOS.md` from the pinned GL Native version and not from the `main` branch. If we assume that GL Native team wouldn't recreate tag with different license file content it would never fail for a usual PRs

evil159: Yeah, using a pinned version would do the trick 😀

OdNairy: Fixed! 🥳

PR #1769, opened 2022-12-01
Fix examples memory leaks

evil159:
When trying to run examples tests on older hardware(iPad mini 4) I noticed that tests are failing because of the test termination due to memory pressure. There are some examples where retain cycles cause the example and its map view to leak. 

This PR addresses those leaks, as well as adds checks to tests to make sure example view controller and its map view are not leaking in the future.


pjleonard37 (approved):
👍

OdNairy (approved):

PR #1770, opened 2022-12-01
Run nightly tests on iPad

evil159:
This PR adds an iPad mini 4 to nightly examples tests run. I noticed that iPads are severely underrepresented when it comes to running tests and with examples tests serving a role of a "smoke test" - it would be useful to run this tests on an iPad as well just make sure there are not critical failures.

This PR depends on https://github.com/mapbox/mapbox-maps-ios/pull/1769 as memory leaks in the existing examples tests make iPad mini 4 terminate test runs due to memory pressure, thus failing them.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -123,7 +123,7 @@ workflows:
           scheme: Examples
           requires:
             - Build Examples tests [non-main]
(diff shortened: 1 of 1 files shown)


pjleonard37 (approved):
👍

OdNairy (approved):

Issue #1746, opened 2022-11-21
Removing PointAnnotations from a PointAnnotationManager causes style layer images to disappear


johnnewman:
## Environment

- Xcode version: 14.0.1
- iOS version: 16.0
- Devices affected: All
- Maps SDK Version: 10.10.0-rc.1, 10.9.1, and 10.9.0 (the bug is not present on 10.8.1 and earlier)

## Observed behavior and steps to reproduce

In my app, I draw PointAnnotations in multiple style layers using multiple PointAnnotationManagers. In SDK 10.9.0 and above, removing PointAnnotations from one PointAnnotationManager will cause PointAnnotations sharing the same `image` name within other PointAnnotationManagers disappear.

The only workaround I have found is to uniquely name images for each PointAnnotationManager you have. That will keep them from disappearing after removing PointAnnotations from separate managers. With this bug, it does not appear possible to reuse images across managers/style layers.

### Repro steps:

1. Draw PointAnnotations across two PointAnnotationManagers.
2. Make PointAnnotations from each manager share `image` names, like "red_icon", "blue_icon", etc.
3. Remove the PointAnnotations from one of the PointAnnotationManagers that share an icon name.
4. The images of all PointAnnotations that share the same icon name will disappear from both PointAnnotationManagers, including the manager that was unchanged.
6. If you happen to add a PointAnnotation back to the map that shares the image name with the PointAnnotations that disappeared, the images of those hidden PointAnnotations will reappear.

In this video, I have two PointAnnotationManagers and I'm using one of the managers to draw a single, random, PointAnnotation. This random annotation has orange highlighted text. When the annotation with the red icon is removed and replaced with an annotation with the blue icon, the red icons in both managers disappear. Similarly, removing the blue icon from the manager causes the blue icons in both managers to disappear.

https://user-images.githubusercontent.com/1591735/203091894-1d3b4d80-2e53-4720-9237-a262249fe00b.mp4

## Expected behavior

The same as SDK 10.8.1 and earlier. Removing PointAnnotations from one PointAnnotationManager should not cause images in other PointAnnotationManagers to disappear.

## Additional links and references

[Here is a gist that was used to capture the screen recording](https://gist.github.com/johnnewman/42b9c747d798a397231e9e36823cf7e6).

johnnewman:
[...]

PR #1775, opened 2022-12-05
Make PointAnnotationManager to not delete images that were not added by it.

maios:
Fixed https://github.com/mapbox/mapbox-maps-ios/issues/1746
PointAnnotationManager will not only delete images that were added by it when those images become unused (i.e when an annotation is removed from the list). If you want to persist an image and use it in multiple annotation managers, you should add the image to Style by `Style/addImage(_:id)` first

## Pull request checklist:


Sources/MapboxMaps/Annotations/Generated/PointAnnotationManager.swift
```diff
@@ -210,12 +213,12 @@ public class PointAnnotationManager: AnnotationManagerInternal {
 
         let newImages = Set(annotations.compactMap(\.image) + [annotationBeingDragged].compactMap(\.?.image))
         let newImageNames = Set(newImages.map(\.name))
-        let unusedImages = addedImages.subtracting(newImageNames)
+        let unusedImages = allImages.subtracting(newImageNames)
 
-        addImagesToStyleIfNeeded(style: style, images: newImages)
-        removeImages(from: style, images: unusedImages)
+        addedImages.formUnion(addImagesToStyleIfNeeded(style: style, images: newImages))
+        addedImages.subtract(removeImages(from: style, images: unusedImages))
```
OdNairy, line 219: Do we need more tests to cover that logic? 🤔

Tests/MapboxMapsTests/Annotations/Generated/PointAnnotationIntegrationTests.swift
```diff
@@ -1774,6 +1774,29 @@ final class PointAnnotationIntegrationTests: MapViewIntegrationTestCase {
         layer = try style.layer(withId: self.manager.layerId, type: SymbolLayer.self)
         XCTAssertEqual(layer.textOpacity, .constant((Style.layerPropertyDefaultValue(for: .symbol, property: "text-opacity").value as! NSNumber).doubleValue))
     }
+
+    func testAddImages() throws {
+        if style.imageExists(withId: "test-image-1") {
+            try style.removeImage(withId: "test-image-1")
+        }
```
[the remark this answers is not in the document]
maios: I added it there to make sure no image exist under that name for the tests, but it is unnecessary so I removed it :-D

Tests/MapboxMapsTests/Annotations/Generated/PointAnnotationIntegrationTests.swift
```diff
@@ -1774,6 +1774,29 @@ final class PointAnnotationIntegrationTests: MapViewIntegrationTestCase {
...
+        try style.addImage(XCTUnwrap(UIImage.emptyImage(with: .init(width: 20, height: 20))), id: "test-image-2")
+
+        var annotation1 = PointAnnotation(coordinate: .random())
+        annotation1.image = .init(image: try XCTUnwrap(UIImage.emptyImage(with: .init(width: 20, height: 20))), name: "test-image-1")
```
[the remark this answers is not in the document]
maios: > BTW what would happen to existing image if you assign different image for the same id? All existing images would be overwritten, right?

good point, if a new image is assigned under the same id using PointAnnotationManager then nothing would happen as we only add/update images to style if its identifier does not exist yet.

https://github.com/mapbox/mapbox-maps-ios/blob/main/Sources/MapboxMaps/Annotations/PointAnnotation%2BImageHandling.swift#L20

There is already a [JIRA ticket](https://mapbox.atlassian.net/browse/MAPSIOS-485) to track this issue so we can address it there

Sources/MapboxMaps/Annotations/Generated/PointAnnotationManager.swift
```diff
@@ -15,7 +15,10 @@ public class PointAnnotationManager: AnnotationManagerInternal {
     }
 
     private var needsSyncSourceAndLayer = false
-    private var addedImages = Set<String>()
+    /// List of images used by this ``PointAnnotationManager``.
+    private var allImages = Set<String>()
+    /// List of images used and added to Style by this ``PointAnnotationManager``.
+    private(set) internal var addedImages = Set<String>()
```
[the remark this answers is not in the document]
maios: Good point, I have added a shared pool as your suggestions

maios:
@OdNairy @evil159 I have changed implementation according to [Roman L suggestion](https://github.com/mapbox/mapbox-maps-ios/pull/1775#discussion_r1042187396)
- Implement `AnnotationImagesManager` which will handle all add/remove images to/from Style between annotation managers. This will keep a list of `AnnotationImagesConsumer` (currently it is a list of `PointAnnotationManager`) - a Style image will be removed if and only if the image is added by this `AnnotationImagesManager` and has 0 consumer.

Sources/MapboxMaps/Annotations/AnnotationImagesManager.swift
```diff
@@ -0,0 +1,89 @@
...
+                category: "Annnotations")
+        }
+    }
+
+    private func canRemoveStyleImage(_ imageName: String) -> Bool {
+        addedAnnotationImages.contains(imageName)
+        && imagesConsumers.allObjects.allSatisfy { !$0.isUsingStyleImage(imageName) }
+        && style.imageExists(withId: imageName)
+    }
+}
```
evil159, line 89: I think image tracking can be sealed inside of the manager itself, this way its users won't need to store the ids of the images they are interested in.

```suggestion
internal final class AnnotationImagesManager: AnnotationImagesManagerProtocol {

    private let style: StyleProtocol
    private var imageReferenceCount = NSCountedSet()

    init(style: StyleProtocol) {
        self.style = style
    }

    func addImage(_ image: UIImage, id: String, sdf: Bool, contentInsets: UIEdgeInsets) {
        imageReferenceCount.add(id)

        guard !style.imageExists(withId: id) else { return }

        do {
            try style.addImage(image, id: id, sdf: sdf, contentInsets: contentInsets)
        } catch {
            Log.warning(
                forMessage: "Could not add image to style due to error: \(error)",
                category: "Annnotations")
        }
    }

    func removeImage(_ imageId: String) {
        defer { imageReferenceCount.remove(imageId) }

        guard imageReferenceCount.count(for: imageId) <= 1 else {
            return
        }

        do {
            try style.removeImage(withId: imageId)
        } catch {
            Log.warning(
                forMessage: "Could not remove image from style due to error: \(error)",
                category: "Annnotations")
        }
    }
}
```

Sources/MapboxMaps/Annotations/AnnotationImagesManager.swift
```diff
@@ -0,0 +1,89 @@
...
+
+extension PointAnnotationManager: AnnotationImagesConsumer {
+
+    func isUsingStyleImage(_ imageName: String) -> Bool {
+        allImages.contains(imageName)
+    }
+}
+
+internal final class AnnotationImagesManager: AnnotationImagesManagerProtocol {
```
evil159, line 33: This component potentially can be useful in other use cases(e.g. `Puck2D`) and it doesn't look like there is anything annotation specific about it, could it be moved closer to Style.swift and renamed e.g. to `StyleImageManager`?

evil159 (commented):
Looking good, just want to discuss an option of making the image tracking component more generic.

Sources/MapboxMaps/Annotations/AnnotationImagesManager.swift
maios: while the implementation using `NSCountedSet` can takeaway the image ids tracking from the consumers, it does not take into account when an image is added from an external source, in that case, that image is claimed by our internal consumers, and will be removed when our consumers release it. We don't want this behavior, if an image is added from an external source, we will not take ownership of it and thus, we should not remove it unless being told by that source.
Furthermore, I'd prefer the current approach, as it can contain information of who is using the image, imo this information can be helpful for debugging.
I do like the idea of extending this into other domain like Puck2D, I will take a look in this :-D

evil159 (approved):

PR #1776, opened 2022-12-05
Disable iPad device tests

evil159:
Due to device tests constant failures iPad is removed from the list of test devices.


OdNairy (approved):
🙁

PR #1802, opened 2022-12-13
Fix scale bar exceeds maximum bar width at high latitude on high zoom.

maios:
Parity with https://github.com/mapbox/mapbox-maps-android/pull/1620
This fixes the issue where the scale bar exceeds its maximum width when `metersPerPoint` is small (usually at high altitude and high zoom level)

 | Before | After |
 | ----- | ----- |
| <video src="https://user-images.githubusercontent.com/30261043/207301095-0b6a413a-7c2e-4fb7-839e-22b5ae9e168f.mp4" width = 250/> | <video src="https://user-images.githubusercontent.com/30261043/207301118-00313447-4c2a-47a3-b7b7-074102b74da0.mp4" width = 250/> |
## Pull request checklist:


pjleonard37 (approved):
Looks great to me. Can you add a changelog entry?

evil159:
The scale bar row size seems to jitter when approaching max zoom levels - in the video below it starts around 7 second mark - if you go frame-by-frame then for zoom level 20.875 the scale bar shows 0.75 m row has a certain size, but on the next frame at zoom level 20.881 the same 0.75 m row is smaller. The same behavior is observed when further increasing the zoom level, as well as in the reverse direction around the same zoom levels.

This looks like an incorrect behavior to me - I think scale bar rows have to grow along when zooming in and shrink only when switching to lower unit or when zooming out.

This is the location I used for zooming: `latitude: -85.03966089515146, longitude: 7.76579639583062`.

https://user-images.githubusercontent.com/1478430/207590069-b7e561ac-b0aa-4666-b719-47738a4d7ab3.MP4

maios:
> The scale bar row size seems to jitter when approaching max zoom levels - in the video below it starts around 7 second mark - if you go frame-by-frame then for zoom level 20.875 the scale bar shows 0.75 m row has a certain size, but on the next frame at zoom level 20.881 the same 0.75 m row is smaller. The same behavior is observed when further increasing the zoom level, as well as in the reverse direction around the same zoom levels.
> 
> This looks like an incorrect behavior to me - I think scale bar rows have to grow along when zooming in and shrink only when switching to lower unit or when zooming out.
> 
> This is the location I used for zooming: `latitude: -85.03966089515146, longitude: 7.76579639583062`.
> 
>  RPReplay_Final1671018945.MP4

good catch, thank you! I checked the formatting of distance label in scale bar [here](https://github.com/mapbox/mapbox-maps-ios/blob/81b9c7cad5243a5d9117205bb0d613d8bdff0e89/Sources/MapboxMaps/Ornaments/DistanceFormatter.swift#L16) and realized we are rounding the value with 0.25 increment, so I changed the rounding behavior of row's distance to the same (rounded down with 0.25 increment)

evil159 (approved):
Works like a charm 🪄

PR #1818, opened 2022-12-20
Add basic signposts for rendering tracing

persidskiy:
This PR adds basic utils for logging performance-related signposts. And introduces couple such signposts for measuring metal drawing performance and MapView initialization time.

In instruments it looks like this:
Screen Shot 2022-12-20 at 2 09 40 PM

I also added  begins and ends of the gestures as points of interest in order make spotting relevant intervals easier.

To enable this feature, consumer can add this setting using settings from MapboxCommon (added it to our Examples):

```swift
import MapboxCommon

let settings = SettingsServiceFactory.getInstance(storageType: SettingsServiceStorageType.nonPersistent)
_ = settings.set(key: "com.mapbox.tracing", value: ["com.mapbox.maps": "platform"])
```

## Pull request checklist:

[MAPSIOS-518](https://mapbox.atlassian.net/browse/MAPSIOS-518)


Sources/MapboxMaps/Foundation/Tracing.swift
```diff
@@ -0,0 +1,98 @@
...
+
+private let subsystem = "com.mapbox.maps"
+
+internal extension OSLog {
+    @available(iOS 12, *)
+    private static let _poi = OSLog(subsystem: subsystem, category: .pointsOfInterest)
+    private static let _platform = OSLog(subsystem: subsystem, category: "platform")
+    private static let tracingEnabled: Bool = SettingsServiceFactory.getInstance(storageType: .nonPersistent).isTracingEnabled
+
+    static let platform: OSLog = tracingEnabled ? _platform : .disabled
```
OdNairy, line 13: Regarding the naming I personally would prefer to see something more project name related like `.map` or `.mmap`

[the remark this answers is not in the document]
OdNairy: BTW, is there a listener for settings service? We could adopt that one for reloading.

Sources/MapboxMaps/Foundation/Tracing.swift
```diff
@@ -0,0 +1,98 @@
+import Foundation
+import os.signpost
```
OdNairy, line 2: Are there any benefits in that form of imports? 
The last time I investigated there were nothing so I would prefer to keep simple `import os`

Apps/Examples/Examples/AppDelegate.swift
```diff
@@ -60,3 +63,9 @@ final class SceneDelegate: NSObject, UISceneDelegate {
         windows.append(window)
     }
 }
+
+private func enableTracing() {
+    // Enable tracing in iOS sdk and GLNative by default in Examples.
+    let settings = SettingsServiceFactory.getInstance(storageType: SettingsServiceStorageType.nonPersistent)
+    _ = settings.set(key: "com.mapbox.tracing", value: ["com.mapbox.maps": "platform"])
```
OdNairy, line 70: Can we have that setting in arguments of Examples scheme? So it would be `enableTracingIfNeeded` function that checks the state of some custom `UserDefaults` key (like `EnableTracing`)

Sources/MapboxMaps/Foundation/Tracing.swift
```diff
@@ -0,0 +1,98 @@
...
+    }
+}
+
+
+private extension SettingsServiceInterface {
+    var isTracingEnabled: Bool {
+        do {
+            // Enable tracing in iOS sdk when signposts for GLNative are enabled.
+            let tracing = try get(key: "com.mapbox.tracing", type: Dictionary<String, String>.self).get()
+            return tracing[subsystem] == "platform"
```
OdNairy, line 91: Is it the only supported value in here? Can it be something other than just a `platform`?

Sources/MapboxMaps/Foundation/Tracing.swift
```diff
@@ -0,0 +1,98 @@
...
+
+        @available(iOS 12, *)
+        var asOSType: OSSignpostType {
+            switch self {
+            case .begin: return .begin
+            case .end: return .end
+            case .event: return .event
+            }
+        }
+    }
```
OdNairy, line 49: Can we use `OSSignpostType` instead?

Sources/MapboxMaps/Foundation/MapView.swift
```diff
@@ -571,20 +582,28 @@ open class MapView: UIView {
 
     @_spi(Metrics) public var metricsReporter: MapViewMetricsReporter?
     private func updateFromDisplayLink(displayLink: CADisplayLink) {
+        let trace = OSLog.platform.beginInterval(name: "MapView.displayLink")
+        defer { trace.end() }
+
         metricsReporter?.beforeDisplayLinkCallback(displayLink: displayLink)
```
OdNairy, line 588: We already have a metricsReporter protocol in here. I was thinking that we might adopt it for signpost needs.
Right now it is used only in benchmarking to calculate fps and drawing time.
I like an idea of optional variable here as it would reduce performance impact as much as possible in case of disabled metrics (which should be default for SDK)

OdNairy (commented):
It is cool to see a working signpost solution for iOS 12 without a need to bump the min version. Well done! 👍 
Can you also take a look on adopting the [`OSSignposter`](https://developer.apple.com/documentation/os/ossignposter) class? It has a special state that improve state checking for signpost reports.

Sources/MapboxMaps/Foundation/Tracing.swift
persidskiy: unfortunately, no :(
Even though the type is available for iOS < 12, it's static properties like `end, begin,  event` are available only from iOS 12

OdNairy: Oh, right
Pretty bad 🙁

Sources/MapboxMaps/Foundation/Tracing.swift
OdNairy: As [`NSArgumentDomain`](https://developer.apple.com/documentation/foundation/nsargumentdomain) doesn't support arrays and dictionaries, may be we can reconsider that key structure?

Sources/MapboxMaps/Foundation/Tracing.swift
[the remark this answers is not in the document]
evil159: I think it's not worth supporting the setting dynamically, especially considering that core maps reads this setting once. Mostly tracing is not an ad-hoc activity, so makes sense to have it static.
IMO using `SettingsServiceFactory` factory for enabling tracing is a bit deceiving, it gives the impression that tracing can be turned on/off at runtime, but that won't be respected.
Ideally we would indeed have it as an environment variable, e.g. `MAPBOX_MAPS_ENABLE_TRACING` that user can set in Xcode scheme.

Sources/MapboxMaps/Foundation/Tracing.swift
```diff
@@ -0,0 +1,96 @@
...
+            return ._poi
+        } else {
+            return .disabled
+        }
+    }()
+
+    func beginInterval(name: StaticString, beginMessage: String? = nil) -> SignpostInterval {
+        let interval = SignpostInterval(log: self, name: name)
+        interval.begin(message: beginMessage)
+        return interval
```
evil159, line 25: Discussed with @persidskiy non-zero impact of creating a struct, closure and capturing variables by the closure in production. In the spectrum of zero, almost-zero, infinity - we deemed it to have almost-zero impact on production code.

Sources/MapboxMaps/Foundation/MapView.swift
```diff
@@ -571,20 +582,28 @@ open class MapView: UIView {
...
         }
+        trace.event(message: "Participants")
 
         cameraAnimatorsRunner.update()
+        trace.event(message: "Camera animations")
 
         if needsDisplayRefresh {
             needsDisplayRefresh = false
+            let trace = OSLog.platform.beginInterval(name: "MetalView.draw")
+            defer { trace.end() }
```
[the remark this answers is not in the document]
OdNairy: @persidskiy Did you try to simplify that structure by calling `end` function from the deinit?

Sources/MapboxMaps/Foundation/Tracing.swift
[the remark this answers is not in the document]
OdNairy: I think environment variable is insufficient for tracing functionality because it would limit debugging functionality with runs without Xcode. UserDefaults is a great mechanism that can be supported from the Scheme as well as leave an option to enable functionality persistently.
I'm agree with @evil159 that we can drop an idea of dynamic setting support.

evil159 (approved):

Apps/DebugApp/DebugApp.xcodeproj/xcshareddata/xcschemes/DebugApp.xcscheme
```diff
@@ -81,7 +81,7 @@
    </LaunchAction>
    <ProfileAction
       buildConfiguration = "Release"
-      shouldUseLaunchSchemeArgsEnv = "YES"
+      shouldUseLaunchSchemeArgsEnv = "NO"
```
OdNairy, line 84: Why do we need that setting to be changed?

Sources/MapboxMaps/Foundation/Tracing.swift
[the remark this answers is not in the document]
OdNairy: As we discussed with @persidskiy , there is no option to collect signpost metrics on non-debug run. 
Than env solution is totally fine for me.

OdNairy (approved):

PR #1831, opened 2023-01-03
Fix check-api-compatibility for release workflow.

maios:
- Dump baseline api as requirement for check-api-compatibility.
- Dump latest api from build-dynamic job.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -251,6 +251,8 @@ workflows:
           requires: []
       - trigger-metrics-v2:
           <<: *release-anchor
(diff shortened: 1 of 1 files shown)


OdNairy (changes requested):
There is no check step. To make it work we need:
1. Base SDK dump ✅
2. Latest SDK dump ✅
3. Comparison two dumps (persisted in workspace) in the separate job (take a look on `check-api-compatibility` job)

OdNairy (approved):
My bad, check api check is already here
Well done!

PR #1826, opened 2022-12-22
Allow expressions for text-color in format expression

persidskiy:
Adds ability to use expressions for `text-color` property in `format` expression. https://docs.mapbox.com/mapbox-gl-js/style-spec/expressions/#types-format

## Pull request checklist:
 - [x] Describe the changes in this PR, especially public API changes.
 - [ ] Include before/after visuals or gifs if this PR includes visual changes.
    
 - [ ] Write tests for all new functionality. Put tests in correct [Test Plan](https://github.com/mapbox/mapbox-maps-ios/tree/main/Tests/TestPlans) (Unit, Integration, All)
   - [ ] If tests were not written, please explain why.
 - [ ] Add documentation comments for any added or updated public APIs.
 - [ ] Add any new public, top-level symbols to the Jazzy config's `custom_categories` (scripts/doc-generation/.jazzy.yaml)
 - [ ] Add a changelog entry to to bottom of the relevant section (typically the `## main` heading near the top).
 - [ ] Update the guides (internal access only), README.md, and DEVELOPING.md if their contents are impacted by these changes.
 - [ ] If this PR is a `v10.[version]` release branch fix / enhancement, merge it to `main` first and then port to `v10.[version]` release branch.

PRs must be submitted under the terms of our Contributor License Agreement [CLA](https://github.com/mapbox/mapbox-maps-ios/blob/main/CONTRIBUTING.md#contributor-license-agreement).

[MAPSIOS-520](https://mapbox.atlassian.net/browse/MAPSIOS-520)

PR #1834, opened 2023-01-05
Bugfix/mapsios 520 support decoding expression format options

maios:
Address https://github.com/mapbox/mapbox-maps-ios/pull/1826 without breaking changes.

## Pull request checklist:

### DIFF
--- a/Sources/MapboxMaps/Style/Types/ExpressionOptions.swift
+++ b/Sources/MapboxMaps/Style/Types/ExpressionOptions.swift
@@ -47,12 +47,21 @@ public struct FormatOptions: Codable, Equatable, ExpressionArgumentConvertible {
 
     /// Applies a scaling factor on text-size as specified by the root layout property.
     public var fontScale: Double?
+    /// Applies a scaling factor on text-size as specified by the root layout property.
+    /// If specified, this value will override ``fontScale`` when encoding.
+    public var fontScaleExpression: Expression?
 
     /// Overrides the font stack specified by the root layout property
     public var textFont: [String]?
+    /// Overrides the font stack specified by the root layout property
+    /// If specified, this value will override ``textFont`` when encoding.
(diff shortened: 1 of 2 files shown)


Sources/MapboxMaps/Style/Types/ExpressionOptions.swift
```diff
@@ -47,12 +47,21 @@ public struct FormatOptions: Codable, Equatable, ExpressionArgumentConvertible {
 
     /// Applies a scaling factor on text-size as specified by the root layout property.
     public var fontScale: Double?
+    /// Applies a scaling factor on text-size as specified by the root layout property.
+    /// If specified, this value will override ``fontScale`` when encoding.
+    public var fontScaleExpression: Expression?
```
persidskiy, line 52: Hi Mai, thank you for picking it up!

It's good workaround, but I have another idea - what it if we use computed properties for backward compatibility and `Value<T>` as backing store:
```swift
public var fontScaleValue: Value<Double>?
public var fontScale: Double? {
    get { fontScaleValue?.asConstant }
    set { fontScaleValue = newValue.map { .constant($0) } }
}

internal enum CodingKeys: String, CodingKey {
    case fontScaleValue = "font-scale"
}

extension Value {
    var asConstant: T? {
        switch self {
        case let .constant(c): return c
        case .expression: return nil
        }
    }
}
```

It can be better because:
- there is a single source of truth, the value in `fontScale` and `fontScaleValue` can't diverge after user modifications.
- no need to write custom encoders/decoders
- for v11 we can just remove old `fontScale` property and rename `fontScaleValue` to `fontScale`. If customers use `fontScaleValue` it's also trivial to migrate 

What do you think?

persidskiy:
Closing this pr in favor of fixed https://github.com/mapbox/mapbox-maps-ios/pull/1826

PR #1835, opened 2023-01-05
update license

sbma44:
update license to reflect current language associated with Mapbox TOS

*NOTE:* the `scripts/release/collect_license.py` file (which is where `LICENSE-template.md` is used) still counts on the presence of a Carthage Cartfile. Obviously, we no longer have this since the move to SPM. Should this script be removed from the project, and the dependency licenses managed manually? Or can the logic be ported over to use an SPM equivalent?

I believe this PR can be merged, but work to address ☝️ should be ticketed.

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -3,13 +3,9 @@
 Mapbox Maps for iOS version 10.0
 Mapbox Maps iOS SDK
 
-Copyright © 2021 Mapbox
+Copyright &copy; 2021 - 2023 Mapbox, Inc. All rights reserved.
 
-All rights reserved.
-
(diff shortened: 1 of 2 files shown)


evil159 (approved):

evil159:
> update license to reflect current language associated with Mapbox TOS
> 
> _NOTE:_ the `scripts/release/collect_license.py` file (which is where `LICENSE-template.md` is used) still counts on the presence of a Carthage Cartfile. Obviously, we no longer have this since the move to SPM. Should this script be removed from the project, and the dependency licenses managed manually? Or can the logic be ported over to use an SPM equivalent?
> 
> I believe this PR can be merged, but work to address ☝️ should be ticketed.

@sbma44 thanks for keeping the license updated! 

Good point about the `collect_license` the script, I think we should remove it altogether as repurposing it for SPM pretty much require a rewrite.

sbma44:
Thanks @evil159 ! Marking this as complete. But: shouldn't `LICENSE-template.md` also be removed, if the script that uses it has been removed?

PR #1855, opened 2023-01-18
Enable Expression to be created without an operator

### LINKED ISSUE #1571: clusterProperties does not fully support Style Specification.
## Environment

- Xcode version: 13.4.1
- iOS version: 15.6.1
- Devices affected: All devices are affected.
- Maps SDK Version: v10.8.0

it is not possible to express `[["+", ["accumulated"], ["get", "sum"]], ["get", "scalerank"]]` via `Expression`, therefore it is not possible to use `[["+", ["accumulated"], ["get", "sum"]], ["get", "scalerank"]]` to express a cluster property.

## Expected behavior

I expect to be able to use `[["+", ["accumulated"], ["get", "sum"]], ["get", "scalerank"]]` as a value in the dictionary `clusterProperties`.

In order to express `[["+", ["accumulated"], ["get", "sum"]], ["get", "scalerank"]]` and other such expressions, the type for the value of the clusterProperties dictionary must support an array of two expressions.

I believe this can be accomplished by adding another `Operator` called `expression` with associated value of type `Expression`. This will require removing `String` raw type from the enum. Custom encoder/decoder function/init will need to be added as well.

Another possible solution is to change the type of `clusterProperties` to `[String: [Expression.Element]]`. This simple solution is less type safe but much easier to implement. It would be nice to make `Expression.elements` public if this route is taken.

## Additional links and references

[clusterProperties Style spec](https://docs.mapbox.com/mapbox-gl-js/style-spec/sources/#geojson-clusterProperties)

[clusterProperties API spec](https://docs.mapbox.com/ios/maps/api/10.8.0/Structs/GeoJSONSource.html#/s:10MapboxMaps13GeoJSONSourceV17clusterPropertiesSDySSAA10ExpressionVGSgvp)

pjleonard37:
Resolves #1571 

This PR updates Expression, adding an init that does not require an operator. With this change, GeoJSONSource's `clusterProperties` will align with the style spec (see [here](https://docs.mapbox.com/mapbox-gl-js/style-spec/sources/#geojson-clusterProperties)). Following the style spec,  a cluster property needs to be an array of two elements with several accepted formats:
 
```
["propertyKey" : [operator, [mapExpression]]] 
["propertyKey" : [[operator, ['accumulated'], ['get', propertyKey]], [mapExpression]]] 
["propertyKey" : [[reduceExpression], [mapExpression]]]
```

Currently, only the first format works, and not more complex options described in the [style spec](https://docs.mapbox.com/mapbox-gl-js/style-spec/sources/#geojson-clusterProperties): 

> For more advanced use cases, in place of operator, you can use a custom reduce expression that references a special ["accumulated"] value, e.g.: {"sum": [["+", ["accumulated"], ["get", "sum"]], ["get", "scalerank"]]}

This PR allows for Expression creation without an operator so that the additional formats work. I additionally updated the Symbol Clustering Example to show how the `clusterProperties` API can work with these formats. 

## Pull request checklist:
    
### DIFF
--- a/Apps/Examples/Examples/All Examples/Annotations/PointAnnotationClusteringExample.swift
+++ b/Apps/Examples/Examples/All Examples/Annotations/PointAnnotationClusteringExample.swift
@@ -93,16 +93,12 @@ class PointAnnotationClusteringExample: UIViewController, ExampleProtocol {
             UIColor.lightPink
         }
 
-        // Create a cluster property to add to each cluster feature
-        // This will be added to the cluster textField below
-        let clusterProperty: [String: Expression] = ["pointString": Exp(.string) { "Count:\n" }]
-
         // Select the options for clustering and pass them to the PointAnnotationManager to display
         let clusterOptions = ClusterOptions(circleRadius: .expression(circleRadiusExpression),
                                             circleColor: .expression(circleColorExpression),
                                             textColor: .constant(StyleColor(.black)),
                                             textField: .expression(Exp(.concat) {
-                                                Exp(.get) {"pointString"}
+                                                Exp(.string) { "Count:\n" }
                                                 Exp(.get) {"point_count"}
                                             }),
                                             clusterRadius: 75)
--- a/Apps/Examples/Examples/All Examples/Annotations/PointAnnotationClusteringExample.swift
+++ b/Apps/Examples/Examples/All Examples/Annotations/PointAnnotationClusteringExample.swift
@@ -93,15 +93,30 @@ class PointAnnotationClusteringExample: UIViewController, ExampleProtocol {
             UIColor.lightPink
         }
 
+        // Create expression to get the total count of hydrants in a cluster
+        let sumExpression = Exp {
+            Exp(.sum) {
+                Exp(.accumulated)
+                Exp(.get) { "sum" }
+            }
+            1
+        }
+
+        // Create a cluster property to add to each cluster
+        let clusterProperties: [String: Expression] = [
+            "sum": sumExpression
+        ]
+
         // Select the options for clustering and pass them to the PointAnnotationManager to display
         let clusterOptions = ClusterOptions(circleRadius: .expression(circleRadiusExpression),
                                             circleColor: .expression(circleColorExpression),
                                             textColor: .constant(StyleColor(.black)),
                                             textField: .expression(Exp(.concat) {
                                                 Exp(.string) { "Count:\n" }
-                                                Exp(.get) {"point_count"}
+                                                Exp(.get) {"sum"} // alternatively, you can use the built-in "point_count" property: Exp(.get) {"point_count"}
                                             }),
-                                            clusterRadius: 75)
+                                            clusterRadius: 75,
+                                            clusterProperties: clusterProperties)
         let pointAnnotationManager = mapView.annotations.makePointAnnotationManager(id: clusterLayerID, clusterOptions: clusterOptions)
         pointAnnotationManager.annotations = annotations
         pointAnnotationManager.delegate = self
--- a/Apps/Examples/Examples/All Examples/Annotations/PointAnnotationClusteringExample.swift
+++ b/Apps/Examples/Examples/All Examples/Annotations/PointAnnotationClusteringExample.swift
@@ -105,8 +105,7 @@ class PointAnnotationClusteringExample: UIViewController, ExampleProtocol {
                                                 Exp(.get) {"pointString"}
                                                 Exp(.get) {"point_count"}
                                             }),
-                                            clusterRadius: 75,
-                                            clusterProperties: clusterProperty)
+                                            clusterRadius: 75)
         let pointAnnotationManager = mapView.annotations.makePointAnnotationManager(id: clusterLayerID, clusterOptions: clusterOptions)
         pointAnnotationManager.annotations = annotations
         pointAnnotationManager.delegate = self
--- a/Sources/MapboxMaps/Annotations/ClusterOptions.swift
+++ b/Sources/MapboxMaps/Annotations/ClusterOptions.swift
@@ -33,14 +33,32 @@ public struct ClusterOptions: Equatable {
     /// clustered points. Has the form `{"property_name": [operator, map_expression]}`.
     /// `operator` is any expression function that accepts at
     /// least 2 operands (e.g. `"+"` or `"max"`) — it accumulates the property value from clusters/points the
-    /// cluster contains; `map_expression` produces the value of a single point.
+    /// cluster contains; `map_expression` produces the value of a single point. Example:
     ///
-    /// Example: `{"sum": ["+", ["get", "scalerank"]]}`.
+    /// ``Expression`` syntax:
+    /// let expression = Exp(.sum) {
+    ///     Exp(.get) { "scalerank" }
+    /// }
+    /// clusterProperties: ["sum": expression]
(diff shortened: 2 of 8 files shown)


Sources/MapboxMaps/Annotations/ClusterOptions.swift
```diff
@@ -33,12 +33,30 @@ public struct ClusterOptions: Equatable {
...
     /// least 2 operands (e.g. `"+"` or `"max"`) — it accumulates the property value from clusters/points the
-    /// cluster contains; `map_expression` produces the value of a single point.
+    /// cluster contains; `map_expression` produces the value of a single point. Example:
     ///
-    /// Example: `{"sum": ["+", ["get", "scalerank"]]}`.
+    /// ``Expression`` syntax:
+    /// let expression = Exp(.sum) {
+    ///     Exp(.get) { "scalerank" }
+    /// }
+    /// clusterProperties: ["sum": expression]
```
maios, line 42:
```suggestion
    /// ```
    /// let expression = Exp(.sum) {
    ///     Exp(.get) { "scalerank" }
    /// }
    /// clusterProperties: ["sum": expression]
    /// ```
```
you can fence code block by either 3-backsticks (`) or tildes (~)

Sources/MapboxMaps/Annotations/ClusterOptions.swift
```diff
@@ -33,12 +33,30 @@ public struct ClusterOptions: Equatable {
...
+    ///
+    /// ``Expression`` syntax:
+    /// let expression = Exp {
+    ///     Exp(.sum) {
+    ///         Exp(.accumulated)
+    ///         Exp(.get) { "sum" }
+    ///     }
+    ///     Exp(.get) { "scalerank" }
+    /// }
+    /// clusterProperties: ["sum": expression]
```
maios, line 57: same here we could fence the code block
```suggestion
    /// ``Expression`` syntax:
    /// ```
    /// let expression = Exp {
    ///     Exp(.sum) {
    ///         Exp(.accumulated)
    ///         Exp(.get) { "sum" }
    ///     }
    ///     Exp(.get) { "scalerank" }
    /// }
    /// clusterProperties: ["sum": expression]
    /// ```
```

maios (changes requested):
can we also support decoding an operatorless Expression? Right now Expression decode [requires the first element to be an Operator](https://github.com/mapbox/mapbox-maps-ios/blob/c7de4a03cd27df362ceccd5d2c289781e4155983/Sources/MapboxMaps/Style/Types/Expression.swift#L80)

Apps/Examples/Examples/All Examples/Annotations/PointAnnotationClusteringExample.swift
```diff
@@ -93,20 +93,15 @@ class PointAnnotationClusteringExample: UIViewController, ExampleProtocol {
...
         let clusterOptions = ClusterOptions(circleRadius: .expression(circleRadiusExpression),
                                             circleColor: .expression(circleColorExpression),
                                             textColor: .constant(StyleColor(.black)),
                                             textField: .expression(Exp(.concat) {
-                                                Exp(.get) {"pointString"}
+                                                Exp(.string) { "Count:\n" }
                                                 Exp(.get) {"point_count"}
                                             }),
-                                            clusterRadius: 75,
-                                            clusterProperties: clusterProperty)
```
zmiao, line 109: the clustering behavior is removed, which will not reflect the example's name: PointAnnotationClustering.  Maybe still keep the `clusterProperties` by adding a sum property, even though it is functioning the same as `point_count`

pjleonard37: Without `clusterProperties` there is still clustering behavior (the default is `nil`). I do like the idea of showing all of the options though so I'll add one back in.

Sources/MapboxMaps/Annotations/ClusterOptions.swift
pjleonard37: Updated in most recent commit

Tests/MapboxMapsTests/Style/ExpressionTests/ExpressionTests.swift
```diff
@@ -81,24 +79,6 @@ final class ExpressionTests: XCTestCase {
         XCTAssertThrowsError(try JSONDecoder().decode(Expression.self, from: data))
     }
 
-    // MARK: - Helpers
-    func verifyExpressionArgument(for expression: Expression, toMatch argument: Expression.Argument, at index: Int) {
```
pjleonard37, line 85: I removed this helper function as it was only used once

Sources/MapboxMaps/Annotations/ClusterOptions.swift
pjleonard37: Updated in most recent commit

pjleonard37:
Thanks for the reviews @maios & @zmiao! I've updated the code based on your comments.

Sources/MapboxMaps/Style/Types/Expression.swift
```diff
@@ -17,20 +17,34 @@ public struct Expression: Codable, CustomStringConvertible, Equatable {
 
     /// The operator of this expression
     public var `operator`: Operator {
-        guard let first = elements.first, case Element.operator(let op) = first else {
-            fatalError("First element of the expression is not an operator.")
+        var firstElement = elements.first
+        /// If the expression starts with an argument instead of an operator, return the first operator of a contained expression if available.
+        if case let .argument(firstArgument) = firstElement,
+           case let .expression(firstExpression) = firstArgument {
+            firstElement = firstExpression.elements.first
```
evil159, line 24:
```suggestion
            return firstExpression.operator
```
Use recursion to support expressions of any depth?

Sources/MapboxMaps/Style/Types/Expression.swift
```diff
@@ -17,20 +17,34 @@ public struct Expression: Codable, CustomStringConvertible, Equatable {
...
+        }
+
+        guard let first = firstElement, case Element.operator(let op) = first else {
+            fatalError("First element of the expression is not an operator nor another expression.")
         }
         return op
```
maios, line 30: nit: can we shorten like this?
```suggestion
        switch elements.first {
            case .operator(let op): return op
            case .argument(.expression(let expression)): return expression.operator
            default:
                fatalError("First element of the expression is not an operator nor another expression.")
        }
```

maios: though I'm still a little bit concern with this approach, it doesn't make much sense to me that the Operator of an operatorless Expression is the first argument's operator. Could we update the doc of this computed property to reflect this change at least?

Sources/MapboxMaps/Style/Types/Expression.swift
pjleonard37: Using this approach I think that makes sense -- will commit this suggestion.

Sources/MapboxMaps/Style/Types/Expression.swift
pjleonard37: Agreed that it is a bit strange. An alternative approach to avoid the fatalError could be to return an empty `Operator`. 

> Could we update the doc of this computed property to reflect this change at least?
Yes, can do

maios (approved):

PR #1868, opened 2023-01-20
SwiftUI project setup

persidskiy:
This PR adds the basic wrapper around MapView for use in SwiftUI

- The new SDK is usable only from SwiftPM for now
- API is not stable :)
- New examples added to existing Examples app

## Pull request checklist:
    
[MAPSIOS-574](https://mapbox.atlassian.net/browse/MAPSIOS-574)


Sources/MapboxMapsSwiftUI/MapboxView.swift
```diff
@@ -0,0 +1,87 @@
...
+    var getstureOptions: GestureOptions = GestureOptions()
+
+    private let initialOptions: InitialOptionsProvider?
+
+    /// Creates an instance showing scpecisif region.
+    ///
+    /// - Parameters:
+    /// - camera: The camera state to display.
+    /// - initialOptions: A closure to provide initial map parameters. It gets called only once when `MapboxView` is created.
+    public init(camera: Binding<CameraState>, initialOptions: InitialOptionsProvider? = nil) {
```
maios, line 35: if there is `cameraOptions` in `MapInitOptions`, would it override the binding camera, or would it be ignored?

Sources/MapboxMapsSwiftUI/MapboxView.swift
```diff
@@ -0,0 +1,87 @@
...
+extension MapboxView {
+    private func set<T>(_ keyPath: WritableKeyPath<MapboxView, T>, _ value: T) -> Self {
+        var updated = self
+        updated[keyPath: keyPath] = value
+        return updated
+    }
+
+    /// Sets camera bounds.
+    public func cameraBounds(_ cameraBounds: CameraBoundsOptions) -> Self {
+        set(\.cameraBounds, cameraBounds)
```
maios, line 65: why is it better than `self.cameraBounds = cameraBounds`? Is there a benefit of doing this using WritableKeyPath?

Sources/MapboxMapsSwiftUI/Util.swift
```diff
@@ -0,0 +1,35 @@
...
+    func add(_ cancelable: Cancelable) {
+        cancelables.append(cancelable)
+    }
+
+    func cancel() {
+        let c = cancelables
+        cancelables.removeAll()
+        c.forEach { $0.cancel() }
+    }
+    deinit {
```
maios, line 25: let's add a newline before this line.

Sources/MapboxMapsSwiftUI/MapboxViewProxy.swift
```diff
@@ -0,0 +1,30 @@
+import SwiftUI
```
maios, line 1: seems like this import is not needed here

Package.swift
```diff
@@ -31,6 +34,10 @@ let package = Package(
                 .copy("MapboxMaps.json")
             ]
         ),
+        .target(
+            name: "MapboxMapsSwiftUI",
+            dependencies: ["MapboxMaps"]
```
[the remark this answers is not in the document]
persidskiy: I'll remove it for a while, until we make a binary distribution of the library

Sources/MapboxMapsSwiftUI/MapboxView.swift
```diff
@@ -0,0 +1,87 @@
+@_exported import MapboxMaps
+import SwiftUI
+
+public extension CameraState {
+    /// Initializes CameraState with center and zoom.
+    init(center: CLLocationCoordinate2D, zoom: CGFloat) {
+        self.init(center: center, padding: .zero, zoom: zoom, bearing: 0, pitch: 0)
```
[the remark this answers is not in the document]
persidskiy: You must be talking about the `CameraOptions` which has the nullable fields and represents "request" for change some properties.

The `CameraState` is different - it represents actual values of the camera and thats what we want for SwiftUI, since `@State` represents the source of truth for any value.

 If the user wants to update only several properties, they can just modify them: 
```swift
Button("Change Camera") {
  camera.bearing = 180
  camera.pitch = 20
}
```

Sources/MapboxMapsSwiftUI/MapboxView.swift
persidskiy: The camera will override it. 
Ideally, I would exclude cameraOptions from `MapInitOptions` in SwiftUI in order to have only one way to set a camera, but it would require creating  alternative to MapInitOptions which in overcomplication at this moment imo.

azarovalex:
@persidskiy I wonder why you've decided to set min target to iOS 14 instead of iOS 13? Are there any features that are you missing in iOS 13? I would assume it's `@StateObject`, but you doesn't seem to use it.

Apps/Examples/Examples/SiwftUI Examples/AnnotationsExample.swift
```diff
@@ -0,0 +1,43 @@
...
+            .annotations(cities.map {
+                var annotation = PointAnnotation(
+                    id: $0.name,
+                    coordinate: $0.location
+                )
+                annotation.image = .init(
+                    image: UIImage(named: "marker")!, name: "marker")
+                return annotation
+            })
+            .edgesIgnoringSafeArea(.all)
```
azarovalex, line 33: `edgesIgnoringSafeArea` is deprecated, if we are aiming at iOS 14, let's replace with https://developer.apple.com/documentation/swiftui/view/ignoressafearea(_:edges:)

Apps/Examples/Examples/SiwftUI Examples/AnnotationsExample.swift
```diff
@@ -0,0 +1,43 @@
...
+    var cities = [
+        City(name: "Helsinki", location: .helsinki),
+        City(name: "New York", location: .newYork),
+        City(name: "London", location: .london),
+        City(name: "Berlin", location: .berlin)
+    ]
+
+    var body: some View {
+        MapboxView(camera: $camera)
+            .annotations(cities.map {
```
azarovalex, line 24: I guess this won't allow us to change annotations in runtime?
I would expect this method to accept a `Binding` but we need to be cautious here not to recreate the whole MapView each time annotations change.

Apps/Examples/Examples/SiwftUI Examples/AnnotationsExample.swift
```diff
@@ -0,0 +1,43 @@
...
+@available(iOS 14.0, *)
+struct AnnotationsExample: View {
+    struct City {
+        var name: String
+        var location: CLLocationCoordinate2D
+    }
+
+    @State var camera = CameraState(center: .zero, zoom: 1)
+
+    var cities = [
```
azarovalex, line 15: I guess this won't allow us to change annotations in runtime?
We can mark `cities` with @State but it will probably recreate the whole MapView each time annotations change, idk if `MapboxView` is fine with it.

Apps/Examples/Examples/SiwftUI Examples/FeaturesQueryExample.swift
```diff
@@ -0,0 +1,112 @@
...
+@available(iOS 14.0, *)
+struct FeaturesQueryExample: View {
+    @State private var camera = CameraState(center: .newYork, zoom: 10)
+
+    @State private var queryResult: QueryResult? = nil
+
+    var body: some View {
+            Map(camera: $camera)
+                .styleURI(.streets)
+                .onMapTapGesture(action: { _, res in
```
azarovalex, line 13: Consider renaming `action` to `perform` to be consistent with SwiftUI?
https://developer.apple.com/documentation/swiftui/view/ontapgesture(count:perform:)

azarovalex: And probably `res` to `result`, I thought it means `resources` at the first glance😅

Apps/Examples/Examples/SiwftUI Examples/FeaturesQueryExample.swift
```diff
@@ -0,0 +1,112 @@
...
+            Text("Nothing found 😔")
+        } else {
+            List {
+                Section {
+                    ForEach(result.features) { f in
+                        VStack(alignment: .leading) {
+                            FeatureView(f: f)
+                        }
+                    }
+                } header: {
```
azarovalex, line 67: Consider using this https://developer.apple.com/documentation/swiftui/section/init(_:content:)-90be4
So the sample just a little bit shorter

Apps/Examples/Examples/SiwftUI Examples/FeaturesQueryExample.swift
```diff
@@ -0,0 +1,112 @@
...
+struct ResultView: View {
+    let result: QueryResult
+    var body: some View {
+        if result.features.isEmpty {
+            Text("Nothing found 😔")
+        } else {
+            List {
+                Section {
+                    ForEach(result.features) { f in
+                        VStack(alignment: .leading) {
```
azarovalex, line 63: Probably no need in `VStack` here, it doesn't seem to do anything, even without it FeatureView is correctly aligned. (at least on my device😁)
If you still need to align you can use `.frame(alignment: .leading)` modifier.

Apps/Examples/Examples/SiwftUI Examples/Util/CameraDebugView.swift
```diff
@@ -0,0 +1,69 @@
...
+    @Binding var camera: CameraState
+    func body(content: Content) -> some View {
+        content.safeOverlay(alignment: alignment) {
+            CameraDebugView(camera: $camera)
+        }
+    }
+}
+
+@available(iOS 14.0, *)
+extension View {
```
azarovalex, line 36: imho we should constraint the type to MapView here:
1) It doesn't make much sense for other views and we don't want to clog up autocomplete.
2) We won't need to pass camera binding to this method, we can grab it from the type itself and it makes more sense to me.

While I was writing this I realized it's part of the sample, so not too important, but still...

Apps/Examples/Examples/SiwftUI Examples/SwiftUIExamples.swift
```diff
@@ -0,0 +1,46 @@
+import SwiftUI
+import UIKit
+@_spi(Experimental) import MapboxMapsSwiftUI
+
+@available(iOS 14.0, *)
+struct SwiftUIExamples: View {
+    var onClose: () -> Void
```
azarovalex, line 7: Won't `@Environment(\.presentationMode) var presentationMode` and `presentationMode.wrappedValue.dismiss()` 
or 
`@Environment(\.dismiss) private var dismiss`

work instead?

Apps/Examples/Examples/SiwftUI Examples/SwiftUIExamples.swift
```diff
@@ -0,0 +1,46 @@
...
+    var body: some View {
+        NavigationView {
+            List {
+                NavigationLink("Show me the map!", destination: Map())
```
azarovalex, line 11: Not ignoring safe area, is it expected? Visually looks like a bug.

Sources/MapboxMapsSwiftUI/MapCoordinator.swift
```diff
@@ -0,0 +1,95 @@
...
+            mapView.mapboxMap.onEvery(event: .mapLoaded) { [weak self] _ in
+                guard let self = self else { return }
+                self.actions?.onMapLoaded?(self.mapView.mapboxMap)
+            }.addTo(bag)
+
+            pointAnnotationManager = mapView.annotations.makePointAnnotationManager()
+        }
+    }
+
+    private func withoutNofifications(_ block: () throws -> Void) rethrows {
```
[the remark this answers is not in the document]
maios: that's internal though :-/

Sources/MapboxMapsSwiftUI/MapCoordinator.swift
```diff
@@ -0,0 +1,95 @@
...
+    private func onTapGesure(_ gesture: UIGestureRecognizer) {
+        queriesBag.cancel()
+        guard let actions = actions else {
+            return
+        }
+        let location = gesture.location(in: mapView)
+        actions.onMapTapGesture?(location)
+
+        actions.tapActionsWithQuery.map { options, action in
+            self.mapView.mapboxMap.queryRenderedFeatures(with: location, options: options) { result in
```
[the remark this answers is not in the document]
evil159: It won't be possible to reuse the result from QRF for all tap actions as every action may have varying RenderedQueryOptions.

Sources/MapboxMapsSwiftUI/MapCoordinator.swift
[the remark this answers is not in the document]
evil159: We could work out a strategy of enabling SwiftUI module to use parts of the MapsSDK until https://github.com/apple/swift-evolution/blob/main/proposals/0386-package-access-modifier.md lands :)

Sources/MapboxMapsSwiftUI/Map.swift
```diff
@@ -0,0 +1,152 @@
...
+public struct InternalMap: UIViewRepresentable {
+    public typealias InitialOptionsProvider = () -> MapInitOptions
+    public typealias TapAction = (CGPoint) -> Void
+    public typealias TapQueryAction = (CGPoint, (Result<[QueriedFeature], Error>)) -> Void
+
+    typealias TapActionWithQueryPair = (options: RenderedQueryOptions?, action: TapQueryAction)
+    struct Actions {
+        var onMapLoaded: MapLoadedAction?
+        var onMapTapGesture: TapAction?
+        var tapActionsWithQuery = [TapActionWithQueryPair]()
```
evil159, line 35: This seems to be quite tailored to a particular use case of QRF on tap, this should already be achievable for the user by hooking in tap gesture and performing QRF themselves:
```swift
.onTapGesture { location in
    mapView.mapboxMap.queryRenderedFeatures(with: location, options: options) { result in
          model.property = result
    }
```
Is there a need for a dedicated action for this?

Apps/Examples/Examples/SwiftUI Examples/ClusteringExample.swift
```diff
@@ -0,0 +1,168 @@
+import SwiftUI
+@_spi(Experimental) import MapboxMapsSwiftUI
+
+@available(iOS 14.0, *)
+struct ClusteringExample : View {
```
pengdev, line 5: we should probably align the examples across Android and iOS regarding compose/swift UI, let's try to do cross team review on the examples we are going to provide.

Sources/MapboxMaps/Foundation/Extensions/Core/CoordinateBounds.swift
```diff
@@ -5,6 +5,9 @@ import MapboxCoreMaps
 
 public extension CoordinateBounds {
 
+    /// Returns a bounds covering the entire (unwrapped) world.
+    static var world: CoordinateBounds { Self.__world() }
```
pengdev, line 9: is it required to add the extension here?

Sources/MapboxMaps/Foundation/MapboxMap.swift
```diff
@@ -31,7 +31,7 @@ internal protocol MapboxMapProtocol: AnyObject {
     func camera(for coordinateBounds: CoordinateBounds, padding: UIEdgeInsets, bearing: Double?, pitch: Double?) -> CameraOptions
     func coordinate(for point: CGPoint) -> CLLocationCoordinate2D
     func point(for coordinate: CLLocationCoordinate2D) -> CGPoint
-    func performWithoutNotifying(_ block: () -> Void)
```
pengdev, line 34: is this main sdk change required by the swiftUI sdk? would it be possible to extend it in the swiftUI module?

Sources/MapboxMapsSwiftUI/Map.swift
evil159: Capturing from the offline convo: We might need more extensive event system which would extend beyond the traditional map events, this is an instance of such event.

Apps/Examples/Examples/Controllers/ExampleTableViewController.swift
```diff
@@ -19,8 +19,16 @@ final class ExampleTableViewController: UITableViewController {
         searchController.obscuresBackgroundDuringPresentation = false
         navigationItem.searchController = searchController
         navigationItem.hidesSearchBarWhenScrolling = false
+        if #available(iOS 14.0, *) {
```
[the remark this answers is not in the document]
evil159: Makes sense, though the fact that we have to limit the SwiftUI examples to iOS 14 might be a good indicator that we need to consider if it's worth supporting iOS 13 at all. We can always bump availability down, given enough demand. Bringing it up would require a major version bump + fair bit of justification.

Apps/Examples/Examples/SiwftUI Examples/MapSettingsExample.swift
```diff
@@ -0,0 +1,91 @@
...
+    @State private var settings = Settings()
+
+    var body: some View {
+        Map(camera: $camera, initialOptions: initialOptions)
+            .cameraBounds(settings.cameraBounds)
+            .styleURI(settings.styleURI)
+            .gestureOptions(settings.gestureOptions)
+            .edgesIgnoringSafeArea(.all)
+            // Force full map reinitialization on every orientation change
+            .id(settings.orientation)
```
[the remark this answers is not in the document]
evil159: TIL, thanks for the extended reply 👍

Sources/MapboxMapsSwiftUI/Util.swift
```diff
@@ -0,0 +1,76 @@
...
+        store.removeAll()
+        c.forEach { $0.cancel() }
+    }
+
+    deinit {
+        cancel()
+    }
+}
+
+final class BlockCancelable: Cancelable {
```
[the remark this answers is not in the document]
evil159: I think this work should be tracked in JIRA - otherwise we'll be accumulating debt with little means to track it.

Sources/MapboxMapsSwiftUI/README.md
```diff
@@ -0,0 +1,5 @@
+#  Mapbox Maps SwiftUI
+
+⚠️ Caution: This module is cunrrently in development and its API is not stable yet. We strongly recommend that it not be used in production.
+
+This module includes a SwiftUI interface for the Mapbox Maps SDK. Its purpose is to offer developers the easiest way to build great apps with Mapbox.
```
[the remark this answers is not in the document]
evil159: I think setting up the project structure and basic docs(readme&changelog) should be in the scope of project setup more than anything. Usually things like this receive little attention down the road.

evil159 (approved):
I like how Map even in this very early phase looks just at home in SwiftUI! Great work @persidskiy and @maios 👍

PR #1881, opened 2023-01-26
Add Turf to LICENSE.md

evil159:
This PR adds turf-swift section to the license file. `generate-license.sh` script can be used to regenerate license content according to latest dependencies versions.

### DIFF
--- a/scripts/license/LICENSE-template.md
+++ b/scripts/license/LICENSE-template.md
@@ -1,5 +1,4 @@
 changequote("""","""")dnl # prevents m4 from being confused with backquotes by changing quotes to non-existent tokens
-<!-- This file is generated, edit LICENSE-header.md to update the license. -->
 ## License
 
 Mapbox Maps for iOS version __MAPS_SDK_VERSION__
@@ -22,4 +21,3 @@ __TURF_LICENSE_CONTENT__
 ---
 
 __CORE_LICENSE__
-<!-- End of generated file. -->
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,4 +1,3 @@
-<!-- This file is generated, edit LICENSE-header.md to update the license. -->
 ## License
 
 Mapbox Maps for iOS version 10.11.0-rc.1
@@ -1446,4 +1445,3 @@ All other rights reserved.
 ```
 
 ---
-<!-- End of generated file. -->


scripts/license/LICENSE-header.md
```diff
@@ -0,0 +1,12 @@
+## License
+
+Mapbox Maps for iOS version 10.0
```
OdNairy, line 3: We probably have to update the header version as well

scripts/license/LICENSE-header.md
```diff
@@ -0,0 +1,12 @@
+## License
+
+Mapbox Maps for iOS version __MAPS_SDK_VERSION__
+Mapbox Maps iOS SDK
+
+Copyright &copy; 2021 - 2023 Mapbox, Inc. All rights reserved.
```
evil159, line 6: @sbma44 Would it make sense to update the end of copyright year range automatically?

scripts/license/generate-license.sh
```diff
@@ -0,0 +1,31 @@
...
+
+if [ -z "${GITHUB_TOKEN}" ]; then
+    echo "This script requires GITHUB_TOKEN variable to be set."
+    exit 1
+fi
+
+COREMAPS_VERSION=$(jq -r .MapboxCoreMaps scripts/release/packager/versions.json)
+TURF_VERSION=$(jq -r .Turf scripts/release/packager/versions.json)
+MAPS_SDK_VERSION=$(jq -r .version Sources/MapboxMaps/MapboxMaps.json)
+LICENSE_HEADER=$(m4 -D __MAPS_SDK_VERSION__="$MAPS_SDK_VERSION" scripts/license/LICENSE-header.md)
```
OdNairy, line 11: TIL: `m4`

scripts/license/generate-license.sh
```diff
@@ -0,0 +1,31 @@
...
+
+TURF_LICENSE_CONTENT=$(gh api -H "Accept: application/vnd.github+json" "/repos/mapbox/turf-swift/contents/LICENSE.md?ref=$TURF_VERSION" --jq ".content" | base64 --decode)
+CORE_LICENSE_CONTENT=$(gh api -H "Accept: application/vnd.github+json" "/repos/mapbox/mapbox-gl-native-internal/contents/LICENSE-iOS.md?ref=maps-v$COREMAPS_VERSION" --jq ".content" | base64 --decode)
+
+
+LICENSE_CONTENT="<!-- This file is generated, edit LICENSE-header.md to update the license. -->
```
OdNairy, line 17: I'm not sure that we can include _generated_ notes in the license.

scripts/license/generate-license.sh
```diff
@@ -0,0 +1,31 @@
...
+### turf-swift, https://github.com/mapbox/turf-swift
+
+\`\`\`
+"$TURF_LICENSE_CONTENT"
+\`\`\`
+
+---
+
+"$CORE_LICENSE_CONTENT"
+<!-- End of generated file. -->"
```
OdNairy, line 29: Would it make sense to apply `m4` approach for the `turf` and `gl-native` substrings as well?

.circleci/config.yml
```diff
@@ -564,22 +564,16 @@ jobs:
...
+            EXPECTED_LICENSE=$(scripts/license/generate-license.sh)
 
-            # Check if length of dependency content is 0
-            if [ ${#DEPENDENCY_CONTENT} -eq 0 ]; then
-              echo "LICENSE-iOS.md is empty"
-              exit 1
-            fi
-
-            if [[ "$(cat LICENSE.md)" == *"$DEPENDENCY_CONTENT"* ]];
+            if [[ "$(cat LICENSE.md)" == "$EXPECTED_LICENSE" ]];
```
OdNairy, line 570: It might be a bit more useful to put replace stored license with generated value then check the diff. And print that difference in case of any

OdNairy (approved):
Great work, @evil159 !
I've left a couple of suggestions but it is already pretty cool 😎

scripts/license/generate-license.sh
evil159: Same same 🤫

.circleci/config.yml
evil159: I though about doing that, but decided to go with more verbose approach and dump the expected license so one could use it to update the license file and make the check happy.

scripts/license/generate-license.sh
evil159: Great idea indeed, I retrofitted `m4` only after submitting the PR to substitute maps sdk version in the header file. Makes total sense to convert header to a template file now.

OdNairy (approved):
@evil159 As we are generating LICENSE.md now, can you please include [_generated_ flag in gitattributes](https://docs.github.com/en/repositories/working-with-files/managing-files/customizing-how-changed-files-appear-on-github)?

scripts/license/LICENSE-template.md
```diff
@@ -0,0 +1,25 @@
...
+### turf-swift, https://github.com/mapbox/turf-swift
+
+```
+__TURF_LICENSE_CONTENT__
+```
+
+---
+
+__CORE_LICENSE__
+<!-- End of generated file. -->
```
OdNairy, line 25: I find it a bad idea to include generation note in the public facing license (and we do include it in the binary form along with release available content). 
While including sub-licenses is fine, I suppose we have to receive legal approval for including extra data/metadata in the license content itself.

evil159: Approved by @sbma44. I thought that adding this note would help to prevent situations when the generated file is modified(when license is updated) mistakenly instead of the template. However, if you think it's a bad idea I'm open to removing these comments.

scripts/license/LICENSE-header.md
evil159: Proceeding with tokenizing the end year as per private conversation.

scripts/license/LICENSE-template.md
OdNairy: I suppose our CI process would prevent such situation, right? 
As Tom already approved it then it fine, but I personally would prefer a "clean" license version 🙂

evil159: True that, license check would fail in this case, I'll remove the comments then

PR #1759, opened 2022-11-28
Updated podspec to use resource_bundles

jameshurst:
This PR updates the CocoaPods podspec to use `resource_bundles` instead of `resources`.

When using CocoaPods with static linking the `resources` will be copied to the root of the `.app`. When copying these resources the build will fail if the application already contains any resources with the same name. This will happen in any application that has its own `Assets.xcassets` since the application will generate an `Assets.car` and the `MapboxMaps` pod will also try to copy its own `Assets.car` to the `.app`.

The recommended solution for this with CocoaPods is to use `resource_bundles`. This will cause CocoaPods to create a bundle with the name specified in `resource_bundles`. The created resource bundle will be copied to either the `.framework` or `.app` depending on the linking method used, resolving any resource conflicts with the hosting application.

## Pull request checklist:
 - [ ] Describe the changes in this PR, especially public API changes.
 - [ ] Include before/after visuals or gifs if this PR includes visual changes.
    
 - [ ] Write tests for all new functionality. Put tests in correct [Test Plan](https://github.com/mapbox/mapbox-maps-ios/tree/main/Tests/TestPlans) (Unit, Integration, All)
   - [ ] If tests were not written, please explain why.
 - [ ] Add documentation comments for any added or updated public APIs.
 - [ ] Add any new public, top-level symbols to the Jazzy config's `custom_categories` (scripts/doc-generation/.jazzy.yaml)
 - [ ] Add a changelog entry to to bottom of the relevant section (typically the `## main` heading near the top).
 - [ ] Update the guides (internal access only), README.md, and DEVELOPING.md if their contents are impacted by these changes.
 - [ ] If this PR is a `v10.[version]` release branch fix / enhancement, merge it to `main` first and then port to `v10.[version]` release branch.

PRs must be submitted under the terms of our Contributor License Agreement [CLA](https://github.com/mapbox/mapbox-maps-ios/blob/main/CONTRIBUTING.md#contributor-license-agreement).

PR #1888, opened 2023-01-30
Fix static linking cocoapods

evil159:
This PR is based on https://github.com/mapbox/mapbox-maps-ios/pull/1759:
> When using CocoaPods with static linking the resources will be copied to the root of the .app. When copying these resources the build will fail if the application already contains any resources with the same name. This will happen in any application that has its own Assets.xcassets since the application will generate an Assets.car and the MapboxMaps pod will also try to copy its own Assets.car to the .app.
>
> The recommended solution for this with CocoaPods is to use resource_bundles. This will cause CocoaPods to create a bundle with the name specified in resource_bundles. The created resource bundle will be copied to either the .framework or .app depending on the linking method used, resolving any resource conflicts with the hosting application.

Additionally code signing is disabled for Cocoapods distribution pending Cocoapods issue https://github.com/CocoaPods/CocoaPods/issues/11402 to be included in the upcoming release.

### DIFF
--- a/MapboxMaps.podspec
+++ b/MapboxMaps.podspec
@@ -19,11 +19,14 @@ Pod::Spec.new do |m|
   m.swift_version = '5.5'
 
   m.source_files = 'Sources/MapboxMaps/**/*.{swift,h}'
-  m.resource_bundles = { 'MapboxMapsResources' => ['Sources/**/*.{xcassets,strings}', 'Sources/MapboxMaps/MapboxMaps.json'] }
 
 # Xcode 14.x throws an error about code signing on resource bundles, turn it off for now.
 #  m.pod_target_xcconfig = { 'CODE_SIGNING_ALLOWED' => 'NO' }
 # configuration.build_settings['CODE_SIGNING_ALLOWED'] = 'NO'
+
+  m.subspec 'Resources' do |r|
+    r.resource_bundles = { 'MapboxMapsResources' => ['Sources/**/*.{xcassets,strings}', 'Sources/MapboxMaps/MapboxMaps.json'] }
+  end
 
   m.dependency 'MapboxCoreMaps', '10.11.0-rc.1'
   m.dependency 'MapboxMobileEvents', '1.0.10'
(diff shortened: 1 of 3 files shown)


MapboxMaps.podspec
```diff
@@ -19,7 +19,10 @@ Pod::Spec.new do |m|
   m.swift_version = '5.5'
 
   m.source_files = 'Sources/MapboxMaps/**/*.{swift,h}'
-  m.resources = ['Sources/**/*.{xcassets,strings}', 'Sources/MapboxMaps/MapboxMaps.json']
+  m.resource_bundles = { 'MapboxMapsResources' => ['Sources/**/*.{xcassets,strings}', 'Sources/MapboxMaps/MapboxMaps.json'] }
+
+  # Xcode 14.x throws an error about code signing on resource bundles, turn it off for now.
```
persidskiy, line 24: Let's add a TODO to remove this when cocoapods 1.12 is released, they fixed this issue https://github.com/CocoaPods/CocoaPods/pull/11723

Sources/MapboxMaps/Foundation/Extensions/Bundle+MapboxMaps.swift
```diff
@@ -11,7 +11,16 @@ extension Bundle {
         #if SWIFT_PACKAGE
         return Bundle.module
         #else
-        return Bundle(for: BundleLocator.self)
+        // When using frameworks this bundle will be our `.framework` bundle.
+        // When using static linking this bundle will be the the host application's `.app` bundle.
+        let bundle = Bundle(for: BundleLocator.self)
```
persidskiy, line 16: Basically, this bundle will be used only for binary distribution, right?

I mean, for spm the module bundle will be used, for cocoapods, it will be the MapboxMapsResources.bundle, but for binary distribution we still need BundleLocator.

evil159: Exactly!

Saikishore031:
HI Team,

Could you please help me to build iOS using azure devops pipelines with two provisioning profiles, build is getting succeeded but when trying to install the IPA file in iOS device it is getting crashed. Same code is working in local machine with Xcode 14.2

persidskiy (approved):

PR #1725, opened 2022-11-15
[MAPSIOS-449] update core styles

ank27:
Update core styles to latest version:

MAPBOX_STREETS -> mapbox://styles/mapbox/streets-v12
SATELLITE_STREETS -> mapbox://styles/mapbox/satellite-streets-v12
OUTDOORS -> mapbox://styles/mapbox/outdoors-v12
LIGHT -> mapbox://styles/mapbox/light-v11
DARK -> mapbox://styles/mapbox/dark-v11

## Pull request checklist:
 - [ ] Describe the changes in this PR, especially public API changes.
 - [ ] Include before/after visuals or gifs if this PR includes visual changes.
    
 - [ ] Write tests for all new functionality. Put tests in correct [Test Plan](https://github.com/mapbox/mapbox-maps-ios/tree/main/Tests/TestPlans) (Unit, Integration, All)
   - [ ] If tests were not written, please explain why.
 - [ ] Add documentation comments for any added or updated public APIs.
 - [ ] Add any new public, top-level symbols to the Jazzy config's `custom_categories` (scripts/doc-generation/.jazzy.yaml)
 - [ ] Add a changelog entry to to bottom of the relevant section (typically the `## main` heading near the top).
 - [ ] Update the guides (internal access only), README.md, and DEVELOPING.md if their contents are impacted by these changes.
 - [ ] If this PR is a `v10.[version]` release branch fix / enhancement, merge it to `main` first and then port to `v10.[version]` release branch.

PRs must be submitted under the terms of our Contributor License Agreement [CLA](https://github.com/mapbox/mapbox-maps-ios/blob/main/CONTRIBUTING.md#contributor-license-agreement).

PR #1900, opened 2023-02-08
Fix camera state in MapCoordinator

persidskiy:
## Pull request checklist:
    
[MAPSIOS-656
](https://mapbox.atlassian.net/browse/MAPSIOS-656)

### DIFF
--- a/Apps/Examples/Examples/SwiftUI Examples/Util/Constants.swift
+++ b/Apps/Examples/Examples/SwiftUI Examples/Util/Constants.swift
@@ -20,5 +20,6 @@ extension CameraBoundsOptions {
 }
 
 extension StyleURI {
+    static let streetsV12 = StyleURI(rawValue: "mapbox://styles/mapbox/streets-v12")!
     static let customStyle = StyleURI(rawValue: "mapbox://styles/examples/cke97f49z5rlg19l310b7uu7j")!
 }
--- a/Sources/MapboxMapsSwiftUI/MapCoordinator.swift
+++ b/Sources/MapboxMapsSwiftUI/MapCoordinator.swift
@@ -4,16 +4,16 @@ import SwiftUI
 @_spi(Experimental)
 @available(iOS 13.0, *)
 public final class MapCoordinator {
-    private var camera: Binding<CameraState>?
+    typealias CameraSetter = (CameraState) -> Void
 
     var actions: MapDependencies.Actions?
 
-    private var ignoreNotifications = false
     private var bag = Bag()
     private var queriesBag = Bag()
+    private var setCamera: CameraSetter?
 
-    init(camera: Binding<CameraState>?) {
-        self.camera = camera
+    init(setCamera: CameraSetter?) {
+        self.setCamera = setCamera
     }
 
     var mapView: MapView! {
@@ -27,10 +27,10 @@ public final class MapCoordinator {
                 self?.onTapGesure(gesture)
             }.addTo(bag)
 
-            if let camera = camera {
+            if let setCamera = setCamera {
                 mapView.mapboxMap.onEvery(event: .cameraChanged) { [weak self] _ in
                     guard let self = self else { return }
-                    camera.wrappedValue = self.mapView.cameraState
+                    setCamera(self.mapView.cameraState)
                 }.addTo(bag)
             }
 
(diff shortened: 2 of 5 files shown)


Sources/MapboxMapsSwiftUI/MapCoordinator.swift
```diff
@@ -41,35 +41,45 @@ public final class MapCoordinator {
...
-    func update(from view: InternalMap) {
-        do {
-            try mapView.mapboxMap.performWithoutNotifying {
-                if let camera = view.camera {
-                    mapView.mapboxMap.setCamera(to: CameraOptions(cameraState: camera.wrappedValue))
-                }
+    func update(camera: CameraState?, deps: MapDependencies, colorScheme: ColorScheme) {
+        let mapboxMap = mapView.mapboxMap!
 
-                // TODO: This call can change the camera, but it won't be reflected on camera Binding.
```
persidskiy, line 51: Additionnaly, solved this todo. It seems we don't have better option that just sync the camera state on next loop.

Sources/MapboxMapsSwiftUI/MapCoordinator.swift
```diff
@@ -96,3 +106,27 @@ public final class MapCoordinator {
...
+    }
+}
+
+private func set<T: Equatable>(_ oldValue: T, _ setter: (T) throws -> Void, value: T) {
+    set({ oldValue }, setter, value: value)
+}
+
+private func set<U: AnyObject, T: Equatable>(_ object: U, _ keyPath: ReferenceWritableKeyPath<U, T>, value: T) {
+    set({ object[keyPath: keyPath] }, { object[keyPath: keyPath] = $0 }, value: value)
+}
```
OdNairy, line 132: Should we extract global level function to some context? At least we have to have a separate file for them in my understanding.

Apps/Examples/Examples/SwiftUI Examples/Util/Constants.swift
```diff
@@ -20,5 +20,6 @@ extension CameraBoundsOptions {
 }
 
 extension StyleURI {
+    static let streetsV12 = StyleURI(rawValue: "mapbox://styles/mapbox/streets-v12")!
```
OdNairy, line 23: Is it time to present v12 streets style?

OdNairy (approved):
LGTM, couple of nits

Sources/MapboxMapsSwiftUI/MapCoordinator.swift
persidskiy: Yep, I was thinking about that, yet found out that they are actually quite tied to what we need here in MapCoordinator. At least, the error handled in certain non-generic way and properties can be set in several non-standardized ways.

Apps/Examples/Examples/SwiftUI Examples/Util/Constants.swift
maios: I don't think we should include this in this PR, we have a PR https://github.com/mapbox/mapbox-maps-ios/pull/1725 to update core styles but it's currently blocked.

Sources/MapboxMapsSwiftUI/MapCoordinator.swift
```diff
@@ -96,3 +106,27 @@ public final class MapCoordinator {
...
+
+private func wrapError(_ body: () throws -> Void) {
+    do {
+        try body()
+    } catch {
+        print("error: \(error)") // TODO: Logger
+    }
+}
+
+private func set<T: Equatable>(_ getter: () -> T, _ setter: (T) throws -> Void, value: T) {
```
maios, line 118:
```suggestion
private func set<T: Equatable>(_ getter: @autoclosure () -> T, _ setter: (T) throws -> Void, value: T) {
```
can this be used with `@autoclosure`? if so, the follow function. `private func set<T: Equatable>(_ oldValue: T, _ setter: (T) throws -> Void, value: T)` can be removed

maios: and can we rethrows here and let the callsite decided what do they want to do with errors?

maios (approved):

Sources/MapboxMapsSwiftUI/MapCoordinator.swift
persidskiy: Thanks for the idea, we actually don't need a closure for getter at all.

persidskiy: Since updateView is called by SwiftUI runtime we don't want any error to be thrown here, so the basic idea is that we catch all those errors and log/display them

PR #1937, opened 2023-03-24
Sync/internal to public main

maios:


OdNairy (approved):
LGTM
It seems that we are missing changelog entries in new features. Would it make sense to add them in that PR or a separate one?

PR #2111, opened 2024-02-03
Replace occurrences of `VisionOS` with `visionOS`

azarovalex:
According to [Apple guidelines](https://developer.apple.com/visionos/submit/#describing-your-app), `visionOS` should not be capitalized,  even when it’s the first word in a sentence. Think about it like writing `iOS` or `macOS`.

Ideally we should also fix the changelog in GitHub Releases: https://github.com/mapbox/mapbox-maps-ios/releases/tag/v11.2.0-beta.1

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ Mapbox welcomes participation and contributions from everyone.
 
 ### Features ✨ and improvements 🏁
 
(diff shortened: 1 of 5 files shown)


azarovalex:
Closing in favor of exactly the same PR in internal repo, it will be synced with this repo at some point later.

PR #2274, opened 2024-12-27
Allow building Examples for visionOS.

halmueller:


halmueller:
There sure is a lot of noise in `project.pbxproj`.  I am working in Xcode 16.2. Would be a cleaner merge if it's against the same tool version you're using internally. The fix was just to check the "visionOS" box for `UIColor+random.swift`.

PR #2419, opened 2026-04-20
[gestures] Add option to avoid waiting for the single tap to fail when double-tapping to zoom

### LINKED ISSUE #1560: Allow clients to conditionally set the failure dependency between the single and double tap gestures
## New Feature

With the changes in Mapbox 10.4.0 and above, the single tap gesture is delayed before it is recognized. This is due to the 10.4.0 change that requires the failure of the map's double tap gesture before the single tap gesture is recognized. ([Link to the code change here](https://github.com/mapbox/mapbox-maps-ios/pull/1113/files#diff-2118c6d8d6532a74869edccb411adbd32bc698c2d2557be7d899484e975570feR133)). It would be great if this could be conditionalized, so that clients could opt out of this dependency, allowing both gestures to recognize simultaneously. This could be done via a flag or by delegation, instead of using the `require(toFail:)` API in the internal initializer of GestureManager.

## Why

In our app, this new gesture dependency causes the single-tap selection of map annotations to be noticeably slower. This makes the app feel sluggish. If we could conditionally disable this dependency, that would make the single tap gesture much faster. It's understood that this would revert the tap behavior to the way it was pre-10.4.0, where the double tap gesture fires simultaneously with the single tap gesture. This may not be desirable to all clients, so I believe being able to conditionalize this dependency would be best!

nkartashov:
## What/why
There is [an issue](https://github.com/boolder-org/boolder-ios/issues/56) in the app called Boolder where the double-tap to zoom was disabled because after the changes in the SDK the double-tap started requiring a single tap to fail which introduced a delay and resulted in bad UX. This change adds an option to revert back to the old behavior. The option does not do anything by default.

Fixes: #1560

## Pull request checklist:
    
   - [x] If tests were not written, please explain why.
 - [x] Add any new public, top-level symbols to the DocC custom catatlog (Sources/MapboxMaps/Documentation.docc/API Catalogs)
 - [x] Update the guides (internal access only), README.md, and DEVELOPING.md if their contents are impacted by these changes.
 - [x] If this PR is a `v10.[version]` release branch fix / enhancement, merge it to `main` first and then port to `v10.[version]` release branch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ Mapbox welcomes participation and contributions from everyone.
 
 ## Features ✨ and improvements 🏁
 * Expose `FeaturesetFeature.originalFeature` property.
+### Features ✨ and improvements 🏁
+* Add `GestureOptions.singleTapRequiresDoubleTapToFail` so apps can allow single-tap gestures to resolve without waiting for double-tap zoom recognition.
 
 ## 11.23.0-rc.1 - 20 April, 2026
 * Use TileStore::setRootPath(path) and TileStore::create() instead of deprecated TileStore::create(path).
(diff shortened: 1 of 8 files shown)


nkartashov:
@persidskiy could you take a look and give feedback on this?

PR #2426, opened 2026-06-24
Bugfix/ensure reduceMemoryUse is called on main thread

MarnixArnold:
When a memory warning notification is received, it is not handled properly resulting in an internal error. Console log reports:
```
[Error, bindgen]: Function Map::reduceMemoryUse called from a thread that is not owning the object
=================================================================
Main Thread Checker: UI API called on a background thread: -[UIView bounds]
```

The cause of this is that the `MapView` class does not handle the memory warning notification on the main thread, but the function `reduceMemoryUse()` touches the UI API (`[UIView bounds]`). The fix is to mark `reduceMemoryUse()` as `@MainActor` to ensure it is always called on the main thread, and to wrap the call site in a MainActor Task.


OdNairy:
Thanks for a suggestion

Do you happen to have a reproducible example? It doesn't seem right for `didReceiveMemoryWarning` message to arrive out of the main thread:
- https://stackoverflow.com/questions/7564013/is-a-memory-warning-received-on-the-main-thread-in-ios

OdNairy:
Gonna close it until we have confirmation that `didReceiveMemoryWarning` could happen out of main thread

